PR #1, opened 2019-06-22
Add initial list of owners for event triggering ⚡

bobcatfish:
Initial list of owners will be the folks who created the design (see
https://github.com/tektoncd/pipeline/issues/315 for design docs and
discussions)


bobcatfish:
I'm also going to add a team with write access to the repo which contains the same folks to make sure they won't be blocked.

Issue #8, opened 2019-06-23
Create a skeleton controller that does nothing


bobcatfish:
It will be easier for ppl to start adding functionality if we add an empty controller that doesn't actually reconcile anything, just:

* Has all the config needed to build and deploy with `ko`
* Will start running and invoke a reconcile loop of some kind that does nothing (maybe just logs hello world?)

## Additional info

Continuing to use knative/pkg is probably our best bet for now, in which case we can get a lot of this by more or less copying https://github.com/tektoncd/pipeline/blob/master/cmd/webhook.

ncskier:
PR https://github.com/tektoncd/triggers/pull/18 completes this issue.

Issue #7, opened 2019-06-23
Make instructions in DEVELOPMENT guide work


bobcatfish:
# Expected behavior

Folks who try to follow along with our DEVELOPMENT.md should be able to actually run the commands.

(of course one way to meet this would be to delete the stuff we don't have yet, but I think it would put us in a great position to add all the supporting code we need to just make it work!)

# Actual behavior

We have no code yet, so most of the contents of the DEVELOPMENT.md would not work.

I copied most of the DEVELOPMENT.md contents from https://github.com/tektoncd/pipeline/blob/master/DEVELOPMENT.md but didn't copy the supporting code (e.g. hack dir).

Some things I noticed that we could probably make work fairly easily, that would set us up nicely:
- adding the supporting scripts in the `hack`  dir (update-deps, update-codegen)
- making update-codegen work by adding an empty or nearly empty _type.go file.
- getting `ko apply -f config/` to work, an easy way to do this would be to copy a lot of the config from https://github.com/tektoncd/pipeline/tree/master/config maybe without the bit to deploy the controller and webhook cuz we don't have those yet
- Note that in #8 we'll create a skeleton controller and in #9 we'll create a skeleton webhook

# Additional info

* DEVELOPMENT.md is being added in https://github.com/tektoncd/triggers/pull/2
* This is going to require vendoring in a lot of the deps we need; the first step to make that work will be to figure out what to vendor to get [`update-deps.sh` (or similar)](https://github.com/tektoncd/pipeline/tree/master/hack) to work.

ncskier:
@bobcatfish to clarify, are we making it possible to install `triggers` independently from `pipeline`? If so, we'll need to add yaml for our own Namespace, ClusterRole, ServiceAccount, ClusterRoleBinding, etc. (some of these might be duplicates from `pipeline`; for example, the Namespace could be the same yaml file that's in `pipeline`?).

ncskier:
I followed up with @bobcatfish about my previous comment, and I will proceed with this issue under the assumption that users can install `triggers` without having installed `pipelines` (although they will work best together). The `triggers` project will be installed into the `tekton-pipelines` namespace, but will use its own ServiceAccount.

ncskier:
Closing this issue now that `ko apply -f config/` works, the `hack` directory (with `update-deps` and `update-codegen`) works, and the skeleton controller & webhook exist.

Relavent PRs:

- docs update: https://github.com/tektoncd/triggers/pull/30
[...]

PR #18, opened 2019-07-08
Create a skeleton controller

ncskier:
# Changes

Create the skeleton controller described in issue https://github.com/tektoncd/triggers/issues/8.

- Add skeleton reconciler files for the `eventlistener`, `triggertemplate`, and `triggerbinding`
- Add config files for the controller to get a deployment going with `ko apply -f config/`.
- Add an e2e test with `ko apply -f config/`
- Add the `hack/update-deps.sh` script and a ton of `vendor/` dependencies

~#### WIP:~
- ~I ran into some issues with `dep`, discussed on Slack [here](https://tektoncd.slack.com/archives/GKV9LUVHV/p1562360044024800). The code in this PR builds correctly right now, but I would like to get some feedback about this issue.~
I've added extra constraints to the `Gopkg.toml` file to solve this issue.
- ~There is placeholder code commented out right now that depends on the generated clientset & informer code that is being added under PR https://github.com/tektoncd/triggers/pull/17.~
The PR has been merged, and I created a reconciler for the `EventListener`.
- ~The `triggers` controller Deployment uses the [`tekton-pipelines-controller`](https://github.com/tektoncd/pipeline/blob/master/config/200-serviceaccount.yaml) ServiceAccount. Under issue https://github.com/tektoncd/triggers/issues/7, this can be changed to a triggers ServiceAccount if we do not want to depend on `pipelines` being installed.~
With PR https://github.com/tektoncd/triggers/pull/23 we use our own `triggers` Service Account.

# Submitter Checklist

review them:


vdemeester:
@ncskier might be worth exploring *injection-based* controllers (see https://github.com/tektoncd/pipeline/pull/1015 as a reference)

bobcatfish:
might as well switch to the injection based controller right at the outset if we can! (sorry @ncskier 😅 )

config/controller-service.yaml
```diff
@@ -0,0 +1,29 @@
...
+  labels:
+    app: tekton-triggers-controller
+  name: tekton-triggers-controller
+  namespace: tekton-pipelines
+spec:
+  ports:
+  - name: metrics
+    port: 9090
+    protocol: TCP
+    targetPort: 9090
```
bobcatfish: we have been copying this 9090 port thing around (from knative) - do we even need it? I think it might be for emitting metrics that I'm not sure we're actually emitting? (or maybe the knative libs are?)

(just ignore me if you'd rather not look into this)

pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,137 @@
...
+	"k8s.io/client-go/tools/record"
+)
+
+// Options defines the common reconciler options.
+// We define this to reduce the boilerplate argument list when
+// creating our controllers.
+type Options struct {
+	KubeClientSet     kubernetes.Interface
+	SharedClientSet   sharedclientset.Interface
+	PipelineClientSet pipelineclientset.Interface
```
bobcatfish: oh nice, so we'll have this as a vendored dep already 👍

test/e2e-tests.sh
```diff
@@ -0,0 +1,42 @@
...
+header "Setting up environment"
+
+install_pipeline_crd
+install_triggers_crd
+
+# Run the integration tests
+header "Running Go e2e tests"
+go_test_e2e -timeout=20m ./test || fail_test
+
+success
```
bobcatfish: woot thanks for adding the end to end test boilerplate!

ncskier:
I'll take a look 😅

config/controller-service.yaml
ncskier: Looks like it's for knative/pkg/metrics, although I'm not sure if we're emitting these or not:
https://github.com/tektoncd/pipeline/blob/master/vendor/github.com/knative/pkg/metrics/config.go#L56

ncskier:
@bobcatfish @vdemeester I've updated the controller to use the knative/pkg injection infrastructure. Please take another look at reviewing this PR when you get the chance 🙂 

I only added a reconciler for the `EventListener` at the moment, because I'm not sure if we'll need reconcilers for the other types (and if we need a reconciler we can easily add one); let me know if you have an opinion about this.

I also changed the bash scripts to use `tektoncd/plumbing` instead of `test-infra`.

.golangci.yml
```diff
@@ -0,0 +1,9 @@
+run:
+  build-tags:
+  - e2e
+  skip-dirs:
+  - vendor
+  - pkg/client/clientset/versioned/fake
+linters-settings:
+  errcheck:
+    exclude: .errcheck.txt
```
bobcatfish: ooo what's this file for? ive never seen this before! :D

(question about both the .golangci.yaml and the .errcheck.txt!)

config/config-logging.yaml
```diff
@@ -0,0 +1,50 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-logging-triggers
+  namespace: tekton-pipelines
+data:
+  # Common configuration for all knative codebase
```
bobcatfish: comment out of date haha

config/config-logging.yaml
```diff
@@ -0,0 +1,50 @@
...
+        "levelEncoder": "",
+        "timeEncoder": "",
+        "durationEncoder": "",
+        "callerEncoder": ""
+      }
+    }
+
+  # Log level overrides
+  loglevel.controller: "info"
+  loglevel.webhook: "info"
```
bobcatfish: seems like the level is already at "info", i wonder why these are explicitly also setting info 🤔

config/config-observability.yaml
```diff
@@ -0,0 +1,53 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-observability
```
bobcatfish: interesting, is this configuring functionality we get out of the box with knative/pkg?

config/controller-service.yaml
bobcatfish: nice! thanks for tracking that down :D I've been wondering for so long XD

hack/update-codegen.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2018 The Knative Authors
+# Copyright 2019 The Tekton Authors
```
bobcatfish: hee hee

.golangci.yml
ncskier: This is the config file for our linter `golangci-lint` (more info [here](https://github.com/golangci/golangci-lint#config-file)). `.errcheck.txt` lists files to exclude from testing. However, I don't think we need `.errcheck.txt` anymore, because the linter error we were skipping was fixed [the correct way](https://github.com/tektoncd/triggers/pull/21#discussion_r305095390) in @vincent-pli's webhook PR 😅 

I'm going to remove the `.errcheck.txt` file in a PR that I'm putting together, because I don't think we need it anymore.

config/config-observability.yaml
ncskier: I think so... @vdemeester do you know about the `config-observability` functionality? 😄

PR #24, opened 2019-07-15
Adding type docs

### LINKED ISSUE #14: Create docs for each type
## Expected behavior

Users of Triggers should have a place to look for docs on each CRD that they'll be interacting with.

## Actual behavior

We have no docs, but in #10 we'll be adding the types. The only docs we have at the moment are in [the design doc](https://docs.google.com/document/d/1fngeNn3kGD4P_FTZjAnfERcEajS7zQhSEUaN7BYIlTw/edit#heading=h.iyqzt1brkg3o).

## Additional info

One decision we may want to make at this point is whether we want to have a markdown file per type, the way that [Tekton Pipelines does](https://github.com/tektoncd/pipeline/tree/master/docs),  or if we want this kind of documentation to live with the code and primarily be visible at https://godoc.org/.

If we decide to rely on godocs, we need to have more carefully curated godocs than https://godoc.org/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1 does, specifically:

* We need to avoid including docs here that shouldn't be user facing
* We should have inline examples
* We'd need to include docs like this https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#ordering which aren't necessarily about individual types

It may also be harder to version godocs and present them nicely in an interface such as https://knative.dev/docs/ or https://skaffold.dev/docs/ (which i think both use netlify?)

vtereso:
# Changes

Add documentation for types
Preview [here](https://github.com/vtereso/triggers/tree/Issue14-TypeDocs/docs)

Fixes #14

# Submitter Checklist

review them:


vtereso:
I separated out the files for each type as done in the main Tekton repo, but the files seem a bit sparse and the example yaml for each sort of build upon one another. Does anyone feel it should be combined?

vtereso:
Also, currently the `TriggerBinding` file specifies the service accounts which we do not have yet so I am holding this PR until this is decided.

vtereso:
One other thing to consider is that the `TriggerTemplate` is using `ParamSpec` while `TriggerBinding` is using `ParamSpec`. Maybe we should use only one? I understand that the default field provided in `ParamSpec` might be better suited to `TriggerTemplates`, but it's a thought.

ncskier:
> I separated out the files for each type as done in the main Tekton repo, but the files seem a bit sparse and the example yaml for each sort of build upon one another. Does anyone feel it should be combined?

I don't think they should be combined. We're in an early stage, and it's ok for the files to be short. I think the purpose of these individual files should be to specify the function of each individual type and how it should act.

As for the example yaml building upon one another, maybe you should move/copy these to a primitive `examples` section, or a `tutorial.md` (similar to how pipeline has [`examples`](https://github.com/tektoncd/pipeline/tree/master/examples) and the [`hello world tutorial`](https://github.com/tektoncd/pipeline/blob/master/docs/tutorial.md)).

> One other thing to consider is that the TriggerTemplate is using ParamSpec while TriggerBinding is using ParamSpec. Maybe we should use only one? I understand that the default field provided in ParamSpec might be better suited to TriggerTemplates, but it's a thought.

I think you have a typo here, I'm assuming that you mean the `TriggerTemplate` is using `ParamSpec` while `TriggerBinding` is using `Param`.

This is actually an important distinction that I think we need to keep in place. The `ParamSpec` in the `TriggerTemplate` specifies a `name`, `description`, and `default` (the only required field is `name`). This `ParamSpec` is specifying that the user can supply a `Param` for `name`. To do this, the `Param` in the `TriggerBinding` specifies a `name` and `value`. The `name` specifies which parameter in the `TriggerTemplate` will be given the `value`. So, the separation of the `ParamSpec` and `Param` types is integral to their functionality.

xihaLong:
@ncskier @vtereso I have not done a PoC for triggers that takes into account the concept outlined in "Less Template, More Tekton" https://docs.google.com/document/d/11QWLYPP-yj_K_ghQqHr3LFonFE9bfHwtooYq6qPhn8w Though it certainly has impact here.

In particular, if followed the separation of template vs. trigger would be resolved or not an issue to raise in this context

Without first seeing this PR, I had already predicted such an argument (in a pipeline context):
"Did we make the right call on what to include in the PipelineTemplate and TaskTemplate (currently named Pipeline and Task) objects? Irrelevant! Those objects do not exist and get versioned between releases!"

docs/triggertemplates.md
```diff
@@ -0,0 +1,52 @@
...
+        namespace: ${params.namespace}
+        labels:
+            triggertemplated: true
+      spec:
+        type: git
+        params:
+        - name: revision
+          value: ${params.gitrevision}
+        - name: url
+          value: ${params.namespace}
```
bobcatfish: i think this is meant to be `gitrepositoryurl`

docs/triggertemplates.md
```diff
@@ -0,0 +1,52 @@
...
+            name: simple-pipeline
+        trigger:
+          type: event
+        resources:
+        - name: git-source
+          resourceRef:
+            name: git-source
+```
+
+Similar to [Pipelines](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md), `TriggerTemplates` do not do any actual work, but instead act as the blueprint for what resources should be created. Parameters defined in a `TriggerTemplate` are overridden by those of the same name in a [`TriggerBinding`](triggerbindings.md).
```
bobcatfish: 2 total nitpicks:
- I think we usually wrap long lines like this in md files
- "are overridden by" confuses me a little, maybe something like "can be provided by"? (same comment applies to the TriggerBindings docs also)

bobcatfish:
@xihaLong thanks for the heads up! The implications of the proposal to remove templating (and types that act as templates) is a pretty dramatic departure from our current designs. While I understand the importance of minimizing user impact by getting changes like that in earlier, we wouldn't make that large of a change lightly, so I don't think we should block any of the work we have in flight (e.g. removing templating + template-like types from our triggers design). It helps that this project is super early stages so even making extreme changes would have minimal impact :)

docs/triggertemplates.md
vtereso: I modified the files so each sentence is a new line. That seems like the easiest way to wrap things more or less?

bobcatfish:
Looking great, thanks @vtereso !! 🎉 

/lgtm

Issue #25, opened 2019-07-16
K8S-styled templating - remove triggers' XYZTemplates, template types, template definitions, and variable interpolations from trigger controllers and triggers API


xihaLong:
Tekton going forward, as of last week's working group meeting, is following the template style of K8S, which can be interpreted as specifically avoiding a few things. https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/declarative-application-management.md#what-about-parameterization

2/3rds of that K8S document applies directly to how Tekton will want to pursue templates and reuse of items. I am trying to find exceptions to the rules I have found in that document, though I have none currently. Everything that can be expressed with a template within our API/controller could instead be pre-processed outside of Tekton's view and applied as API calls to Tekton.

In fact, if following a couple of the models for apply -f/ apply -k/ and merging YAML/JSON in a way transparent to K8S, there is a chance to additionally unroll patches and recover original values if desired - though that is not going to be required for now.

Where am I going with this?

Tekton will avoid templating most forms of variable interpolation at the API layer for Tekton.

This will also mean avoiding creating any objects named "XYZTemplate" and/or "XYZTemplateType" in favor of allowing external tools to be chosen by a developer for generating the API calls/YAML/JSON files.

I propose any templating options be pushed to a CLI wrapper layer that makes the user experience for Tekton nicer, though leaves the controllers and API simpler for Tekton itself.

So, I think I am not going too far out there to state: Let us not add too many intelligent forms of template logic into the controllers.

It may be argued that Pipelines and PipelineRuns also follow a similar template logic (PipelineRun is just providing a patch of values over the Pipeline template), though that is a discussion for the pipeline team rather than for Triggers.

It follows to me that Tekton can be simpler and still achieve its goals. What to consider for removal? Anything performing variable interpolation/variable substitution that the user could have provided directly instead. Additionally, most other items that might be interpolated can potentially be generated and turned into a full API call/fully filled out YAML/filled out JSON. If not, then the extra state can be considered exposed via ConfigMap, Secret, or creating a resource/file/envvar for ingestion further along (each with different implications for debugging). I am not 100% sure how we might emulate or reuse the DownwardAPIVolumeFiles variant of handing information along, though it is also a pattern to consider from K8S.

I will first refer to https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/declarative-application-management.md for consideration

[...]

PR #26, opened 2019-07-16
Update README.md on templating and remove 300-triggertemplate.yaml

xihaLong:
Following the Tekton example of reducing Templates in favor of concrete
configurations where possible may have a ripple effect here. Pipelines #850
and the related design documents refer to these updates quite a bit. One
natural outcome may be the removal of Template types in Triggers

This first commit does not at-once update the repo, as I am seeking
feedback first.

Relevant discussion in comments and on design documents here:
Tekton Templating https://docs.google.com/document/d/1h_3vSApIsuiwGkrqSiegi4NVaYG4oVzBquGAhIN6qGM
Declarative Management https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/declarative-application-management.md

updates discussion of #25
updates discussion of https://github.com/tektoncd/pipeline/issues/850

@vdemeester @dlorenc


xihaLong:
Marked as a DRAFT pull request, as I am only opening this to get further feedback. Request is not yet in a position to be merged.


@dlorenc In particular (in the past few minutes) called out a few places where Templates appear in K8S that appears to make patch-styled-templates or "apply -k"-style updates fine within a controller. That is, the base Template is also exposed to the API and registered as a resource to Kubernetes.

xihaLong:
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/ 
has a resource with "Template" in the name, as the example.

This may be considered partially at odds with https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/declarative-application-management.md which argues against any intermediate representations becoming "canon" or supported representations.

The example I will provide is for refactoring parts of the API.

If there is no intermediate XYZTemplate resource / CRD, then when I decide to split XYZTemplate into ABCTemplate + DEFTemplate there is no API update. Also, consumers of XYZConcreteResource are still able to mix/remix the set of values required for XYZConcreteResource however they wish external to the controller or external to K8S entirely

It may make sense to allow or encourage these intermediate files, though API-level care (eg. eternal support going forward) needs taken when publicly exposing and declaring support for such a detail

xihaLong:
k8s api https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/

xihaLong:
As far as things "called" template or that approach a "patch-style application" of values into YAMl/JSON, of a few hundred interfaces/endpoints, there is PodTemplate.

Noting the few other places this might have been apparent, though shows up as not introducing templates, per se
VolumeClaimTemplate can be considered mis-named, as it is an array of PersistentVolumeClaims, with no sort of templating apparent on reading the interface.
JobTemplateSpec, aside from Jobs being mostly abandoned/not beneficial to Tekton currently, is mostly named like this to refer to PodTemplates for Pods/Containers that are being noted as "instantiating these PodTemplates causes a Job to run"

xihaLong:
I will propose in pipelines that there is only kept 1 item of Pipeline/PipelineRun/Task/TaskRun, recursively defined to allow multi-layer tasks (current preference is for Pipeline). Tekton can potentially operate by monitoring Pods directly that are marked to indicate they are Tekton-produced pods without needing to create a "Run" object. Run objects are analogous to Pods in K8S, from what I can read in the docs. In fact, PodTemplates might meet all needs for "templating" or similar that is desirable within Tekton itself, removing templating from Tekton by pushing it down into PodTemplates and above Tekton into tools external to K8S entirely.

What we have remaining in Tekton is a control flow logic layer on top of PodTemplates (is there not one in PodTemplate currently?) and perhaps a handful of made-for-cicd containers and catalog entries.

xihaLong:
Less Template, More Tekton https://docs.google.com/document/d/11QWLYPP-yj_K_ghQqHr3LFonFE9bfHwtooYq6qPhn8w opened in response to comment and direct message

bobcatfish:
Hey @xihaLong ! Thanks for opening a PR: I feel like this discussion is getting pretty fragmented between the issues and PRs and docs that are talking about it.

Can we maybe try to reduce the number of places we are talking about it? I propose we close this PR for now until we've made a decision about what we want to do. I'd like to go forward (until we decide otherwise) with the designs we've been discussing so far (which includes variable replacement within the event triggering mechanisms, but fortunately in no way _requires_ it)

bobcatfish:
Gonna close this for now, can continue discussion in the doc.

PR #31, opened 2019-07-22
Remove unnecessary linter file exclusion

ncskier:
# Changes
- Remove unnecessary linter file exclusions

# Submitter Checklist

review them:

### DIFF
--- a/.errcheck.txt
+++ b/.errcheck.txt
@@ -1,1 +1,0 @@
-(*github.com/tektoncd/pipeline/vendor/go.uber.org/zap.SugaredLogger).Sync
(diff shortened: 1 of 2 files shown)


EliZucker:
Do you think there will be files in the future that should be excluded from the linter's built-in error checking? Because then we will just have to rewrite this. Other than that /lgtm .

ncskier:
I'm not sure. I assume that we'll usually want to follow the linter's guidelines for best-practices. So until we find a good reason to exclude something from the linter's error checking, I don't think we should have the `.errcheck.txt` file. I really don't feel strongly about this though, so I could also just change this PR to delete the contents of `.errcheck.txt` if you think that's a better solution. 👍

Issue #19, opened 2019-07-08
Handle arbitrary event payloads


vtereso:
# Expected Behavior
The main strength of `tektoncd/triggers`, and more specifically, a `triggerBinding` is that it will act as the glue enabling you to extract fields of an event payload to be used in a `triggerTemplate` to create arbitrary resources. 

# Current Behavior
We don't have anything for this

# Background
Since events can have any structure, we will not always have a definition for the struct to unmarshal into as done prior [here](https://github.com/tektoncd/experimental/blob/master/tekton-listener/cmd/tekton-listener/main.go#L155). Further, likely only a few of the many fields will be utilized within a `triggerTemplate` where unmarshalling the entire structure will be costly/wasteful. 

# Ideas
- Have a code gen step that will inject the library and specific type into the `triggerBinding` pods/listeners AOT, but this seems unnecessary, costly (as mentioned previous) and in conflict with the idea of having a dynamic solution. 
- I made a quick prototype using the internal `encoding/json` library and `map[string]interface{}`, but there are some tools out there (https://github.com/tidwall/gjson seems like a good option) that are much more performant.

vincent-pli:
Could we use `reflect`

bobcatfish:
> Could we use `reflect`

With the caveat that this is something I've heard other people say, and not something I've really experimented with myself, I'm under the impression that using reflection has a significant performance impact. (That and code that uses reflection can be a little confusing to follow!)

vtereso:
> I made a quick prototype using the internal encoding/json library and map[string]interface{}
This is the approach using reflect, but as mentioned, there is a considerable performance impact.

bobcatfish:
Another way to tackle this would be to initially target only creation of `PipelineResources` and `PipelineRuns`, which would reduce the scope required a bit and we could add support for arbitrary resources later (but if we can get it all at once then great!)

vtereso:
> Another way to tackle this would be to initially target only creation of `PipelineResources` and `PipelineRuns`, which would reduce the scope required a bit and we could add support for arbitrary resources later (but if we can get it all at once then great!)

I understood this work item to just be the event side extraction only, which be the logic to set `TriggerBinding` params

bobcatfish:
> I understood this work item to just be the event side extraction only, which be the logic to set TriggerBinding params

Oh good call! My bad, I didn't read it fully and assumed it was about specifying arbitrary templates in the Template - just ignore me 😅

PR #34, opened 2019-07-23
Handle arbitrary payloads

vtereso:
# Changes
Parse arbitrary payloads (valid json []byte) and extract values as specified by event binding interpolation. To be used by the eventListener pod(s) in the future.
Fixes: https://github.com/tektoncd/triggers/issues/19

# Submitter Checklist

review them:


vtereso:
I set the interpolation syntax as `$()` rather than `${}`, but it's centralized to the parse package so I can easily change it if people have gripes one way or the other. I know that Tekton was potentially looking to use this syntax? Also, within the `json` package tests, I needed to use the golang internal `json` package so I called it `gojson`, but maybe that's not the most favorable name?

pkg/parse/json/json.go
```diff
@@ -0,0 +1,34 @@
+package json
+
+import (
+	"fmt"
+	"github.com/tektoncd/triggers/pkg/parse"
+	"github.com/tidwall/gjson"
```
iancoffey: TIL

bobcatfish:
> I set the interpolation syntax as $() rather than ${}, but it's centralized to the parse package so I can easily change it if people have gripes one way or the other.

Let's do it - also that means we should change over to $() asap in pipelines now that we've gone ahead with this - ping to https://github.com/tektoncd/pipeline/issues/850 ;)

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -61,13 +61,13 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 		c.Logger.Infof("event listener %q in work queue no longer exists", key)
 		return nil
-		} else if err != nil {
-			c.Logger.Errorf("Error retreiving EventListener %q: %s", name, err)
-			return err
-		}
+	} else if err != nil {
+		c.Logger.Errorf("Error retreiving EventListener %q: %s", name, err)
+		return err
+	}
```
bobcatfish: 😅 nice catch!

pkg/parse/validate.go
```diff
@@ -0,0 +1,65 @@
...
+	return eventRegex.MatchString(path)
+}
+
+// Validates the interpolation used by TriggerTemplates
+func ValidTemplateBinding(path string) bool {
+	return paramsRegex.MatchString(path)
+}
+
+// Access paths created are separated using '.'
+// If baseAccess and recursiveAccess are both false, the desired regex will match 1 field deep
```
bobcatfish: usually the syntax for a comment like this would be for it to start with the name of a function 🤔

pkg/parse/validate.go
```diff
@@ -0,0 +1,65 @@
...
+		fieldAccessRegex = fmt.Sprintf("(\\.%s)*", InterpolationFieldRegexString)
+	} else if recursiveAccess {
+		fieldAccessRegex = fmt.Sprintf("(\\.%s)+", InterpolationFieldRegexString)
+	} else if !baseAccess && !recursiveAccess {
+		fieldAccessRegex = fmt.Sprintf("(\\.%s){1}", InterpolationFieldRegexString)
+	}
+	baseAndFieldRegex := interpolationBase + fieldAccessRegex
+	interpolatedRegex := strings.Replace(InterpolationRegex, InterpolationBaseMock, baseAndFieldRegex, 1)
+	return interpolatedRegex
+}
```
bobcatfish: I think writing a test for this would probably feel a bit like a waste of time but I feel like it would be worth it 😇  - and make it easy to see the intended behaviour

pkg/parse/json/json_test.go
```diff
@@ -0,0 +1,552 @@
...
+	  "starred_url": "https://api.github.com/users/Codertocat/starred{/owner}{/repo}",
+	  "subscriptions_url": "https://api.github.com/users/Codertocat/subscriptions",
+	  "organizations_url": "https://api.github.com/users/Codertocat/orgs",
+	  "repos_url": "https://api.github.com/users/Codertocat/repos",
+	  "events_url": "https://api.github.com/users/Codertocat/events{/privacy}",
+	  "received_events_url": "https://api.github.com/users/Codertocat/received_events",
+	  "type": "User",
+	  "site_admin": false
+	}
+}`,
```
bobcatfish: this is pretty minor but it might be easier to read this if we put the payload into separate files?

pkg/parse/json/json_test.go
```diff
@@ -0,0 +1,552 @@
...
+	}
+	for i := 0; i < len(payloads); i++ {
+		byteMessage := []byte(payloads[i])
+		m := make(map[string]interface{})
+		err := gojson.Unmarshal(byteMessage, &m)
+		if err != nil {
+			t.Error("Failed to unmarshal payload")
```
bobcatfish: maybe include which payload?

pkg/parse/json/json_test.go
```diff
@@ -0,0 +1,552 @@
...
+	  "organizations_url": "https://api.github.com/users/Codertocat/orgs",
+	  "repos_url": "https://api.github.com/users/Codertocat/repos",
+	  "events_url": "https://api.github.com/users/Codertocat/events{/privacy}",
+	  "received_events_url": "https://api.github.com/users/Codertocat/received_events",
+	  "type": "User",
+	  "site_admin": false
+	}
+}`,
+	}
+	for i := 0; i < len(payloads); i++ {
```
bobcatfish: if you use `t.Run` for each payload,  each potential failure would be represented as a different test failing :D

pkg/parse/json/json_test.go
```diff
@@ -0,0 +1,552 @@
...
+		eventBindings = append(eventBindings, parse.Interpolate(currentLocation))
+		expectedValues = append(expectedValues, string(b))
+		nestedMap, ok := value.(map[string]interface{})
+		if ok {
+			nestedBindings, nestedValues := eventBindingDigger(t, nestedMap, currentLocation)
+			eventBindings, expectedValues = append(eventBindings, nestedBindings...), append(expectedValues, nestedValues...)
+		}
+	}
+	return
+}
```
bobcatfish: im really glad to see the test logic getting a lot of love!

at the moment if I was reading these tests (and especially updating them) I'd have a bit of trouble understanding exactly what was going on b/c there is so much logic in the test function - and I'd be nervous to trust it b/c it itself is untested code

What do you think about breaking up the test into multiple functions, and actually having unit tests for those (including `eventBindingDigger`)?

pkg/parse/json/json_test.go
```diff
@@ -0,0 +1,552 @@
...
+			// Grab token to determine value type
+			expectedBufferCopy := bytes.NewBuffer(expectedBuffer.Bytes())
+			decoder := gojson.NewDecoder(expectedBufferCopy)
+			token, err := decoder.Token()
+			if err != nil {
+				t.Errorf("Error getting token from expected bytes for field %s", eventBindings[i])
+				continue
+			}
+			_, ok := token.(gojson.Delim)
+			if ok { // Compare json structures
```
bobcatfish: it's a bit confusing to me (probably b/c I have waaaay less context here :pray:) about what we're testing, e.g. what conditions would make this test fail? (probably could fix that with comments!)

pkg/parse/json/json_test.go
vtereso: Fair point.

pkg/parse/json/json_test.go
vtereso: I've seen things of this nature, but have never done it before myself. Seems good.

pkg/parse/json/json_test.go
bobcatfish: > However, if we were include many example payloads, manually enumerating all of their potential mappings might become cumbersome

I see what you mean! At the moment we only have 2, so I feel like manually including the mappings might be feasible? We could revisit as we want to add more.

pkg/parse/json/json_test.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+			payloads = append(payloads,b)
+			fileNameTrimmed := strings.TrimSuffix(path,".json")
+			testNames = append(testNames, strings.Replace(strings.Title(fileNameTrimmed),"/","",-1))
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("Unable to load example payloads: %v",err)
+	}
```
bobcatfish: i'd move grabbing all these files to a separate function (and probably calling t.Helper inside it!)

Issue #39, opened 2019-07-23
Update EventListener to consume CloudEvent


bobcatfish:
# Expected Behavior

Once we have #36 the next step is for the running binary to expose a Knative addressable endpoint that can be invoked with a CloudEvent!

It doesn't have to actually do anything with the CloudEvent at this point but maybe could log something fun?

# Actual Behavior

Once we have #36 we'll have a running pod but it won't be listening for CloudEvents yet

# Additional Info

If we have #37 at this point, we should expand the end to end test to hit the running EventListener with a predefined CloudEvent. If not, we'll need to create a follow up issue in the #33 epic to add that to the end to end test.

vtereso:
From what I have discovered, vendors (GitHub, DockerHub, etc.) do not emit CloudEvents and it would be necessary to have some middle-layer such as Knative eventing to generate them. Parsing of the CloudEvent structure is supported just like any other (arbitrary payload), but since there is an overhead to generate this it seems unnecessary to have Trigger/this work item use Knative and CloudEvents. Some people have already expressed interest to avoid using Knative. Should someone opt to use it however, it would be simple enough for configure a rediretion from the Knative sync to the Triggers addressable service. So in short, maybe we should not be directly supporting CloudEvent or Knative.

vincent-pli:
I think we need make decision before really implementation.
I prefer to support `CouldEvent`, reasons:
1. That's true the vendors do not emit `CouldEvent` but we can leverage `knative-contrib` to do the translation, the problem is some sources in `knative-contrib` is depends on `knative-serving`, like `github source`, I guess maybe we need implements ourselves.
2. Without `CloudEvent` we need define tekton's event/message structure, same, the vendor do not emit it, we also need middleware. Furthermore, I remember we try to do `filter` based on `event`. so `Couldevent` is a good available choice.

I think we try to avoid depends on `Knative` directly, `CloudEvent` is ok I guess.

BTW. I wonder to know how to expose our `receiver/listener`(deployment) to outer world?
k8s `Service` or `Ingress`, `istio`? or leave to end user?

vtereso:
> I think we need make decision before really implementation.
> I prefer to support `CouldEvent`, reasons:
> 
>     1. That's true the vendors do not emit `CouldEvent` but we can leverage `knative-contrib` to do the translation, the problem is some sources in `knative-contrib` is depends on `knative-serving`, like `github source`, I guess maybe we need implements ourselves.
> 
[...]

PR #43, opened 2019-07-25
EventListener creates/deletes dummy Deployment + Service

ncskier:
# Changes

- When an EventListener resource is created, its corresponding Deployment + Service is created
- When an EventListener resource is deleted, its corresponding Deployment + Service is deleted

Note: the Service and EventListener Pod logic are a placeholders for now, and will be updated as we see fit under issue https://github.com/tektoncd/triggers/issues/39.

Thanks @iancoffey, most of this code is a rehash of your tekton-listener prototype 😁  

# Submitter Checklist

review them:


ncskier:
@bobcatfish Are there any tests that I should include in this PR? Or are we handling tests under other issues right now?

cmd/eventlistener/main.go
```diff
@@ -0,0 +1,35 @@
...
+	fmt.Fprintf(w, "Hello from the EventListener Pod! \n")
+}
+
+func main() {
+	log.Print("EventListener Pod started")
+	log.Printf("Listen and serve on port 8082")
+
+	http.HandleFunc("/", handler)
+	log.Fatal(http.ListenAndServe(":8082", nil))
+}
```
bobcatfish: late to the party but: @ncskier in a follow up PR could we add a README for the event listener binary? sometimes it can be hard to get up to speed on what all the stuff in `cmd` is for when you're new to it

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -66,8 +77,119 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+			OwnerReferences: []metav1.OwnerReference{*el.GetOwnerReference()},
+			Labels:          labels,
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{"app": el.Name},
+			Type:     corev1.ServiceTypeLoadBalancer,
+			Ports: []corev1.ServicePort{
+				{
+					Protocol: corev1.ProtocolTCP,
+					Port:     int32(8082),
```
bobcatfish: looks like 8082 is repeated a couple times, would be good to use a constant (or command line param? or even config map?) for it

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -66,8 +77,119 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+			return err
+		}
+	} else if !reflect.DeepEqual(oldService, service) {
+		// Update the EventListener Service
+		_, err = c.KubeClientSet.CoreV1().Services(el.Namespace).Update(service)
+		if err != nil {
+			c.Logger.Errorf("Error updating EventListener Service: %s", err)
+			return err
+		}
+	}
```
bobcatfish: in future iterations i think we should start moving this logic out of hte reconciler as much as possible into a separate package or two :D (with unit tests 😇 )

(as few methods and as little logic in the reconciler as possible i always say!)

we should also introduce some reconciler level tests ("integration" tests) like https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/v1alpha1/taskrun/taskrun_test.go

ncskier:
Thanks for the feedback! It's never too late, and always helpful 😄

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
ncskier: Good idea to move the logic into a separate package 👍 
Also, I'm already ahead of you on the reconciler tests, doing the work in my [el-reconciler-test](https://github.com/ncskier/triggers/tree/el-reconciler-test) branch

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
ncskier: I think ideally this will eventually be provided in the EventListener spec (maybe we could even have the option to provide an entire ServiceSpec)

PR #51, opened 2019-07-30
Implement TriggerResourceTemplate Spec as []byte blob

vtereso:
# Changes
Implement the `TriggerResourceTemplate` using []byte blob. Since any field may have a binding, it is impossible to guarantee validation on any. Since arbitrary resources can be templated, this evaluates to arbitrary json. Since we are already parsing json bytes, storing the message in this form is easy and better than an internal `map[string]string`. 

Another option was to disallow interpolation on certain fields to get some form of validation, but I think we can revisit this one we are able to create resources. This way, we can determine if the flexibility of templating _any field_ is worthwhile and if this validation is necessary. 

# Submitter Checklist

review them:

_See [the contribution guide](https://github.com/tektoncd/triggers/blob/master/CONTRIBUTING.md) for more details.


vtereso:
[This](https://play.golang.org/p/-YLMFm7cyQr) is a small playground program that would use this as an example.

docs/triggertemplates.md
```diff
@@ -53,4 +53,7 @@ spec:
 ```
 
-Also, any parameters defined a [`TriggerBinding`](triggerbindings.md) are passed into to the `TriggerTemplate` before resource creation.
+Any parameters defined in the [`TriggerBinding`](triggerbindings.md) are passed into to the `TriggerTemplate` before resource creation.
+
+To enable support for any resource type, the resource templates are internally resolved as byte blobs.
```
bobcatfish: 👍

vdemeester:
@bobcatfish @vtereso Any reason for not using `runtime.RawExtension` ? (see https://godoc.org/k8s.io/apimachinery/pkg/runtime#RawExtension)

bobcatfish:
@vdemeester in my case I just didn't know about it! Sounds like it's better than the approach we took? Maybe you can give us some pointers? :heart: :pray:

gabemontero:
Noting the relative age of this PR, I perhaps may need to register this comment elsewhere (perhaps as a new issue, perhaps in the Tekton Dynamic File Resource proposal ?  pl).  But full disclosure I discovered this PR when researching why raw JSON was currently being used.

Please do direct me to other locations to register this if that is more appropriate.

But minimally one advantage of switching [TriggerResourceTemplate](https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/trigger_template_types.go#L51) to https://godoc.org/k8s.io/apimachinery/pkg/runtime#RawExtension as @vdemeester notes is that it gets us on the path to using the k8s dynamic rest mapper APIs, which in turn will allow the trigger controller to employ k8s subject access reviews for the k8s objects to be created.

And if the associated user in question does not have permission to create such object, abort their creation.

Imagine a user creating a TriggerTemplate whose raw json attempts to create a a privileged object that can do nasty things.

I do see from `Tekton 2019 Roadmap Reflections` that security is still further down the roadmap, and has not been a primary focus per se *yet*, but it never hurts to start laying some groundwork, right :-) ?

Anyway I've got some cycles, and can start crafting a PR for consideration, at least for simply switching to RawExtension, if folks are amenable.  I can throw some temporary commits on what the SAR flow might look like as well if the discussion goes that way.

thanks

gabemontero:
quick update - I do see some use of unstructured as I've started to prep for my PR ... as I move forward, I'll add comments in the PR around if/how the move to runtime.RawExtension impacts the use of unstructured and ultimately SARs

thanks

dibyom:
Hey @gabemontero 

Thanks for writing this up! I think a new issue would be a great place to discuss this. I think https://github.com/tektoncd/triggers/issues/77 related to SAR checks.

@wlynch also added a Dynamic Client which maybe be related: https://github.com/tektoncd/triggers/pull/238

gabemontero:
> Hey @gabemontero
> 
> Thanks for writing this up! I think a new issue would be a great place to discuss this. I think #77 related to SAR checks.

will do @dibyom thanks ... and looking at https://github.com/tektoncd/triggers/issues/77 I probably should have caught the presence of that (as I work with @pmorie ;-) )

I'll respond to the latest comments there ... that said, again to some degree, to leverage runtime.RawExtension vs. what there today is simply best practice, regardless if it helps us get to SARs

> 
> @wlynch also added a Dynamic Client which maybe be related: #238

OK cool yes that could help the move to SARs along ... I'll take a close look

dibyom:
> that said, again to some degree, to leverage runtime.RawExtension vs. what there today is simply best practice, 

Yes! I agree :)

PR #52, opened 2019-07-30
Issue38 test builders

vtereso:
# Changes
Add test builders to simplify future unit tests

# Submitter Checklist

review them:

### DIFF
--- a/pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
+++ b/pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/system"
 	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	"github.com/tektoncd/triggers/test"
+	bldr "github.com/tektoncd/triggers/test/builder"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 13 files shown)


bobcatfish:
Ah interesting, i think this review has the changes from #51 as well so either we'll want to merge that first or split them up so those changes aren't in this PR.

bobcatfish:
(p.s. i think there's a linter issue?)

test/builder/meta.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+// TypeMeta sets the TypeMeta struct with default values
+func TypeMeta(kind, apiVersion string) TypeMetaOp {
+	return func(m *metav1.TypeMeta) {
+		m.Kind = kind
+		m.APIVersion = apiVersion
+	}
+}
```
bobcatfish: ooo a meta builder, i like it :D

test/builder/meta_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func TestObjectMeta(t *testing.T) {
+	// Create struct
+	objectMeta := new(metav1.ObjectMeta)
+	// Modify
+	tb.ObjectMeta(
+		tb.Label("key", "value"),
+		tb.OwnerReference("something", "SomeKind"),
+	)(objectMeta)
+	// Expected stub
```
bobcatfish: imo this comment and the two above kinda just repeat what the code says so I don't think we need them - I like the idea of only including comments when the code can't be made clear enough on its own to not need them (https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/ seems like a decent representation of this argument)

vtereso: This area was slightly different than the others since I had to create a meta object and then pass it to the object/type meta `modifiers`, which might seem like a builder in comparison to the other files. I thought the first comment _could_ be helpful and then I got carried away. I agree they should be omitted.

vtereso:
> Ah interesting, i think this review has the changes from #51 as well so either we'll want to merge that first or split them up so those changes aren't in this PR.

I didn't want to have to make another PR to "complete" the builders (even if just a first pass) so I integrated the type definition so that I could make a modifier function for it.

vtereso:
I don't think any of the resources should actually be owners? The meta modifier for `OwnerReferences` might be incorrect.

vtereso:
I would like to see https://github.com/tektoncd/triggers/pull/57 and https://github.com/tektoncd/triggers/issues/55 discussed/closed before this is merged so it won't have to be update prematurely.
/hold

ncskier:
I added an e2e test in PR https://github.com/tektoncd/triggers/pull/53 that creates an EventListener. If you're feeling especially generous, could you please update [the test](https://github.com/tektoncd/triggers/blob/master/test/eventlistener_test.go#L39) to use a test builder function? 🙂

dibyom:
Given that #57 and #55 are both done now, is this PR good to go now @vtereso ?

vtereso:
Should be good for review. Tagging some folks: @dibyom @vdemeester  @ncskier
/hold cancel

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/system"
 	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	"github.com/tektoncd/triggers/test"
+	bldr "github.com/tektoncd/triggers/test/builder"
```
dibyom: v minor Nit. Just a comment. In most pipeline tests, we use `tb` as the identifier

vtereso: In a few different locations, `tb` was used as a shorthand name for `TriggerBinding` so I make a different alias for the package.

dibyom: ahh, makes sense!

Issue #38, opened 2019-07-23
Create builder package for tests


bobcatfish:
# Expected Behavior

[The builder package in pipelines](https://github.com/tektoncd/pipeline/tree/master/test/builder) is super handy for instantiating instances of the pipeline CRDs in unit tests! Would be great to have something like this (and tests for the builders themselves in triggers :D

# Actual Behavior

We don't have this yet!

When @vdemeester added this to pipelines it took him quite a long time and I can tell you from experience that refactoring tests to use these builders later is not a ton of fun, so it'd be great if we could get these in sooner rather than later 💃

bobcatfish:
example of how this can make tests more succinct https://github.com/tektoncd/pipeline/pull/1044/files

dibyom:
Fixed in #52

PR #53, opened 2019-07-30
EventListener e2e test

ncskier:
# Changes

- Helper functions for initializing & tearing down the e2e tests
- EventListener e2e test
- README for tests

## WIP
- ~Need to add README for tests~
- ~Questions~
  - ~We don't need the [`TestMain`](https://github.com/tektoncd/pipeline/blob/master/test/init_test.go#L137) function that Pipelines has do we?~
We do not need the `TestMain` function
- No [builder](https://github.com/tektoncd/triggers/issues/38) package at the moment, so I'm creating the EventListener manually

# Submitter Checklist

review them:


test/clients.go
```diff
@@ -0,0 +1,79 @@
...
+package test
+
+import (
+	"testing"
+
+	knativetest "github.com/knative/pkg/test"
+	pipelineclientset "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+	triggersclientset "github.com/tektoncd/triggers/pkg/client/clientset/versioned"
+	"k8s.io/client-go/kubernetes"
+	// "github.com/tektoncd/triggers/pkg/client/clientset/versioned/typed/triggers/v1alpha1"
```
bobcatfish: this is a draft so you're probably already gonna do this but we should delete any commented out code before merging 😇

test/eventlistener_test.go
```diff
@@ -0,0 +1,114 @@
...
+		_, err := c.KubeClient.CoreV1().Services(namespace).Get(el.Name, metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}); err != nil {
+		t.Fatalf("Failed to delete EventListener Service: %s", err)
+	}
+	t.Logf("EventListener's Service was deleted")
+}
```
bobcatfish: niiiice 😎

test/eventlistener_test.go
```diff
@@ -0,0 +1,114 @@
...
+	// Verify the EventListener's Deployment is deleted
+	if err = wait.PollImmediate(interval, timeout, func() (bool, error) {
+		_, err := c.KubeClient.AppsV1().Deployments(namespace).Get(el.Name, metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}); err != nil {
+		t.Fatalf("Failed to delete EventListener Deployment: %s", err)
+	}
```
bobcatfish: sometimes it can make the tests a bit cleaner to move some of this polling logic into a lib like https://github.com/tektoncd/pipeline/blob/42bc3fbd5fc3ed35bfad1f371ef585da92b5fdc6/test/wait.go

but probably not a lot cleaner, and also less valuable if we don't think folks would use the functions in other tests to so kind of a toss up! I'll leave it to your preferences :)

test/eventlistener_test.go
```diff
@@ -0,0 +1,114 @@
...
+		if err != nil && errors.IsNotFound(err) {
+			return false, nil
+		}
+		return true, err
+	}); err != nil {
+		t.Fatalf("Failed to create EventListener Deployment: %s", err)
+	}
+	t.Logf("Found EventListener's Deployment")
+
+	// Verify the EventListener's Service is created
```
bobcatfish: something we haven't discussed (i dont think) - would we expect the EventListener's status section to have info about the resulting deployment + what endpoint it could be reached at? I'm thinking maybe yes? in which case instead of polling for everything, we could poll only for the status of the EventListener and then that should imply the existence of the rest of these resources - and it would be an error if they didnt exist - no polling required!

test/eventlistener_test.go
```diff
@@ -0,0 +1,114 @@
...
+			},
+		},
+	)
+	if err != nil {
+		t.Fatalf("Failed to create EventListener: %s", err)
+	}
+	t.Logf("Created EventListener %s in namespace %s", el.Name, el.Namespace)
+
+	// Verify the EventListener's Deployment is created
+	if err = wait.PollImmediate(interval, timeout, func() (bool, error) {
```
bobcatfish: i wonder if we want to use a timeout of 10 min for these, seems a bit long - makes more sense probably for pipelines which are doing stuff like building images, maybe we could start with something shorter like 30 sec or a minute?

test/init_test.go
```diff
@@ -0,0 +1,183 @@
...
+		flag.Parse()
+		if err := flag.Set("alsologtostderr", "true"); err != nil {
+			t.Fatalf("Failed to set 'alsologtostderr' flag to 'true': %s", err)
+		}
+		logging.InitializeLogger(knativetest.Flags.LogVerbose)
+
+		if knativetest.Flags.EmitMetrics {
+			logging.InitializeMetricExporter(t.Name())
+		}
+	})
```
bobcatfish: if there is any way not to include this it would probably be good, it's not clear to me that we can even use the metrics exporting (i might be wrong tho!)

test/init_test.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func createNamespace(t *testing.T, namespace string, kubeClient kubernetes.Interface) {
+	t.Helper()
+	t.Logf("Create namespace %s to deploy to", namespace)
```
bobcatfish: maybe this is something you can tell me from running the tests yourself: im under the impression that the `t.Log` functions buffer their output and therefore won't show up until a test is done (or some random flush point during the test). If that's the case, then i think using `log` is better b/c folks running these tests can see incremental progress (since they can be slow)

(This same issue exists in pipelines and is on my sad secret TODO list to fix :(((( - if you can confirm this behaviour i pledge to make a real issue in the pipeline backlog about it asap XD)

test/init_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	if err := wait.PollImmediate(interval, timeout, func() (bool, error) {
+		_, err := kubeClient.CoreV1().ServiceAccounts(namespace).Get(defaultSA, metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
+			return false, nil
+		}
+		return true, err
+	}); err != nil {
+		t.Fatalf("Failed to get SA %q in namespace %q for tests: %s", defaultSA, namespace, err)
+	}
```
bobcatfish: hm, can you explain a bit more about the purpose of this check? i would think it would be safe to assume the default account always exists?

ncskier: I put it in because that's what the [Pipelines project](https://github.com/tektoncd/pipeline/blob/master/test/init_test.go#L52) does 😅 But I can definitely remove this check if it isn't necessary.

test/init_test.go
ncskier: The [plumbing](https://github.com/tektoncd/plumbing/blob/master/scripts/presubmit-tests.sh#L295) repo currently supports the `--emit-metrics` flag (see [here](https://github.com/tektoncd/plumbing/blob/master/scripts/presubmit-tests.sh#L295)). However, I don't know if anyone's using this option. @vdemeester do you know if the `--emit-metrics` option is being used?

ncskier:
In the [pull-tekton-triggers-integration-tests](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_triggers/53/pull-tekton-triggers-integration-tests/1156539512168386562/), I am getting this error 
```
Failed to create kubernetes clientset from config file at /root/.kube/config: No Auth Provider found for name "gcp"
```
The error comes from trying to initialize the kubernetes clientset [here](https://github.com/tektoncd/triggers/pull/53/files#diff-4dc41e2a54dd91bc4e05804909237ea5R59).
@bobcatfish @vdemeester how is the test cluster set up? Do we need to do some additional setup to the test cluster to fix this error?

test/init_test.go
vdemeester: On `tektoncd` it is not being used I don't think

ncskier: Ok, thanks Vincent. I don't know if I should keep it then. Since the plumbing repo currently references an `--emit-metrics` flag, I'm leaning toward keeping it in. However, I don't feel strongly either way 😁

test/eventlistener_test.go
ncskier: Good point! I created a new issue [here](https://github.com/tektoncd/triggers/issues/54). Please update the issue if you have more to add on the subject 👍

test/init_test.go
ncskier: So, from what I understand, not using `t.Logf` can mess up where your logs get printed. So while they might be logged in "real time," the logs might look jumbled (some info here https://stackoverflow.com/questions/23205419/how-do-you-print-in-a-go-test-using-the-testing-package).

There is an approved proposal open to stream `t.Log` output. Looks interesting, although I don't know if/when it will be available: https://github.com/golang/go/issues/24929.

test/eventlistener_test.go
ncskier: I created one for now. Although given that we'll probably be [polling on the EventListener's `Status`](https://github.com/tektoncd/triggers/pull/53#discussion_r309023062) in the future, I don't know if these functions will stay around for too long 😄

test/init_test.go
ncskier: > Metrics will be emitted for these Wait methods tracking how long test poll for.

(from pipeline's [`test/README`](https://github.com/tektoncd/pipeline/tree/master/test))
Hm, looks like the pipeline's `wait` library is [emitting metrics](https://github.com/tektoncd/pipeline/blob/master/test/wait.go#L76-L77). This data could be useful in the future, although I don't know if anyone's using this data right now.

vdemeester: yes :+1:

test/init_test.go
vtereso: Maybe pipelines needs this in case no service account is provided so that there is a fall back?

Issue #55, opened 2019-07-31
Proposal to move TriggerBinding-TriggerTemplate connection out of TriggerBinding definition


EliZucker:
In the docs for `TriggerBinding`s, we say:

> Although similar, the separation of these two resources was deliberate to **encourage TriggerTemplate definitions to be reusable**

I don't disagree with this idea and I don't think `TriggerTemplate` reusability should change. 

However, for many users, it may end up being even more important for `TriggerBinding`s to be reusable! `TriggerBinding` reusability is severely limited in the current design because within the definition, a specific `TriggerTemplate` is referenced within the `TemplateRef`. 

To give an idea of what I'm saying, imagine a user who is getting tons of GitHub events from various contexts/sources. Depending on the situation, the user has defined several `TriggerTemplate`s, and each will execute different jobs like building code, performing test, moderating comments, etc. Now, for each of those `TriggerTemplate`s, which are intended to be called upon separately in differently contexts, the user must completely redefine or copy-paste the payload->parameter logic, because that information is within `TriggerBinding`, which is tied to a specific `TriggerTemplate` via `TemplateRef`. This could become extremely tedious especially for users with complex payloads.

Bottom line, I think `TriggerTemplate`s and `TemplateRef`s should each be reusable for their own reasons, and the way to do this is to move the `TriggerBinding`→`TriggerTemplate` connection representation out of the `TriggerBinding` itself. This could be done by moving the connection information a layer up into `EventListener`, or to create a new resource, representing that connection and possible conditional logic and then listing those new connection resources within `EventListener`.

My proposal would be to move the connection information into `EventListener`. This would mean removing the `templateRef` field from `TriggerBinding`s, and then an `EventListener` could look something like this:

```
apiVersion: tekton.dev/v1alpha1
kind: EventListener
metadata:
  name: simple-listener
spec:
  connections:
    - triggerBinding: simple-pipeline-binding
      triggerTemplate: simple-pipeline-template
    - triggerBinding: another-pipeline-binding
      triggerTemplate: another-pipeline-template

```

Conditional/filtering information would also go within the `connections` field. This way, the payload-variable logic is truly encapsulated and reusable within `triggerBinding`. I see `EventListener` analogous to a `PipelineRun` or `TaskRun`, in that it provides project-specific configuration details like connections, filtering, or perhaps default parameter/payload values, while `triggerBinding` and `triggerTemplate` can be reused as needed.

vtereso:
[...]

PR #57, opened 2019-08-01
Update TriggerBinding type to only have one event type

ncskier:
# Changes

- Remove Event type from the list of bindings in the TriggerBinding spec.
  - The EventListener can have multiple TriggerBindings, so each TriggerBinding having multiple events seems like overlapping functionality. So, with this PR each TriggerBinding will only have one event type.

# Submitter Checklist

review them:


EliZucker:
If we do this then I propose something like what was mentioned in #55 . With this PR, TriggerBindings will not be reusable at all because of the single TriggerTemplate tied to it in its spec.

ncskier:
@EliZucker My intention was only to tie each TriggerBinding to one event. I didn't mean to mess with the reusability of TriggerBindings. I just made an update; would this be a better change?

vincent-pli:
If `TriggerBinding` only has one event type, have to put the `Condition/Router` to `EventListener`.

ncskier:
@vincent-pli why would this mean the Condition/Router would have to go in the `EventListener`? The EventListener Pod can still use the Condition/Router info in the `TriggerBinding` to route events.

vincent-pli:
@ncskier 
Based on this pr, the `TriggerBinding` will only has one event type, that's means if a `EventListener` want to handle a dozen of event type, it need refer to dozen of `TriggerBinding`.
if the `Condition/Router` is in the `TriggerBinding`, it need get all `TriggerBinding` from  k8s cluster and check them one by one to find the right one.

ncskier:
@vincent-pli Yes, that is correct. At the moment I think that we should limit each `TriggerBinding` to supporting one event type, because this keeps the `TriggerBinding` resource from becoming too bloated & confusing. To address your technical concern about the `EventListener` needing to check each binding one-by-one, if I'm not mistaken, I think we can have the `EventListener` maintain an internal map of its `TriggerBinding`s' routing information, to reduce the time it takes to iterate over the bindings.

Also, unless we only allow the `EventListener` to use one `TriggerBinding`, then this iterating process is inevitable. This PR is intended to fix/simplify the overlapping nature of `EventListeners` having multiple `TriggerBindings`, and `TriggerBindings` also receiving multiple events. I think the best solution is to force each `TriggerBinding` to only receive one event type specified by the `Condition/Router` information.

ncskier:
It looks like we're doing a bigger refactor of the `TriggerBinding` in PR https://github.com/tektoncd/triggers/pull/63, so I'm going to close this PR

PR #60, opened 2019-08-02
Update EventListener deployment. Set up EventListener pod trigger cli…

vtereso:
Update EventListener deployment. Set up EventListener pod trigger client to pull all necessary resources from EventListener ref.

# Changes

# Submitter Checklist

review them:

### DIFF
--- a/cmd/eventlistener/main.go
+++ b/cmd/eventlistener/main.go
@@ -16,20 +16,54 @@ limitations under the License.
 package main
 
 import (
-	"fmt"
+	triggersClientset "github.com/tektoncd/triggers/pkg/client/clientset/versioned"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
 	"log"
 	"net/http"
+	"os"
 )
 
 func handler(w http.ResponseWriter, r *http.Request) {
-	log.Printf("EventListener Pod received a request: %+v", r)
-	fmt.Fprintf(w, "Hello from the EventListener Pod! \n")
+	log.Printf("EventListener pod received a request: %+v", r)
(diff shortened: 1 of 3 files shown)


cmd/eventlistener/main.go
```diff
@@ -17,19 +17,65 @@ package main
...
+	return triggerBindings, nil
+}
+
 func handler(w http.ResponseWriter, r *http.Request) {
-	log.Printf("EventListener Pod received a request: %+v", r)
-	fmt.Fprintf(w, "Hello from the EventListener Pod! \n")
+	log.Printf("EventListener pod received a request: %+v", r)
+	triggerBindings, err := triggerBindings()
+	if err != nil {
+		log.Fatal(err)
```
ncskier: I don't think this should be a Fatal error. I don't think we want to have the Pod crash if a TriggerBinding cannot be found.

cmd/eventlistener/main.go
```diff
@@ -17,19 +17,65 @@ package main
...
+	}
+	for i := range triggerBindings {
```
ncskier: nit: we're looping through the `triggerBindings` twice here (because we already looped through the bindings in the `triggerBinding` function).

I think it would be better to get the TriggerBindings from the client here, and remove the `triggerBindings` function. So you could do something like:
```
for _, bindingRef := range eventListener.Spec.TriggerBindingRefs {
  tb, err := triggersClient.TriggersV1alpha1().TriggerBindings(bindingRef.Namespace).Get(bindingRef.Name, v1.GetOptions{})
  if err != nil {
    log.Printf("Error getting TriggerBinding %v", bindingRef)
  }
  ...
}
```

PR #61, opened 2019-08-05
Use examples as part of documentation. Add utility/test to keep docs …

vtereso:
Added example yaml files
Add tests to ensure valid yaml
Add utility to update docs and test to keep them fresh

# Changes

# Submitter Checklist

review them:

### DIFF
--- a/docs/triggerbindings.md
+++ b/docs/triggerbindings.md
@@ -5,24 +5,38 @@ These parameters are passed into the referenced `TriggerTemplate`.
 Although similar, the separation of these two resources was deliberate to encourage `TriggerTemplate` definitions to be reusable.
 Further, it is as this level that the service account is connected, which specifies what permissions the resources will be created (or at least attempted) with.
 
+<!-- FILE: examples/triggerbindings/triggerbinding.yaml -->
 ```YAML
 apiVersion: tekton.dev/v1alpha1
 kind: TriggerBinding
 metadata:
-  name: simple-pipeline-binding
-  namespace: default
+  name: pipeline-binding
+  namespace: tekton-pipelines
 spec:
-  templateRef:
-    name: simple-pipeline-template
-  serviceAccount: default
-  event:
-    class: cloudevent
-    type: com.github.push
-  params:
-    - name: gitrevision
-      value: $(event.head_commit.id)
-    - name: gitrepositoryurl
-      value: $(event.repository.url)
+  templatebindings:
+    - templateref:
+        name: pipeline-template
+      params:
+        - name: gitrevision
+          value: $(event.head_commit.id)
+        - name: gitrepositoryurl
+          value: $(event.repository.url)
+```
+```YAML
+apiVersion: tekton.dev/v1alpha1
+kind: TriggerBinding
+metadata:
+  name: pipeline-binding
+  namespace: tekton-pipelines
+spec:
+  templatebindings:
+    - templateref:
+        name: pipeline-template
+      params:
+        - name: gitrevision
+          value: $(event.head_commit.id)
+        - name: gitrepositoryurl
+          value: $(event.repository.url)
 ```
 
 One or more `TriggerBindings` are collected together into an [`EventListener`](eventlisteners.md), which is where the pod is actually instantiated that "listens" for the respective events.
--- a/hack/update-docs.sh
+++ b/hack/update-docs.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Tekton Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/library.sh
+cd ${REPO_ROOT_DIR}
+
+# Whether this script is being used for validation or not
+DIFF=
+# Syntax format for code fences
+SYNTAX="YAML"
+CODE_FENCE='^```'
+# Comment that will specify the file to embed within $SYNTAX code fence
+COMMENT='<!-- *FILE: *(.*) *-->'
+EMPTY='^[[:space:]]*$'
+
+# Parse flags
+while getopts ":d" opt; do
+  case ${opt} in
+    d )
+        DIFF="true"
+        ;;
+    * )
+        echo "Invalid Option: -$OPTARG" 1>&2
+        exit 1
+        ;;
+  esac
+done
+shift $((OPTIND -1))
+
+doc_files="$(find docs -name "*.md")"
+for file in ${doc_files};do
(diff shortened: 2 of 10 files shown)


vtereso:
The error for stale docs can be seen [here](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_triggers/61/pull-tekton-triggers-build-tests/1158756437564854274/)

cmd/webhook/main.go
```diff
@@ -100,7 +100,7 @@ func main() {
 			v1alpha1.SchemeGroupVersion.WithKind("TriggerTemplate"): &v1alpha1.TriggerTemplate{},
 		},
 		Logger:                logger,
-		DisallowUnknownFields: true,
+		DisallowUnknownFields: false,
```
EliZucker: what is the reason for changing this to false? Not saying it's wrong just trying to understand!

vtereso: The structure of the templated resources is arbitrary, therefore we cannot know what the structure of the data will be. [This](https://github.com/tektoncd/triggers/pull/61/files#diff-590359aadf35a5cd62893ca03297feeeR16) field (or any other) does not have an explicit mapping in the [`TriggerResourceTemplate`](https://github.com/tektoncd/triggers/commit/209ae98bbcf337d3c97eed5b7a124b1ff6052603), which would result in the admission webhook preventing any `TriggerTemplate` from being created.

vtereso: Oddly, the `DisallowUnknownFields` check seems to look for explicit mappings rather than simply checking if the json unmarshals correctly.

ncskier: Hm, I was able to create the `TriggerTemplate` from the examples with `DisallowUnknownFields: true`. But this was from the `master` branch, so maybe something else is different on this branch?

vtereso: I have no idea why it failed for me locally 😒 Happy to put it back

ncskier:
Can you please add a reference to the `hack/update-docs.sh` script in the [DEVELOPMENT.md](https://github.com/tektoncd/triggers/blob/master/DEVELOPMENT.md#iterating) file? Probably under the `Iterating` section

ncskier:
Also, can you please add some instructions to the `test/README.md` about running the e2e yaml tests; for example, Pipelines has [this](https://github.com/tektoncd/pipeline/tree/master/test#running-yaml-tests).

You might have to rebase with `master` to get the latest `test/README.md`.

hack/update-docs.sh
```diff
@@ -0,0 +1,89 @@
...
+        if [[ ${line} =~ ${FENCE_ON} ]];then
+            fenced="true"
+            continue
+        fi
+        if [[ ${line} =~ ${FENCE_OFF} ]];then
+            fenced="false"
+            continue
+        fi
+        # Write to replacement file
+        if [[ ${fenced} == false ]];then
```
ncskier: I think this will delete any code fences that do not have the matching comments with a file, because the line will not be copied over when `fenced=true`

hack/update-docs.sh
```diff
@@ -0,0 +1,89 @@
...
+            continue
+        fi
+        # Write to replacement file
+        if [[ ${fenced} == false ]];then 
+            # Copy line
+            echo "${line}" >> ${new_file}
+            # Inline file
+            if [[ "$line" =~ ${COMMENT} ]];then
+                echo '```'$SYNTAX >> ${new_file}
+                cat ${BASH_REMATCH[1]} >> ${new_file}
```
ncskier: nit: Should this report an error if the file can't be found?

vtereso: The `cat` utility should  output if any file does not exist like this:
```
cat fakefile1.md >> fakefile2.md
cat: fakefile1.md: No such file or directory
```

hack/update-docs.sh
vtereso: This was the way I designed it so it will always recreate the code fence when the comment is detected and then ignore the one that was already there.

test/presubmit-tests.sh
```diff
@@ -27,8 +27,11 @@ export DISABLE_MD_LINTING=1
 
 source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/presubmit-tests.sh
 
+
 function post_build_tests() {
     golangci-lint run
```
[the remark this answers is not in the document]
vtereso: Didn't think about this. Nice catch.

hack/update-docs.sh
vtereso: I see what you mean now. Pushed a fix.

hack/update-docs.sh
ncskier: Small nit here is that if the script errors, you'll be left with those `update-*.md` files. One change could be to use this cleanup function and trap on EXIT:
```
function cleanup {
  tmp_files=$(find docs -name "${TMP_PREFIX}*.md")
  for file in ${tmp_files}; do
    rm "${file}"
  done
}

trap cleanup EXIT
```

If you make this change, then the temp prefix should probably be something more unique than `update`; for example:
```
TMP_PREFIX='pleasedontusethistmpprefix'
```

docs/triggerbindings.md
```diff
@@ -5,24 +5,38 @@ These parameters are passed into the referenced `TriggerTemplate`.
...
+spec:
+  templatebindings:
+    - templateref:
+        name: pipeline-template
+      params:
+        - name: gitrevision
+          value: $(event.head_commit.id)
+        - name: gitrepositoryurl
+          value: $(event.repository.url)
 ```
```
ncskier: Oops, looks like there's a duplicate code fence here.

test/e2e-tests.sh
```diff
@@ -24,10 +24,12 @@ source $(dirname $0)/e2e-common.sh
 initialize $@
 
 header "Setting up environment"
-
 install_triggers_crd
 install_pipeline_crd
 
+header "Running yaml tests"
+$(dirname $0)/e2e-tests-yaml.sh || fail_test
```
ncskier: Let me know if I'm mistaken, but I think `fail_test` exits the script. I think it could be beneficial to show the results from both the yaml and regular e2e tests. So, can you please change this to store the failure in a variable, and then run `fail_test` after the other e2e tests have run as well (see [Pipelines](https://github.com/tektoncd/pipeline/blob/master/test/e2e-tests.sh#L42) for a reference).

test/e2e-tests-yaml.sh
```diff
@@ -0,0 +1,21 @@
...
+    # Apply TriggerBindings
+    for file in $(find ${REPO_ROOT_DIR}/examples/triggerbindings/ -name *.yaml | sort); do
+    kubectl ${op} -f ${file}
+    done
+
+    # Apply EventListeners
+    for file in $(find ${REPO_ROOT_DIR}/examples/eventlisteners/ -name *.yaml | sort); do
+    kubectl ${op} -f ${file}
+    done
+done
```
ncskier: I don't think that this test actually tracks the failures. If one resource fails to be created, then the script will still return a success. We can probably track a return code like elsewhere in this PR:
```
failed=0
...
kubectl ${op} -f ${file} || failed=1
...
(( failed )) && fail_test
success
```

ncskier:
Sorry @vtereso, you're going to have to make a new PR for this.

test/e2e-tests-yaml.sh
```diff
@@ -0,0 +1,21 @@
+source $(dirname $0)/e2e-common.sh
+
+set -o errexit
+set -o pipefail
+
+for op in apply delete;do
+    # Apply TriggerTemplates
+    for file in $(find ${REPO_ROOT_DIR}/examples/triggertemplates/ -name *.yaml | sort); do
+    kubectl ${op} -f ${file}
```
ncskier: nit: please indent this line

Issue #62, opened 2019-08-06
Proposal to rename TriggerBinding


EliZucker:
If we're including conditionals and headermatching in `TriggerBinding`, then I would propose renaming it to something like `TriggerSifter` (open to alternatives).

I suggest this because the Trigger**Binding** name was created to reflect how it **binds** values from the event payload onto the `TrigerTemplate`, however it will now also be doing filtering and headermatching. I think the name should change to reflect the fact that more processing is done than just binding values.

ncskier:
This is a good point, but I still don't mind the name `TriggerBinding` since we're binding event data to a set of output parameters (almost like a transformer). Plus there's conditional logic. So, I don't think it's _necessary_ to change the name, but if we find one that sounds better, I wouldn't be opposed to an update.

I don't really like `TriggerSifter` because I think it loses the idea of the `TriggerBinding` having output information. But in the same vein, maybe something like `TriggerRouter` or `EventRouter`?

Since we're trying to make the `TriggerBinding` more reusable, maybe we can use the word `Template` with something like `EventTemplate`?

bobcatfish:
I feel like these days `TriggerBinding` really _only_ does binding, it no longer does anything conditional (I don't think?) since we moved that functionality into [interceptors](https://github.com/tektoncd/triggers/blob/master/docs/eventlisteners.md#event-interceptors) so I think the name was good in the end and we could close this issue?

PR #63, opened 2019-08-06
Issue55 reusable bindings

### LINKED ISSUE #55: Proposal to move TriggerBinding-TriggerTemplate connection out of TriggerBinding definition

vtereso:
# Changes
Modify `EventListener`/`TriggerBinding` to promote `TriggerBinding` reuse. Without this change, the additions of conditions and headers within the `TriggerBinding` would have prevented them from being reusable. Now, input parameters can be specified from the `EventListener` such that `TriggerBindings` can be used between multiple `EventListeners`. 

Also, this PR connects `TriggerBindings` and `TriggerTemplates` in a one-to-one relationship. My reasoning is that when validation is added, it would be most clear to validate all outputParams from the `TriggerBinding` against all params in the `TriggerTemplate`. In this way, the template should likely do everything necessary rather than segment everything between different templates. Also, between templates there is no way to cross reference generated resources, which could be another point of confusion if this was supported.

UPDATE: As per the WG discussion, it seems we are aligned in the one-to-one `TriggerBinding`-`TriggerTemplate` approach.

Fixes #55 

# Submitter Checklist

review them:


vtereso:
@EliZucker how does this look?

vtereso:
Waiting on https://github.com/tektoncd/triggers/pull/61 so that I can update the examples/docs
/hold

vtereso:
This PR places connects a serviceAccount per trigger, but another approach could be to have a single serviceAccount for the entire EventListener.

README.md
```diff
@@ -14,8 +14,8 @@ Naturally, CI/CD events contain information that should:
...
 - [`TriggerTemplate`](docs/triggertemplates.md) - Templates resources to be created (e.g. Create PipelineResources and PipelineRun that uses them)
-- [`TriggerBinding`](docs/triggerbindings.md) - Instantiates resources in TriggerTemplate using event fields
-- [`EventListener`](docs/eventlisteners.md) - Wraps TriggerBinding(s) into an addressable endpoint (the event sink)
+- [`TriggerBinding`](docs/triggerbindings.md) - Validates events and extracts payload fields
+- [`EventListener`](docs/eventlisteners.md) - Connects `TriggerBindings` and `TriggerTemplates` into an addressable endpoint (the event sink)
```
bobcatfish: just spitballing, we could provide even more detail here, something like:

`Exposes an [addressable endpoint (event sink)](maybe link to what addressable is) . When hit, uses params extracted from payloads by the `TriggeringBinding` to instantiate resources specified in `TriggerRemplates`

README.md
```diff
@@ -14,8 +14,8 @@ Naturally, CI/CD events contain information that should:
 
 The Tekton API enables functionality to be seperated from configuration (e.g. [Pipelines](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md) vs [PipelineRuns](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md)) such that steps can be reusable, but it does not provide a mechanism to generate the resources (notably, [PipelineRuns](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md) and [PipelineResources](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#pipelineresources)) that encapsulate these configurations dynamically. Triggers extends the Tekton architecture with the following CRDs:
 - [`TriggerTemplate`](docs/triggertemplates.md) - Templates resources to be created (e.g. Create PipelineResources and PipelineRun that uses them)
-- [`TriggerBinding`](docs/triggerbindings.md) - Instantiates resources in TriggerTemplate using event fields
-- [`EventListener`](docs/eventlisteners.md) - Wraps TriggerBinding(s) into an addressable endpoint (the event sink)
+- [`TriggerBinding`](docs/triggerbindings.md) - Validates events and extracts payload fields
```
bobcatfish: I feel like the fact that we are describing the functionality of this with an "and" is a strong clue that this object now has two responsibilities, and I wonder if it would make sense to split this into two (one for validation, one for extracting payloads). @vtereso @ncskier @EliZucker @dibyom do you know if there are any existing issues where this is being discussed? if not I'm happy to make one :D

docs/eventlisteners.md
```diff
@@ -9,6 +11,10 @@ metadata:
...
   namespace: tekton-pipelines
 spec:
-  triggerbindingrefs:
-    - name: pipeline-binding
+  triggers:
+    - triggerBinding:
+        name: pipeline-binding
+      triggerTemplate:
+        name: pipeline-template
+      serviceAccountName: default
```
bobcatfish: putting `serviceAccount` here makes sense to me!

btw for some reason we call this `serviceAccount` in pipelines so would be good to use consistent naming if we can (e.g. in taskrun: https://github.com/tektoncd/pipeline/blob/master/docs/taskruns.md#service-account)

docs/triggerbindings.md
```diff
@@ -1,9 +1,8 @@
 # TriggerBindings
-As per the name, `TriggerBindings` bind events to `TriggerTemplates`.
-`TriggerBindings` build atop `TriggerTemplates` by enabling you to capture fields within an event payload and store them as parameters.
-These parameters are passed into the referenced `TriggerTemplate`.
-Although similar, the separation of these two resources was deliberate to encourage `TriggerTemplate` definitions to be reusable.
-Further, it is as this level that the service account is connected, which specifies what permissions the resources will be created (or at least attempted) with.
+As per the name, `TriggerBindings` bind against events/triggers.
+`TriggerBindings` enabling you to capture fields within an event payload and store them as parameters. 
+The separation of `TriggerBindings` from `TriggerTemplates` was deliberate to encourage reuse
+since a template can we well defined where the incoming data (whether event sources or data structure) might be variable.
```
bobcatfish: I think something is odd in the middle of this sentence, maybe "since a template can _be_ well defined"? 

i think the reason tho is really that you might be able to define one `TriggerBinding` that works for multiple `TriggerTemplates`?

docs/triggerbindings.md
```diff
@@ -13,30 +12,11 @@ metadata:
...
-      params:
-        - name: gitrevision
-          value: $(event.head_commit.id)
-        - name: gitrepositoryurl
-          value: $(event.repository.url)
+  outputParams:
+    - name: gitrevision
+      value: $(event.head_commit.id)
+    - name: gitrepositoryurl
+      value: $(event.repository.url)
```
bobcatfish: can we show `inputParams` and the other fields in this example as well, and document how they work?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -26,12 +27,25 @@ import (
...
 type EventListenerSpec struct {
-	TriggerBindingRefs []TriggerBindingRef `json:"triggerbindingrefs,omitempty"`
+	Triggers []Trigger `json:"triggers,omitempty"`
+}
+
+// +k8s:deepcopy-gen=true
+type Trigger struct {
+	TriggerBinding     TriggerBindingRef  `json:"triggerBinding"`
+	TriggerTemplate    TriggerTemplateRef `json:"triggerTemplate"`
+	ServiceAccountName string             `json:"serviceAccountName"`
```
bobcatfish: can we have docstrings for all the structs + attributes we're adding?

(looks like we need to turn on more linting maybe, there should be linting that catches when exported values don't have docstrings :thinking: )

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -26,12 +27,25 @@ import (
...
 }
 
 type TriggerBindingRef struct {
-	Name       string `json:"name,omitempty"`
-	Namespace  string `json:"namespace,omitempty"`
+	Name       string             `json:"name"`
+	APIVersion string             `json:"apiversion,omitempty"`
+	Params     []pipelinev1.Param `json:"params,omitempty"`
```
bobcatfish: hm do `params` need to be here? what does it mean to pass params to `triggerBinding`? (might be something that can be cleared up by comments + examples)

pkg/apis/triggers/v1alpha1/event_listener_validation.go
```diff
@@ -27,8 +27,8 @@ func (t *EventListener) Validate(ctx context.Context) *apis.FieldError {
 }
 
 func (s *EventListenerSpec) Validate(ctx context.Context) *apis.FieldError {
-	if len(s.TriggerBindingRefs) == 0 {
-		return apis.ErrMissingField("spec.triggerbindingrefs")
+	if len(s.Triggers) == 0 {
+		return apis.ErrMissingField("spec.triggers")
```
bobcatfish: should we validate the contents of the `triggers` also?

docs/triggerbindings.md
vtereso: Since we don't have conditionals/header matching yet, I omitted them for the time being.

pkg/apis/triggers/v1alpha1/event_listener_validation.go
vtereso: There is a separate issue open for validation so I didn't opt to tackle it here just yet. Also, since we are still in a bit of flux, didn't want this to be cumbersome.

README.md
vtereso: Aside from the [renaming issue](https://github.com/tektoncd/triggers/issues/62) and the [reusability issue](https://github.com/tektoncd/triggers/issues/55) (this issue), I don't believe there are. _Maybe_ they could be divided, but since the `TriggerBinding` is the sole place/resource that the event/payload is managed, I think it makes sense that it does both. This PR gives `TriggerBindings` separate what was previously params into input params so that matching can be done in a reusable way and output params, which are to be used for the `TriggerTemplate`. Alternatively, we could provide access to the headers emit two kinds of parameters from the `TriggerBinding` so that one could be used for validation and one for the `TriggerTemplate`. More or less, wherever the static values are colocated will not be a reusable component and since `EventListeners` are specific instantiations, I believe that was the premise behind this architecture (passing in the values to the `TriggerBinding`).

vtereso: Spoke with @ncskier, he made a point that this new conditional type could also have access to the event, where we would only need one params (the output params) in the `TriggerBinding`.

ncskier: I agree with @bobcatfish that it might be helpful to break the conditionals & header matching out of the EventBinding. I think it would be helpful to have a new issue for this.

I'll post my idea as a comment in the new issue, but for example, we could create a `TriggerCondition` resource which evaluates to `true`/`false`, and which determines whether or not the `TriggerBinding` + `TriggerTemplate` is executed/instantiated. This `TriggerCondition` could be a top-level resource with `TriggerTemplate` and `TriggerBinding` where all of these could be specified in the `EventListener`.

pkg/apis/triggers/v1alpha1/event_listener_types.go
vtereso: Adding `-E golint` to our lint command (`golangci-lint run`) _should_ allow us to get [this behavior](https://github.com/golang/lint/blob/master/lint.go), but when I tried it only complained about naming stutter errors. Maybe someone else wants to l

pkg/apis/triggers/v1alpha1/event_listener_validation.go
bobcatfish: kk sounds good!

pkg/apis/triggers/v1alpha1/event_listener_types.go
bobcatfish: kk! Ive opened https://github.com/tektoncd/plumbing/issues/64 in the plumbing repo so hopefully we can get this for both projects :D (and cli too?)

docs/triggerbindings.md
bobcatfish: hm id rather either:

- not include them in the types until we're ready to document them
- include them in the docs but not have the functionality (which is what we've done for most other types anyway)

my personal feeling is that before 0.1 it's okay to have our docs be aspirational - if we intend to have the functionality in 0.1, i think we can put it in the docs before it actually works (as long as it does before 0.1). but if we _dont_ intend to include it in 0.1, i dont think we should include the types in the code at all

README.md
bobcatfish: Okay I made https://github.com/tektoncd/triggers/issues/71 with my thoughts! I'd like to see the `TriggerCondition` option too @ncskier :D

docs/eventlisteners.md
bobcatfish: ugh but it's called `serviceAccountName` in the rest of kubernetes isn't it :S

image

maybe we should leave this and change pipelines after all

README.md
EliZucker: Resolving this since it branched off into its own issue.

docs/eventlisteners.md
EliZucker: Created relevant pipeline issue [here](https://github.com/tektoncd/pipeline/issues/1219) and resolving this conversation.

pkg/apis/triggers/v1alpha1/event_listener_validation.go
EliZucker: Resolving since separate issue

pkg/apis/triggers/v1alpha1/event_listener_types.go
EliZucker: For now, I have added docstrings for the existing public types.

README.md
EliZucker: I changed the readme to be more specific. Let me know if it isn't what you wanted!

vtereso:
@EliZucker Nice updates. 

I thought it would be nice to have one service account per trigger to separate the permissions, but actually only one service account can be used (at the moment) since each `EventListener` creates a single deployment (where the podTemplate takes one serviceAccount). ServiceAccountName as a result should live directly under spec. Also, since we are likely not using input params in the `TriggerBinding`, maybe params (as prior) is more appropriate than outputParams?

vtereso:
One other improvement could also be made to clean up the json tags:
`json:"triggerBinding"` -> `json:"binding"`
`json:"triggerTemplate"` -> `json:"template"`

EliZucker:
@vtereso All really good points, thanks. Should be updated now.

pkg/apis/triggers/v1alpha1/trigger_binding_validation.go
```diff
@@ -21,10 +21,12 @@ import (
 	"knative.dev/pkg/apis"
 )
 
+// Validate TriggerBinding.
 func (t *TriggerBinding) Validate(ctx context.Context) *apis.FieldError {
 	return t.Spec.Validate(ctx)
 }
 
+// Validate TriggerBindingSpec.
```
bobcatfish: if you feel like it, this is a place where it would make sense to say something like "there is nothing to validate in a TriggerBindingSpec b/c xyz"

cmd/eventlistener/main.go
```diff
@@ -34,15 +34,21 @@ func handler(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			log.Printf("Error getting TriggerBinding %s", trigger.TriggerBinding.Name)
+		}
+		log.Print(tb)
+		tt, err := triggersClient.TektonV1alpha1().TriggerTemplates(listenerNamespace).Get(trigger.TriggerTemplate.Name, v1.GetOptions{})
 		if err != nil {
-			log.Printf("Error getting TriggerBinding %v", bindingRef)
+			log.Printf("Error getting TriggerTemplate %s", trigger.TriggerTemplate.Name)
 		}
+		log.Print(tt)
```
bobcatfish: we're going to want to start adding unit tests for the eventlistener asap but np if you dont want to do it in this PR

pkg/apis/triggers/v1alpha1/event_listener_types.go
EliZucker: These were removed

pkg/apis/triggers/v1alpha1/trigger_binding_validation.go
EliZucker: This comment reminded me there actually is a bit of validation we can do... Added now.

bobcatfish:
Thanks for the commit message update!

Let's do this :sunglasses: 

image

/lgtm

Issue #48, opened 2019-07-30
Add Triggers information to Resources


vtereso:
# Expected Behavior
Any resources created through Triggers should be properly labelled and annotated to specify their origins.

# Actual Behavior
Currently, nothing does this.

# Additional Info
Documentation should be included.

vincent-pli:
I think it should not set `ownerReferences` if the resource defined in `triggerTemplate` already has one.  In such case, we suppose the resource has the garbage collection mechanism, we should keep it.

PR #70, opened 2019-08-14
Add EventListenerStatus: conditions for deployment/service readiness …

vtereso:
Add EventListenerStatus

# Changes
Add EventListenerStatus conditions to reflect Deployment status
Add EventListenerStatus condition for existence of all generated resources (service and deployment)
Generated resources (deployment/service) names now use K8s/Tekton naming generator to generate name, which is stored in EventListener Status
Add EventListenerStatus configuration for service (has domain name for now)
Add status subresources to Clusterrole (required to update status)
Change to wait polling so that it is easier to pass in new check functions

# Submitter Checklist

review them:


vtereso:
Since the EventListener (and the related `TriggerTemplate`/`TriggerBinding`) is pulled each time an event comes in, updates to the `EventListener` shouldn't require any reconciliation at the moment. The update function was removed from the `EventListener` reconciler so that status updates don't reconcile themselves. Alternatively, it could be kept in, but a conditional would guard against reverting the status.

akihikokuroda:
The status doesn't seem to reflect the status of the event listener pod. So events may not be received even if the status is OK.

vtereso:
> The status doesn't seem to reflect the status of the event listener pod. So events may not be received even if the status is OK.

I didn't want to recreate pod/deployment status unnecessarily so if someone wanted to know this, I think it would be more appropriate for them to reference it there. The EventListener status at the moment only goes as deep as the resources being started successfully.

config/200-clusterrole.yaml
```diff
@@ -29,6 +29,9 @@ rules:
   - apiGroups: ["tekton.dev"]
     resources: ["eventlisteners", "triggerbindings", "triggertemplates"]
     verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
+  - apiGroups: ["tekton.dev"]
+    resources: ["eventlisteners/status", "triggerbindings/status", "triggertemplates/status"]
+    verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
```
bobcatfish: For my own education: I thought the above lines (all verbs on eventlisteners, triggerbindings and triggertemplates) would be enough - looks like we do the same thing for pipelines https://github.com/tektoncd/pipeline/blob/master/config/200-clusterrole.yaml :O do you happen to know why we need to do this?

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -81,134 +83,132 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+										Value: el.Name,
+									},
+									corev1.EnvVar{
+										Name:  "LISTENER_NAMESPACE",
+										Value: el.Namespace,
+									},
+								},
+							},
+						},
+					},
```
bobcatfish: it looks like we have some opportunities to set a good precedent early here and break this into separate functions that can be unit tested, e.g.:

- A function to create the needed service, params would include the create function (`c.KubeClientSet.CoreV1().Services(service.Namespace).Create` (passing in functions like this makes unit testing way easier - that way you don't need tho whole reconciler and the unit test can just pass in a function that does whatever you like, in this case either succeeds or has an error)
- A similar function to create the Deployment
- Maybe even a function to get the status?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -35,7 +38,72 @@ type TriggerBindingRef struct {
...
+			Message: fmt.Sprintf("%s started successfully", t),
+		})
+	} else {
+		el.SetCondition(&apis.Condition{
+			Type:    t,
+			Status:  corev1.ConditionFalse,
+			Message: e.Error(),
+		})
+	}
+}
```
bobcatfish: its pretty trivial but id write a couple unit tests for these functions - they wouldnt do much right now but it should be super easy to do and they might be useful later!

config/200-clusterrole.yaml
vtereso: I haven't been able to find formal documentation on this, but since the status is a subresource, it is also "permissioned" like a resource. In this way, I think it makes sense to require separate RBAC permissions. However, it would make sense to have implicit access since the parent access has already been specified (maybe not?). I have never checked, but I assume that when grabbing the resource internally, the status would be populated where this seems buggier than behavioral.

dibyom:
Overall, looking good! There are a few places where we can refactor the code into functions for easier unit tests though I'd be ok with doing that in a cleanup CL.

> Open question: If resources fail to be updated somehow (currently Service/Deployment), what should the status be? I have left it as the success/true condition because the previous would still be ready? I could add another condition for pending updates maybe or set the corresponding condition to unknown/false?

Interesting. I think if there are pending updates, condition unknown makes sense. The previous deployment/service can still be ready but it does not match the updated EventListener spec, so to me it makes sense to keep the conditionReady to unknown/false?  I think we should only set it to condition true only if it matches the updated spec.

On a related note, if the resource update fails do we retry (or plan to retry?) 

> Any labels/etc. that will appropriate distinguish the generated resources could be removed (resource updates) where informers would likely check on these same labels. 

More of a question than a comment: Under what conditions can these labels be removed? e.g. is it only if a user manually removes the labels? Or are there other cases where this is possible?

vtereso:
> On a related note, if the resource update fails do we retry (or plan to retry?)

The reconciler should continually retry until it successfully updates.

> Under what conditions can these labels be removed? e.g. is it only if a user manually removes the labels? Or are there other cases where this is possible?

I don't know your definition of manual, but any service account/user with the necessary permissions should be able to do this either manually or programmatically. Maybe it is unlikely, but entirely possible. We could implement a status that doesn't handle this corner case.

dibyom:
> The reconciler should continually retry until it successfully updates.

Cool. Makes sense. Then maybe the two types of conditions like you proposed makes sense.  Deployment does something similar (I think the conditions are readiness and progressing?).

> Any service account/user with the necessary permissions should be able to do this either manually or programmatically. Maybe it is unlikely, but entirely possible. We could implement status to "correctly" reflect the other resource status where it doesn't handle this corner case.

Yeah, I think not handling the corner case is fine for now. Maybe in the future, we can do something more involved like the admission webhook you mentioned!

vtereso:
@dibyom So the approach you prefer is to also make an informer for the deployment/service and reflect their status right? Adding labels to correctly identify the resources is a separate issue (https://github.com/tektoncd/triggers/issues/48), but I could also do here?

dibyom:
Yes! And I think maybe keeping the labels change in a different PR makes sense?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -61,8 +63,41 @@ type TriggerTemplateRef struct {
...
 
-// EventListenerStatus defines the observed state of EventListener
-type EventListenerStatus struct{}
+const (
+	// ServiceExists is the ConditionType set on the EventListener, which specifies Service existence.
+	ServiceExists apis.ConditionType = "Service"
+	// DeploymentAvaiable is the ConditionType set on the EventListener to reflect the actual DeploymentAvailable status.
+	DeploymentAvailable apis.ConditionType = apis.ConditionType(appsv1.DeploymentAvailable)
+	// DeploymentProgressing is the ConditionType set on the EventListener to reflect the actual DeploymentProgressing status.
+	DeploymentProgressing apis.ConditionType = apis.ConditionType(appsv1.DeploymentProgressing)
```
ncskier: Do we need to account for `DeploymentReplicaFailure`? (https://godoc.org/k8s.io/api/apps/v1#DeploymentConditionType)

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -60,8 +63,72 @@ type EventListenerTemplate struct {
...
+// This does not use the InitializeCondition() provided by the conditionsImpl to avoid setting the happy condition.
+// This is a local change and needs to be persisted to the K8s API elsewhere.
+func (els *EventListenerStatus) InitializeConditions() {
+	for _, condition := range []apis.ConditionType{
+		ServiceExists,
+		DeploymentExists,
+	} {
+		els.SetCondition(&apis.Condition{
+			Type:   condition,
+			Status: corev1.ConditionFalse,
```
ncskier: > InitializeConditions will set all conditions in eventListenerCondSet to unknown for the EventListener.

I think this should be `core.v1ConditionUnknown` like it says in the comment. Is the comment or the code wrong? 😁

vtereso: I will update the docstring to false

test/builder/eventlistener.go
```diff
@@ -86,6 +91,35 @@ func EventListenerTriggerParam(name, value string) EventListenerTriggerOp {
 	}
 }
 
+// EventListenerSpec sets the specified status of the EventListener.
```
ncskier: typo: `EventListenerSpec` -> `EventListenerStatus`

test/builder/eventlistener_test.go
```diff
@@ -46,6 +50,92 @@ func TestEventListenerBuilder(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			builder: EventListener("name", "namespace",
+				EventListenerStatus(
+					EventListenerCondition(
+						v1alpha1.ServiceExists,
+						corev1.ConditionTrue,
+						"Service exists", "",
```
ncskier: nit: Please add a `reason` in these tests. I think we want to verify that the builder fills all of the required `EventListenerCondition` fields.

test/controller_test.go
```diff
@@ -139,8 +139,8 @@ func TestGetTestResourcesFromClients(t *testing.T) {
 				t.Fatal(err)
 			}
 			if !reflect.DeepEqual(*actualTestResources, tt.testResources) {
-				t.Errorf("GetTestResourcesFromClients = %v, expected %v", *actualTestResources, tt.testResources)
-				t.Log(cmp.Diff(*actualTestResources, tt.testResources))
+				t.Errorf("GetTestResourcesFromClients() Fail")
+				t.Logf("Diff request body: -want +got: %s", cmp.Diff(tt.testResources, *actualTestResources))
```
ncskier: nit: I think we should probably change the `reflect.DeepEqual()` above to be `cmp.Diff()` to simplify this.

pkg/apis/triggers/v1alpha1/event_listener_types_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+func TestSetGetCondition(t *testing.T) {
+	tests := []struct {
+		name               string
+		conditions         []*apis.Condition
+		expectedConditions int
+	}{{
+		name:               "No conditions",
+		conditions:         []*apis.Condition{},
+		expectedConditions: 0,
```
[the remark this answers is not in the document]
dibyom: Or the remaining test cases to match the first one? :)

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -93,124 +117,172 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	oldService, err := c.KubeClientSet.CoreV1().Services(el.Namespace).Get(el.Name, metav1.GetOptions{})
+	switch {
+	case err == nil:
+		el.Status.SetExistsCondition(v1alpha1.ServiceExists, nil)
+		el.Status.Configuration.Hostname = listenerHostname(el.Name, el.Namespace)
+		updated := ensureLabels(&oldService.Labels, &labels)
+		updated = ensureLabels(&oldService.Spec.Selector, &labels) || updated
+		if updated {
+			_, err = c.KubeClientSet.CoreV1().Services(el.Namespace).Update(oldService)
+			if err != nil {
```
ncskier: Should we have a Status condition update when we fail to update the Service? Should `ServiceExists` update to `false`? Or is logging this error enough.

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -60,16 +64,16 @@ var _ controller.Reconciler = (*Reconciler)(nil)
...
 // converge the two.
-func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
+func (c *Reconciler) Reconcile(ctx context.Context, key string) (returnError error) {
 	c.Logger.Infof("event-listener-reconcile %s", key)
 	// Convert the namespace/name string into a distinct namespace and name
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		c.Logger.Errorf("invalid resource key: %s", key)
-		return nil
+		return err
```
dibyom: I think we should go back to returning nil (or an instance of PermanentError) here. Otherwise, we will continue to try to reconcile with the invalid resource key:https://github.com/tektoncd/triggers/blob/master/vendor/knative.dev/pkg/controller/controller.go#L330

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -93,124 +117,172 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	case err == nil:
+		el.Status.SetExistsCondition(v1alpha1.ServiceExists, nil)
+		el.Status.Configuration.Hostname = listenerHostname(el.Name, el.Namespace)
+		updated := ensureLabels(&oldService.Labels, &labels)
+		updated = ensureLabels(&oldService.Spec.Selector, &labels) || updated
+		if updated {
+			_, err = c.KubeClientSet.CoreV1().Services(el.Namespace).Update(oldService)
+			if err != nil {
+				c.Logger.Errorf("Error updating EventListener Service: %s", err)
+				reconcileError = werror.Wrap(reconcileError, err.Error())
```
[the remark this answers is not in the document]
dibyom: +1

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -60,8 +63,72 @@ type EventListenerTemplate struct {
...
+		})
+	}
+}
+
+// +k8s:deepcopy-gen=true
+type EventListenerStatus struct {
+	duckv1beta1.Status `json:",inline"`
+	// Configuration stores configuration for the EventListener service
+	Configuration EventListenerConfig `json:"configuration"`
```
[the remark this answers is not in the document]
vtereso: I recall some people mentioning adding other fields so I structured it like to add more in the future. There could be other stuff here like the service port, etc.

test/builder/eventlistener_test.go
vtereso: I just added test cases to ensure the builders match the expected output. Are you referring to an additional field in the anonymous struct?

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -93,124 +117,172 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
-			// Give the Deployment the same name as the EventListener
-			Name: el.Name,
-			// If our EventListener is deleted, then its Deployment should be as well
-			OwnerReferences: []metav1.OwnerReference{*el.GetOwnerReference()},
-			Labels:          labels,
-		},
-		Spec: appsv1.DeploymentSpec{
-			Selector: &metav1.LabelSelector{
+	oldService, err := c.KubeClientSet.CoreV1().Services(el.Namespace).Get(el.Name, metav1.GetOptions{})
+	switch {
```
dibyom: Can we have two functions, say reconcileService and reconcileDeployment that take in the old and new resources? Will make it easy to write targeted unit tests in addition to the reconciler tests

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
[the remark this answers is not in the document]
vtereso: `github.com/pkg/errors` takes an error and returns an error
`golang.org/x/xerrors` formats and returns an error
The error could easily be converted into a string so that we could use xerrors. Ultimately, the only intention here is to wrap the error so that one big one is returned, but since everything is logged, it's not entirely necessary. As long as _some_ error is returned, the reconciler will retry.

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
vtereso: If the service fails to update (same as the deployment), the underlying resource still exists. We could add another condition that we add when updates fail?

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -93,124 +117,172 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	case err == nil:
+		el.Status.SetExistsCondition(v1alpha1.ServiceExists, nil)
+		el.Status.Configuration.Hostname = listenerHostname(el.Name, el.Namespace)
```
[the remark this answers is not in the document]
vtereso: Within `Reconcile()`, I set the deep copy to an empty status and rebuild it to compare against. I was thinking about certain edge cases like if you were delete a condition on the `EventListener`, it wouldn't be able to get reconciled if it wasn't idempotent like this. I'll take a look and see if it can be simplified.

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
vtereso: So return nil, which will not be considered a reconcile failure?

PR #73, opened 2019-08-19
Triggers RBAC: Add Role/RoleBinding reconcile logic

### LINKED ISSUE #68: Triggers RBAC
# Expected Behavior
Users should specify a service account (or a role binding?) in the `EventListener` to be used for resources created as per specific `TriggerTemplate`.

# Actual Behavior
Currently, there is no implementation.

# Additional Info
Currently, the `EventListener` creates a `Deployment` and a load balancer `Service` (to reference the deployment), which enables the pods to be addressable/receive external traffic. The service account tied to the `EventListener` listener pod has permissions necessary to read Triggers resources, but how can it get the permissions necessary for creating resources defined in a particular `TriggerTemplate`, which are arbitrary?

vtereso:
…stener for deployments to access Triggers resources from their namespace. A role/rolebinding must be added to the service account specified in the EventListener because we need to ensure the privileges necessary to grab the Triggers resources without the EventListener deployment pod(s). The privileges necessary for creating the resources in the TriggerTemplate are assumed to already be bound to the service account.

# Changes
Add Role and RoleBinding reconciler logic
Centralize polling into single function using REST client

Fixes #68 

# Submitter Checklist

review them:


akihikokuroda:
The TriggerBindingRef has the namespace so the triggerbindings and triggertemplates can be defined in a different namespace from the one in the eventlistener.  The service account for the eventlistner should be bond to the clusterrole instead of role that has access to the triggerbindings and triggertemplates.

vtereso:
> The TriggerBindingRef has the namespace so the triggerbindings and triggertemplates can be defined in a different namespace from the one in the eventlistener. The service account for the eventlistner should be bond to the clusterrole instead of role that has access to the triggerbindings and triggertemplates.

That's true. The role only allows for the "get" operation so potentially it may be ok to use a `ClusterRole`/`ClusterRoleBinding`. Alternatively, we could remove the namespace fields within the references.

vtereso:
For now, I removed the namespace references. We can easily change this to use `ClusterRoles`/`ClusterRoleBindings` should that be more advantageous in some way.

pkg/rbac/roles.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// RolePostfix is the postfix added to the name of the role that will be created to review trigger resources
```
ncskier: nit: your other comment says `Triggers resources` instead of `trigger resources`. I think I prefer `Triggers resources` 🤷‍♂

pkg/rbac/roles.go
```diff
@@ -0,0 +1,75 @@
...
+	RolePostfix string = "-trigger-role"
+	// RoleBindingPostfix is the postfix added to the name of the role binding that will be bound to the EventListener service account
+	RoleBindingPostfix string = "-trigger-rolebinding"
+)
+
+// AddRoleBinding creates a role and role binding in the EventListener namespace for the EventListener's ServiceAccount to view Triggers resources.
+// Both the role and role binding have owner references on the EventListener.
```
ncskier: nit: `have owner references on the EventListener` sounds backwards to me, since the EventListener owns the Role & RoleBinding. Rephrase ideas: `have the EventListener in their list of owner references` or just `are owned by the EventListener`?

pkg/rbac/roles.go
```diff
@@ -0,0 +1,75 @@
...
+			rbacv1.Subject{
+				Kind:      "ServiceAccount",
+				Name:      eventListener.Spec.ServiceAccountName,
+				Namespace: eventListener.Namespace,
+			},
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "Role",
+			Name:     fmt.Sprintf("%s%s", eventListener.Spec.ServiceAccountName, RolePostfix),
```
ncskier: nit: `fmt.Sprintf("%s%s", eventListener.Spec.ServiceAccountName, RolePostfix)` is repeated from [here](https://github.com/tektoncd/triggers/pull/73/files#diff-ddd5c0a5fb9951edd4647bf70a8e9453R25). Might be nice to store this as a variable somewhere for reuse.

test/eventlistener_test.go
```diff
@@ -35,13 +37,26 @@ func TestEventListenerCreate(t *testing.T) {
...
+	t.Logf("Created ServiceAccount %s in namespace %s", sa.Name, sa.Namespace)
+
 	// Create EventListener
 	el, err := c.TriggersClient.TektonV1alpha1().EventListeners(namespace).Create(
 		&v1alpha1.EventListener{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "my-eventlistener",
 			},
 			Spec: v1alpha1.EventListenerSpec{
+				ServiceAccountName: "my-serviceaccount",
```
ncskier: I think the name you created above was `"default-sa"` instead of `"my-serviceaccount"`. Maybe just use `sa.Name`? 👍

vtereso:
@ncskier Good point. I'll use generateName rather than name so that each one is unique.

ncskier:
I think you could use the EventListener resource name with or instead of the ServiceAccount name.

Also, I just realized that I don't think you have accounted for the scenario when the user updates the ServiceAccountName in the EventListener resource. Is that correct? I don't remember seeing a unit test for it. I'm guessing that in this case we would want to keep the Role, but update or delete/create a new RoleBinding?

vtereso:
> I think you could use the EventListener resource name with or instead of the ServiceAccount name.
> 
> Also, I just realized that I don't think you have accounted for the scenario when the user updates the ServiceAccountName in the EventListener resource. Is that correct? I don't remember seeing a unit test for it. I'm guessing that in this case we would want to keep the Role, but update or delete/create a new RoleBinding?

I'll add this too. 👍

bobcatfish:
/hold

gonna hold off on this for 0.1 and revisit! https://github.com/tektoncd/triggers/issues/68#issuecomment-526367943

Issue #50, opened 2019-07-30
Add Support For Naming Modifiers


vtereso:
# Expected Behavior
Resources that are created by `TriggerTemplates` will clash if the names defined in the resources are static. Add some postfix/modifier (uuid/timestamp) to resources to prevent naming collisions.

Pipelines uses uuid's so it's probably best to follow that convention.

# Actual Behavior
Currently, nothing does this.

# Additional Info
If all resources get an automatic and unknown name modifier (like a postfix), it will be impossible to refer to them from within the `TriggerTemplate`. For this reason, it is important that the same mechanism that uniquely names resources also allows this (sort of uuid) to be exposed for usage throughout the `TriggerTemplate`.

This can be accomplished by exposing another interpolation variable within the `TriggerTemplate` or also extending the `TriggerTemplate` struct.

For example, a variable such as `${id}` (or any such name like `${unique}`) could be utilized throughout a `TriggerTemplate` as follows:
```
apiVersion: tekton.dev/v1alpha1
kind: TriggerTemplate
metadata:
  name: simple-pipeline-template
  namespace: default
spec:
  params:
    - name: gitrevision
      description: The git revision
      default: master
    - name: gitrepositoryurl
      description: The git repository url
    - name: namespace
      description: The namespace to create the resources
  resourcetemplates:
    - apiVersion: tekton.dev/v1alpha1
      kind: PipelineResource
      metadata:
        name: git-source${id.one}
        namespace: ${params.namespace}
        labels:
            triggertemplated: true
      spec:
        type: git
        params:
        - name: revision
          value: ${params.gitrevision}
        - name: url
          value: ${params.gitrepositoryurl}
    - apiVersion: tekton.dev/v1alpha1
      kind: PipelineRun
      metadata:
        name: simple-pipeline-run${id.two}
        namespace: default
        labels:
            triggertemplated: true
      spec:
        pipelineRef:
            name: simple-pipeline
        trigger:
          type: event
        resources:
        - name: git-source
          resourceRef:
            name: git-source${id.one}
```
In this way, some interpolation base (like `${id}`) would have fields that could be implicitly referenced throughout the template (maybe as valid [here](https://github.com/tektoncd/triggers/blob/master/pkg/parse/validate.go#L17)). In any case, this is similar to how `${params}` is functioning.

To be more explicit, we could add another `TriggerTemplateSpec` field like Params where users would specify exactly what unique field names they want to expose. For example:
```
spec:
  params:
    - name: gitrevision
[...]

Issue #64, opened 2019-08-07
House keeper for Pipeline related stuff


vincent-pli:
# Expected Behavior
When `event` received, deployment of `EventListener` will create at least one `PipelineResource` and one `Pipelinerun`, then those stuff will get out of hand of `Trigger`.
That’s caused a problem, after period, there will be tens of hundreds `pipeline` related stuff leaved.

So I think we need setup garbage disposal/house keeper mechanism for the situation.
I saw https://github.com/tektoncd/triggers/issues/48 , but it need delete `EventListener` to clean all stuff. I think `EventListener` is design for long term running.
# Actual Behavior
Currently, nothing for this

vtereso:
I agree that resource build up could/would become cumbersome, especially when we are generating resources based on events. Triggers is enabling users to create these resources simply, which only exacerbates this Tekton issue. Maybe an option could be added to the PipelineRuns/TaskRuns, where resources can be deleted after runs? In terms of handling arbitrary resource deletion, I don't know how much better we can do than `EventListener` `metav1.OwnerReference`-based removal (or some of the sort) or having some time based deletion.

vincent-pli:
Could we add `metav1.OwnerReference` of the resources (except `Pipelinerun`) to a `Pipelinerun`. I believe there is at least one `PipelineRun`, right? 

So the problem is when to delete the `Pipelinerun`
- Use `informer` of `Pipelinerun`, monitor the `Pipelinerun`, delete it when complete.
- Change the `Pipeline` refer by `PipelineRun`, add last task which could send `event` to `EventListener`, when receive event, delete related `Pipelinerun`

vtereso:
`PipelineResources` are created ahead of time compared to `PipelineRuns` (since they do not support being embedded currently?), where likely only some should have owner references. If a resource doesn't have a specific SHA/revisions/tags, it's potentially reusable so deleting it _might_ be bad? That aside, I guess it is possible to add a `metav1.OwnerReference` to referenced `PipelineResources`?

vtereso:
This has also been discussed here: https://github.com/tektoncd/pipeline/issues/544

vincent-pli:
**Proposal**:
1. Design new kind:`TriggerExecutor`
```
apiVersion: tekton.dev/v1alpha1
kind: TriggerExecutor
metadata:
  name: some-executor
spec:
   pipelineruns []*corev1.ObjectReference
```
[...]

PR #74, opened 2019-08-19
Create resource(s) from an event

ncskier:
# Changes
- When the EventListener sink pod receives an event, it will try to create the specified resources
  - Event data will be injected into the resources as specified by the TriggerBinding + TriggerTemplate
- Changed the EventListener sink's port to an environment variable
- Refactored the `parse` package
- Refactored the `eventlistener/main.go` file to move most of the logic into packages 

## Notes
I renamed the `cmd/eventlistener` -> `cmd/eventlistenersink`, because I would like to be verbose about the distinction between an EventListener custom resource, and the EventListener sink pod that Triggers creates to listen for events. Let me know if you have a better name for the sink pod, "EventListener sink pod" is quite a mouthful 😅 

Param values can be an Array or String ([param_types.go](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1alpha1/param_types.go#L68)). However, we only need the string value at the moment, and I don't think that we have discussed using an array parameter. So, this PR disregards the Array value, and assumes Params will be strings. Unless we find a use case for using an Array, we might want to consider creating our own param type that only uses strings in the future.

Until we complete the [RBAC](https://github.com/tektoncd/triggers/issues/68) and [naming](https://github.com/tektoncd/triggers/issues/50) issues, resources can only be create once and the Triggers ServiceAccount needs permission to create the resources.

# Submitter Checklist

review them:


pkg/sink/sink.go
```diff
@@ -0,0 +1,131 @@
...
+// createResource uses the kubeClient to create the resource defined in the
+// TriggerResourceTemplate and returns any errors with this process
+func createResource(kubeClient *kubeclientset.Clientset, rt json.RawMessage) error {
+	// Assume the TriggerResourceTemplate is valid (it has an apiVersion and Kind)
+	result := kubeClient.RESTClient().Post().
+		RequestURI(getRequestURI(rt)).
+		Body([]byte(rt)).
+		SetHeader("Content-Type", "application/json").
+		Do()
+	if result.Error() != nil {
```
vincent-pli: I'm think about if exception occurred when creating resource what should end user do?
- firstly, we should let user know error happen (update status of `eventListener`?)
- then, let user correct it then `apply eventListener` again?
if so, i think we need add something like:
`if apierrors.IsAlreadyExists(result.Error()) {` 
since we don't clean old resource when exception raised last time.

BTW, could we add `owner` for these resource?

pkg/template/params.go
```diff
@@ -0,0 +1,62 @@
...
+		rt = applyParamToResourceTemplate(param, rt)
+	}
+	return rt
+}
+
+// applyParamToResourceTemplate returns the TriggerResourceTemplate with the
+// param value substituted for all matching param variables in the template
+func applyParamToResourceTemplate(param pipelinev1.Param, rt json.RawMessage) json.RawMessage {
+	// Assume the param is valid
+	paramVariable := fmt.Sprintf("$(params.%s)", param.Name)
```
vtereso: Having the path regex equivalent of `eventPathVarRegex` would likely make this more performant/succinct.

pkg/sink/sink.go
```diff
@@ -0,0 +1,131 @@
...
+		log.Printf("Error getting EventListener %s in Namespace %s: %s", r.EventListenerName, r.EventListenerNamespace, err)
+		return
+	}
+
+	event, err := ioutil.ReadAll(request.Body)
+	if err != nil {
+		log.Printf("Error reading event body: %s", err)
+		return
+	}
+	log.Printf("Event: %s", string(event))
```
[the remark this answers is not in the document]
vtereso: Maybe we want to log resource templates as well before they are POSTed to the api server? Or remove this log?

pkg/template/event_test.go
```diff
@@ -0,0 +1,309 @@
...
+		Name:  "paramOneBogusEventPathVar",
+		Value: pipelinev1.ArrayOrString{StringVal: "bar-$(event.bogus.path)-bar"},
+	}
+	paramMultipleBogusEventPathVars = pipelinev1.Param{
+		Name:  "paramMultipleBogusEventPathVars",
+		Value: pipelinev1.ArrayOrString{StringVal: "bar-$(event.bogus.path)-$(event.two.bogus)-$(event.three.bogus)-bar"},
+	}
+	wantParamMultipleBogusEventPathVars = pipelinev1.Param{
+		Name:  "paramMultipleBogusEventPathVars",
+		Value: pipelinev1.ArrayOrString{StringVal: "bar-$(event.bogus.path)-$(event.two.bogus)-$(event.three.bogus)-bar"},
```
vtereso: So we will always opt to use String from the pipelinev1.ArrayOrString, but people could potentially still inject whatever the `$(event)` binding resolves to, which could be an array?

pkg/sink/sink_test.go
```diff
@@ -0,0 +1,54 @@
...
+		rt   json.RawMessage
+		want string
+	}{
+		{
+			name: "tekton.dev/v1alpha1",
+			rt:   json.RawMessage(`{"apiVersion": "tekton.dev/v1alpha1", "kind": "PipelineRun"`),
+			want: "apis/tekton.dev/v1alpha1/pipelineruns",
+		},
+		{
+			name: "tekton.dev/v1alpha1",
```
vtereso: Could you give this a different name to differentiate with the one above?

EliZucker:
I definitely like the `eventlistenersink` renaming! Also, if we don't want the array functionality, we may just want to use `ResourceParam` instead of `Param`, which has the string-only functionality you need (but so far that type has only been used in the context of resources).

pkg/sink/sink.go
```diff
@@ -0,0 +1,131 @@
...
+
+	for _, tbRef := range el.Spec.TriggerBindingRefs {
+		tbNamespace := tbRef.Namespace
+		if tbNamespace == "" {
+			tbNamespace = r.EventListenerNamespace
+		}
+		tb, err := r.TriggersClient.TektonV1alpha1().TriggerBindings(tbNamespace).Get(tbRef.Name, metav1.GetOptions{})
+		if err != nil {
+			log.Printf("Error getting TriggerBinding %s in Namespace %s: %s", tbRef.Name, tbNamespace, err)
```
EliZucker: I know we're logging debug messages a lot at this stage, but this line should eventually be `errorf`right? We could also propagate errors up and check if nil in main.

pkg/template/params.go
```diff
@@ -0,0 +1,62 @@
...
+	allParamsMap := map[string]pipelinev1.ArrayOrString{}
+	for _, paramSpec := range paramSpecs {
+		if paramSpec.Default != nil {
+			allParamsMap[paramSpec.Name] = *paramSpec.Default
+		}
+	}
+	for _, param := range params {
+		allParamsMap[param.Name] = param.Value
+	}
+	allParams := []pipelinev1.Param{}
```
EliZucker: Because we know that the size of `allParams` will invariably be the same size as `allParamsMap`, if we initialize the slice with that known size the slice won't have to repeatedly resized as we append elements; I think this line would be better: `allParams := make([]pipelinev1.Param, len(allParamsMap))`

docs/triggerbindings.md
```diff
@@ -21,6 +21,8 @@ spec:
           value: $(event.head_commit.id)
         - name: gitrepositoryurl
           value: $(event.repository.url)
```
akihikokuroda: Would you add the reference of the syntax of the path string?  It seems supporting this https://github.com/tidwall/gjson/blob/master/SYNTAX.md

pkg/sink/sink.go
ncskier: > I'm think about if exception occurred when creating resource what should end user do?

This is a good point about error handling. In the current implementation, an error would only be displayed in the pod logs. Since we haven't discussed this type of error handling in too much depth, I didn't want to implement anything too complicated in this PR. I like your idea of surfacing errors in the EventListener's status.

I think we should open a new issue to discuss and implement this error handling; should I open a new issue, or do you want to handle this @vincent-pli? 👍 

> then, let user correct it then apply eventListener again?
if so, i think we need add something like:
if apierrors.IsAlreadyExists(result.Error()) {
since we don't clean old resource when exception raised last time.

This is an interesting idea about "re-applying" failed resources. I'm not sure if it will be useful, but I think this indicates that we should have some way of keeping a record of the Resource with its injected event data. At the moment, I do [log the JSON of the resource](https://github.com/tektoncd/triggers/pull/74/files#diff-57fed74cee793226b40869dac1113d50R89) if it fails to be created. I'm not sure if there's anything else I can do in this PR to improve this issue.

> BTW, could we add owner for these resource?

I chose to leave the `owner` alone for the moment, because I'd like to tackle that in the [existing house keeper issue](https://github.com/tektoncd/triggers/issues/64). Plus, I think that once we're able to use the Triggers project, we will gain many insights into how we actually want to clean up the generated resources.

pkg/sink/sink.go
```diff
@@ -0,0 +1,131 @@
...
+		return result.Error()
+	}
+	return nil
+}
+
+// getRequestURI returns the URI for a post request to the kubernetes API
+// REST endpoint to create the TriggerResourceTemplate
+func getRequestURI(rt json.RawMessage) string {
+	// Assume the TriggerResourceTemplate is valid (it has an apiVersion and Kind)
+	apiVersion := gjson.Get(string(rt), "apiVersion").String()
```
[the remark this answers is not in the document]
ncskier: Thanks, good point 😅

pkg/sink/sink.go
```diff
@@ -0,0 +1,131 @@
...
+	kind := gjson.Get(string(rt), "kind").String()
+	namespace := gjson.Get(string(rt), "metadata.namespace").String()
+	var uri string
+	if apiVersion == "v1" {
```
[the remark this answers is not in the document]
ncskier: I referenced [this document](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups) where it says
> The core group, often referred to as the legacy group, is at the REST path /api/v1 and uses apiVersion: v1.

However, I could definitely be missing something, and I have only tested creating Triggers resources.

pkg/sink/sink.go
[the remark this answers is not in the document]
ncskier: I can remove this log. I don't think it's super necessary.

Also, the resource templates are logged [here](https://github.com/tektoncd/triggers/pull/74/files#diff-57fed74cee793226b40869dac1113d50R89-R92):
```
if err != nil {
	log.Printf("Error creating resources from resource template %s: %s", string(rtRaw), err)
	continue
}
log.Printf("Successfully created resource: %s", string(rt.RawMessage))
```

pkg/sink/sink.go
ncskier: Unfortunately, I think the `log` package only has `Printf` and `Fatalf`, there is no `Errorf`.

For propagating the errors, I think I propagate all of the errors up to the `HandleEvent` function in `sink.go`. However, this is a handler function, and is not invoked directly in `main.go`: `http.HandleFunc("/", r.HandleEvent)`, so I don't think I can propogate errors further up into `main.go`?

pkg/template/params.go
ncskier: Thanks, good point 👍

docs/triggerbindings.md
ncskier: I think this additional functionality is a good thing to keep in mind as we move forward. However, at the moment, I don't think we should officially "support" it. This is a first iteration, so I think it would be good to keep things simple for now. Plus, we don't want to grow too dependent on this library while we're still early on in the development process.

pkg/sink/sink.go
EliZucker: Ah, makes sense :+1:

pkg/template/event.go
```diff
@@ -0,0 +1,78 @@
...
+var eventPathVarRegex = regexp.MustCompile(`\$\(event(.[0-9A-Za-z_-]+)*\)`)
+
+// getEventPathFromVar returns the event path given an event path variable
+// $(event.my.path) -> my.path
+func getEventPathFromVar(eventPathVar string) string {
+	// Assume eventPathVar matches the eventPathVarRegex
+	if eventPathVar == "$(event)" {
+		return ""
+	}
+	eventPathRegex := regexp.MustCompile(`\$\(event.(.*)\)`)
```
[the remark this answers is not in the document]
ncskier: This regex was just used to trim off the variable syntax around the json path. But, it's kind of confusing when there's another similar regex. The implementation doesn't really matter, so I just changed the code to use `strings.TrimPrefix` and `strings.TrimSuffix` 👍

pkg/template/params.go
ncskier: I don't think I understand how a regex would be better here. Are you saying it would be much faster to create a regex with the `param.Name`, then use the `regexp` package for replacing values instead of `strings.Replace` ?

Or are you saying I can organize the code more succinctly. Maybe I could turn this line into a function: `fmt.Sprintf("$(params.%s)", param.Name)` ?

pkg/template/event_test.go
ncskier: Yep, that's my intention. I'm thinking whatever is injected, we will handle as a string (at least for now). This simplifies things as we get the code working for our first iteration. Also, JSON objects support [these types](https://www.w3schools.com/js/js_json_datatypes.asp), so if we directly support Arrays, then I think we'll also have to put some thought into supporting other types as well.

How to "type" the event data is something that I think we will have to figure out eventually, but for now I think Strings are ideal.

ncskier:
Thank you for the great feedback 🙂 

@EliZucker 
> Also, if we don't want the array functionality, we may just want to use ResourceParam instead of Param, which has the string-only functionality you need (but so far that type has only been used in the context of resources).

Thanks, I didn't know about this type 👍  Unfortunately, I couldn't find a matching `ParamSpec` that we could use. We need both `Param` & `ParamSpec` together, so it might be better to just create two new types of our own. What do you think, should I leave the code as it is right now or maybe create a couple of new types?

@vtereso 
> Within /pkg/template, event.go and params.go both deal with params.
The event.go file sets the params and params.go sets the TriggerTemplate.
Potentially better names?

Admittedly, I am not great at naming things. You're right that there are params in both files, so `params.go` is probably not the best name 😅 Do you have a suggestion for a better name? I could combine the files, but I wanted to keep the "event" templating separate.

> In any case, we should have an e2e test to ensure resources are created successfully.

You're right that we should have an e2e test, so I started writing one. But while I was writing the test I realized that it will not work without the [RBAC issue](https://github.com/tektoncd/triggers/issues/68). Since our e2e tests run in a new namespace, the EventListener sink pod will not have the correct permissions to get the Triggers resources and create the templated resources.

I have a branch [here](https://github.com/ncskier/triggers/tree/issue47-e2e) with my e2e test, which I can create a PR for after the RBAC PR is merged.

pkg/sink/sink.go
```diff
@@ -0,0 +1,130 @@
...
+
+	for _, tbRef := range el.Spec.TriggerBindingRefs {
+		tbNamespace := tbRef.Namespace
+		if tbNamespace == "" {
+			tbNamespace = r.EventListenerNamespace
+		}
+		tb, err := r.TriggersClient.TektonV1alpha1().TriggerBindings(tbNamespace).Get(tbRef.Name, metav1.GetOptions{})
+		if err != nil {
+			log.Printf("Error getting TriggerBinding %s in Namespace %s: %s", tbRef.Name, tbNamespace, err)
+			continue
```
khrm: Why are we just logging error and then continuing the iteration? Why not stop the iteration? Similar approach is being used below.

pkg/sink/sink.go
vincent-pli: I guess we need consider these issues, seems they are all corner cases but will sharply improve usability, I will log some issue to trace them, thanks

pkg/sink/sink.go
ncskier: I was doing this because the EventListener has a list of TriggerBindings which I thought should act independently of one another. Just because one of the TriggerBindings fails, it doesn't necessarily mean that the others will fail.

vtereso:
> Do you have a suggestion for a better name?
Maybe `event_binding.go`/`trigger_binding.go` and `event_template.go`/`trigger_template.go`? Something of that sort.

pkg/sink/sink.go
ncskier: Thanks!

ncskier:
Thanks for the ideas @vtereso, I went with renaming the `template/params.go` file to `template/resource.go`

cmd/eventlistenersink/main.go
```diff
@@ -0,0 +1,68 @@
...
+		TriggersClient:         triggersClient,
+		EventListenerName:      elName,
+		EventListenerNamespace: elNamespace,
+	}
+
+	// Listen and serve
+	log.Printf("Listen and serve on port %s", port)
+	http.HandleFunc("/", r.HandleEvent)
+	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
+}
```
bobcatfish: i like how we've got the sink package, separate from the logic here :D

I think we could even break this up and add some unit tests around it too - even tho there's not too much here (which makes it easier to break up :D):

- function to get the configuration (env var values)
- function to get the clients
- function to pass execution to the http handler

pkg/sink/sink.go
```diff
@@ -0,0 +1,130 @@
...
+				err = createResource(r.KubeClient, rtRaw)
+				if err != nil {
+					log.Printf("Error creating resources from resource template %s: %s", string(rtRaw), err)
+					continue
+				}
+				log.Printf("Successfully created resource: %s", string(rtRaw))
+			}
+		}
+	}
+}
```
bobcatfish: `HandleEvent` looks like a great candidate to break up into separate functions with unit tests - at the moment it would probably be difficult to write unit tests for `HandleEvent` will full coverage (I think we should probably have at least one tho!)

One thing that we do in Pipelines that (mostly) works is separating following all the references in a CRD from validating + from taking any action.

So it could be interesting to structure this function as something like (where each line is a separate function)

- Get all the CRD instances (i.e. follow all the Refs) needed -> probably return a new object that contains all the referenced objects together, e.g. slice of triggerbinding + triggertemplate
- Apply all the events (i.e. do the interpolation, `ApplyEventToParams`, `AddDefaultParamsFromSpecs`, `ApplyParamsToResourceTemplate`
- Create all the Resources

Ideally `HandleFunc` could tell us at a glance what all is being done each time its called, e.g. here's one idea of how it could look:
```go

type ResolvedBinding struct {
    TriggerBinding b
    TriggerTemplate t
}

func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Request) {
	el, err := r.TriggersClient.TektonV1alpha1().EventListeners(r.EventListenerNamespace).Get(r.EventListenerName, metav1.GetOptions{})
	if err != nil {
		log.Printf("Error getting EventListener %s in Namespace %s: %s", r.EventListenerName, r.EventListenerNamespace, err)
		return
	}

 	event, err := ioutil.ReadAll(request.Body)
	if err != nil {
		log.Printf("Error reading event body: %s", err)
		return
	}

        // Slice of ResolvedBinding
        bindings, err := ResolveBindings(el.Spec.TriggerBindingRefs, r.TriggersClient.TektonV1alpha1().TriggerBindings(tbNamespace).Get,
r.TriggersClient.TektonV1alpha1().TriggerTemplates(ttNamespace).Get)
      if err {}

       // Some object that contains the resources to create, maybe slice of json.RawMessage?
       resourcesToCreate, err := GetNewResources(bindings)
       if err {}

       if err := CreateResources(resourcesToCreate, kubeClient); err != nil { ... }<-- kubeClient interface looks v. hard to mock :S

```

Also note that a handy trick for functions that need to use the clients is to pass the functions around, e.g.

https://github.com/tektoncd/pipeline/blob/b8100856fedda3cb77f31b9487d093af5597fc7c/pkg/reconciler/taskrun/resources/taskresourceresolution.go#L38-L44

This lets unit tests create their own `GetResource` functions which is way less hassle than mocking clients.

So we'd be able to make unit tests (hopefully!) fairly easily for `ResolveBindings` and `GetNewResources` (and probably `CreateResources` too but I dunno how to mock out that RESTClient :thinking: )

pkg/sink/sink_test.go
```diff
@@ -0,0 +1,59 @@
...
+*/
+
+package sink
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+func Test_getRequestURI(t *testing.T) {
```
bobcatfish: nice! great candidate for unit tests :D

pkg/template/event.go
```diff
@@ -0,0 +1,77 @@
...
+		} else {
+			eventValue = gjson.GetBytes(event, eventPath).String()
+		}
+		if eventValue == "" {
+			return param, xerrors.Errorf("Error event path %s not found in the event %s", eventPath, string(event))
+		}
+		param.Value.StringVal = strings.Replace(param.Value.StringVal, eventPathVar, eventValue, -1)
+	}
+	return param, nil
+}
```
bobcatfish: i like these functions that are focused and each have a clear responsibility!

pkg/template/event.go
```diff
@@ -0,0 +1,77 @@
...
+	if eventPathVar == "$(event)" {
+		return ""
+	}
+	return strings.TrimSuffix(strings.TrimPrefix(eventPathVar, "$(event."), ")")
+}
+
+// ApplyEventToParams returns the params with each event path variable replaced
+//  with the appropriate data from the event.
+// ApplyEventToParams returns an error when the event path variable is not found
+// in the event.
```
bobcatfish: total nitpick: im not used to seeing the name of the function mentioned twice like this, with the error behavior described on a new line - if this is common in other projects im all for it but when i saw this first I thought that the first docstring lines were accidentally duplicated.

pkg/template/event.go
```diff
@@ -0,0 +1,77 @@
...
+	for _, eventPathVar := range eventPathVars {
+		eventPath := getEventPathFromVar(eventPathVar)
+		var eventValue string
+		if eventPath == "" {
+			eventValue = string(event)
+		} else {
+			eventValue = gjson.GetBytes(event, eventPath).String()
+		}
+		if eventValue == "" {
+			return param, xerrors.Errorf("Error event path %s not found in the event %s", eventPath, string(event))
```
bobcatfish: nice use of `xerrors` :D

pkg/template/event.go
```diff
@@ -0,0 +1,77 @@
...
+// with the appropriate data from the event.
+// applyEventToParam returns an error when the event path variable is not found
+// in the event.
+func applyEventToParam(event []byte, param pipelinev1.Param) (pipelinev1.Param, error) {
+	// Get each event path variable in the param
+	eventPathVars := eventPathVarRegex.FindAllString(param.Value.StringVal, -1)
+	for _, eventPathVar := range eventPathVars {
+		eventPath := getEventPathFromVar(eventPathVar)
+		var eventValue string
+		if eventPath == "" {
```
bobcatfish: maybe a comment here? i dont quite see why an empty `eventPath` means that we want to treat the value of `event` differently

pkg/template/event_test.go
```diff
@@ -0,0 +1,309 @@
...
+		if !tt.wantValid {
+			name = fmt.Sprintf("invalid: %s", tt.eventPathVar)
+		}
+		t.Run(name, func(t *testing.T) {
+			if valid := eventPathVarRegex.MatchString(tt.eventPathVar); valid != tt.wantValid {
+				t.Errorf("eventPathVarRegex.MatchString() = %t, want = %t", valid, tt.wantValid)
+			}
+		})
+	}
+}
```
bobcatfish: i think this is the first time ive seen unit tests for a regex but i like it!

pkg/template/event_test.go
```diff
@@ -0,0 +1,309 @@
...
+var (
+	testEventJSON       = json.RawMessage(`{"one": "onevalue", "two": {"two": "twovalue"}, "three": {"three": {"three": {"three": {"three": "threevalue"}}}}}`)
+	paramNoEventPathVar = pipelinev1.Param{
+		Name:  "paramNoEventPathVar",
+		Value: pipelinev1.ArrayOrString{StringVal: "bar"},
+	}
+	wantParamNoEventPathVar = pipelinev1.Param{
+		Name:  "paramNoEventPathVar",
+		Value: pipelinev1.ArrayOrString{StringVal: "bar"},
+	}
```
bobcatfish: this makes me wonder if the interface to `applyEventToParam` could be just values in and values out? would be easier to write unit tests for!

i.e. instead of passing a param into `applyEventToParam`, pass a string in, and get a string out

(cut it seems like the name of the param is never relevant to the test case)

pkg/template/event_test.go
```diff
@@ -0,0 +1,309 @@
...
+		{
+			// name: "error one eventpath not found",
+			args:    args{event: testEventJSON, param: paramOneBogusEventPathVar},
+			wantErr: true,
+			want:    wantParamOneBogusEventPathVar,
+		},
+		{
+			// name: "error multiple eventpath not found",
+			args:    args{event: testEventJSON, param: paramMultipleBogusEventPathVars},
+			wantErr: true,
```
bobcatfish: i am generally not a fan of using a boolean like `wantErr` in a table driven test like this - if a test case results in an error, the logic in the test is usually pretty different from the logic if it succeeds so it seems easier to write, read and maintain if instead we have a completely separate test for errors, so something like `Test_applyEventToParam` and `Test_applyEventToParam_errors`

pkg/template/event_test.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := ApplyEventToParams(tt.args.event, tt.args.params)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("ApplyEventToParams() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("ApplyEventToParams() = %v, want %v", got, tt.want)
```
bobcatfish: it can be really helpful to use `cmp.Diff` for this kinda assertion, so you can report exactly what was different instead of making the person running the test figure it out, e.g.:

```
if diff := cmp.Diff(got, tt.want); diff != "" {
    t.Errorf("Get PullRequest: -want +got: %s", diff) // it's helpful to include that "-want +got" bit cuz its hard to reason it out on your own
}
```

pkg/template/resource.go
```diff
@@ -0,0 +1,64 @@
...
+		allParamsMap[param.Name] = param.Value
+	}
+	allParams := make([]pipelinev1.Param, len(allParamsMap))
+	i := 0
+	for name, value := range allParamsMap {
+		allParams[i] = pipelinev1.Param{Name: name, Value: value}
+		i++
+	}
+	return allParams
+}
```
bobcatfish: often it turns out to be better in the longrun to separate mutating an object from any other logic, so something like:

- Get all the params with their provided values
- Get all the params with default values
- Merge the two

Or maybe that's pretty much what's happening and a different name might make that clear, e.g. `MergeInDefaultParams`

pkg/template/resource.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"strings"
+
+	pipelinev1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+)
+
+// AddDefaultParamsFromSpec returns the params with the addition of all
+// paramSpecs that have default values and are already in the params list
+func AddDefaultParamsFromSpecs(params []pipelinev1.Param, paramSpecs []pipelinev1.ParamSpec) []pipelinev1.Param {
+	allParamsMap := map[string]pipelinev1.ArrayOrString{}
```
bobcatfish: i wonder if a map would actually be a better structure for passing around the params + their values in general? (i.e. early in `HandleEvent`, we could convert params to a map)

test/presubmit-tests.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2018 The Knative Authors
+# Copyright 2019 The Tekton Authors
```
bobcatfish: hee hee

cmd/eventlistenersink/main.go
```diff
@@ -0,0 +1,68 @@
...
+		log.Fatal("EL_NAME env not found")
+	}
+	elNamespace := os.Getenv("EL_NAMESPACE")
+	if elNamespace == "" {
+		log.Fatal("EL_NAMESPACE env not found")
+	}
+	port := os.Getenv("PORT")
+	if port == "" {
+		log.Fatal("PORT env not found")
+	}
```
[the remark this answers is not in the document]
iancoffey: iirc - I started with cmd line args in the experimental listener, and moved everything to env. At the time, building the cli args felt hackier and I think I ran into the common string formatting issues, but that was then. Whatever works best now is good with me.

iancoffey:
This is looking good me. @bobcatfish has some great feedback already added - Ive little to add but am excited to see this land.

bobcatfish:
I was so focused on unit tests that I just realized that there's no end to end test with this one - would be great if we could include one :D :heart:

vincent-pli:
That's great work, just one suggestion,
the resource creating, could we consider to adopt this project later: https://github.com/jcrossley3/manifestival
this has been adopted by `Knative/serving-operator` and `tekton/operator`.

for now, it's not ready to our case, it only support parse resource definition from `file system` or `url`, but easy to enhance.

With this, we can easy to implements function like: owner injection, namespace injection or any logic for pre or post.

pkg/sink/sink.go
khrm: Got it. Thanks for the clarification.

pkg/sink/sink.go
```diff
@@ -0,0 +1,130 @@
...
+		SetHeader("Content-Type", "application/json").
+		Do()
+	if result.Error() != nil {
+		return result.Error()
+	}
+	return nil
+}
+
+// getRequestURI returns the URI for a post request to the kubernetes API
+// REST endpoint to create the TriggerResourceTemplate
```
khrm: nit: Shouldn't it be to `create the resource defined in the TriggerResourceTemplate `  instead of  `create the TriggerResourceTemplate`

pkg/sink/sink.go
```diff
@@ -0,0 +1,130 @@
...
+	var uri string
+	if apiVersion == "v1" {
+		uri = "api/v1"
+	} else {
+		uri = path.Join(uri, "apis", apiVersion)
+	}
+	if namespace != "" {
+		uri = path.Join(uri, "namespaces", namespace)
+	}
+	uri = path.Join(uri, fmt.Sprintf("%ss", strings.ToLower(kind)))
```
vtereso: I wonder if the resources we are created could/would have some plural form where we shouldn't just assume the plural just adds an 's' like we do here: https://github.com/tektoncd/pipeline/blob/master/config/300-clustertask.yaml#L21. Maybe we could make a request to obtain the plural form?

cmd/eventlistenersink/main.go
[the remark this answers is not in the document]
ncskier: I personally don't have much of a preference between the two. Since args are inherently documented by the usage details, they're a little more verbose. So, I'll change these to args 👍

pkg/template/event.go
ncskier: Thanks! 🙂

pkg/template/event.go
ncskier: Not a convention, and just had the same reaction when I glanced at the code snippet here 😆 I'll update it 👍

pkg/template/event.go
ncskier: I added a comment: `// $(event) has an empty eventPath, so use the entire event as the eventValue`

pkg/template/event_test.go
ncskier: Hm, that's a good point. We don't actually use the `.Names` anywhere 👍 
However, it will add the annoyance of converting from `[]pipelienv1.Param` -> `[]string` and then back again `[]string` -> `[]pipelinev1.Param`. I'll try it out 🙂

pkg/template/event_test.go
ncskier: Interesting, ok. Thanks for the advice 👍

pkg/template/event_test.go
ncskier: Thanks! Good point, that'll be a lot nicer to read 👍

ncskier: Although, if I'm not mistaken, we want to do `cmp.Diff(tt.want, got)` for `-want +got`?

pkg/template/resource.go
ncskier: Yes, I think `MergeInDefaultParams` would be a better name.

Are you also suggesting that I split this function into two functions? One that gets all the params with default values (turns `[]pipelinev1.ParamSpec` into `[]pipelinev1.Param`). Then the second which merges the `params []pipelinev1.Param` and `defaultParams []pipelinev1.Param`?

pkg/template/resource.go
ncskier: It's nicer in a lot of ways. I'll keep this in mind when addressing the other comments 👍

pkg/sink/sink.go
ncskier: Thanks, good catch 🙂

pkg/sink/sink.go
ncskier: True, I wonder how kubernetes does it. I assume I can make a request to the kube client to get the plural name of a resource, although I'm not sure what I would query. I'll look into this, but does anyone know off the top of their head?

ncskier:
Thanks for the awesome feedback! 🙂 

This build's broken by the latest PR, but I'll be working on fixing that along with addressing all of the feedback. 

> Speaking of commit messages, could you add some more context to the commit message?

I've updated the commit message; please let me know if it is still lacking 👍 

> I just realized that there's no end to end test with this one - would be great if we could include one :D

I'm working on an e2e test, although I'll need the [RBAC](https://github.com/tektoncd/triggers/pull/73) issue to be merged before I can test creating any resources other than Triggers resources.

pkg/sink/sink.go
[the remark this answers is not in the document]
vtereso: Just ran into this while working on RBAC. https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/rbac/v1/rbac_client.go Seems like only corev1 uses "/api"

ncskier:
Thanks for the feedback @bobcatfish, I really appreciate it! 🙂 I just pushed another update which:
- refactored `HandleEvent` into smaller functions with unit tests & integration tests
- put back the example JSON tests; now updated to fit the current implementation (these are now in `pkg/template/event_test.go`)
- updated the commit message 😁 
- separated error tests from non-error tests

Please let me know how this looks.

pkg/sink/initialization_test.go
```diff
@@ -0,0 +1,91 @@
...
+			}
+			if err := flag.Set("port", tt.port); err != nil {
+				t.Errorf("Error setting flag %s: %s", PORT, err)
+			}
+			if sinkArgs, err := GetArgs(); err == nil {
+				t.Errorf("GetArgs() did not return error when expected; sinkArgs: %v", sinkArgs)
+			}
+		})
+	}
+}
```
bobcatfish: yessss such test coverage 😎

bobcatfish:
/lgtm

image

iancoffey:
Wonderful! Great stuff

PR #85, opened 2019-08-28
Handle Referenced Resource Naming within TriggerTemplates

vtereso:
# Changes
Add `$(uid)` interpolation variable to be used within `TriggerTemplates` so that resources can cross reference one another (like PipelineResources and PipelineRun). 

# Submitter Checklist

review them:


vtereso:
I left out an integration test since I don't know of a good way to get the `$(uid)` to assert against. Open to suggestions.

docs/triggertemplates.md
```diff
@@ -1,8 +1,6 @@
 # TriggerTemplates
 A `TriggerTemplate` is a resource that can template resources.
 `TriggerTemplates` have optional parameters that can be substituted **anywhere** within the resource template.
-If resources do not have a name specified, it will default to value of the resource kind.
-Further, all resources names will have a unique timestamp postfix to avoid naming conflicts.
```
ncskier: Thanks for keeping the docs up to date! 🙂

docs/triggertemplates.md
```diff
@@ -48,11 +46,13 @@ spec:
         resources:
         - name: git-source
           resourceRef:
-            name: git-source
+            name: git-source-$(uid)
 ```
 
 Any parameters defined in the [`TriggerBinding`](triggerbindings.md) are passed into to the `TriggerTemplate` before resource creation.
+The `$(uid)` variable can be used through resource templates, which appends a postfix like the `generateName` metadata field, but allows the template be well defined.
```
ncskier: > ...which appends a postfix like the `generateName` metadata field, but allows the template be well defined.

nit: This is unintentionally slightly inaccurate, because the `$(uid)` variable does not `append a postfix`. Instead, the `$(uid)` is being substituted with a unique string. This can be used to append a postfix, but it can also be used for other purposes.

pkg/template/event.go
```diff
@@ -92,6 +92,8 @@ func getEventPathValue(event []byte, eventPath string) (string, error) {
 	return eventPathValue, nil
 }
 
+// NewResources returns all resources defined when applying the event to the
+// TriggerTemplate and TriggerBinding in the ResolvedBinding.
```
ncskier: Thanks for adding a descriptive comment to this public function 😄 Shouldn't our linter have caught this?

pkg/template/event_test.go
```diff
@@ -414,6 +416,8 @@ func Test_ApplyEventToParams_error(t *testing.T) {
 }
 
 func Test_NewResources(t *testing.T) {
+	// This seeds Uid() to return 'dpms2'
+	rand.Seed(0)
```
ncskier: Shouldn't this line be moved into the `t.Run` function, so that `rand` will be seeded at the beginning of every `t.Run` test case?

pkg/template/event_test.go
```diff
@@ -512,6 +516,37 @@ func Test_NewResources(t *testing.T) {
...
+								Value: pipelinev1.ArrayOrString{StringVal: "$(event.foo)", Type: pipelinev1.ParamTypeString},
+							},
+						},
+					},
+				},
+			},
+			want: []json.RawMessage{
+				json.RawMessage(`{"rt1": "bar-dpms2"}`),
+			},
+		},
```
ncskier: nit: can you please add a test case for multiple `$(ud)`s?

pkg/template/resource.go
```diff
@@ -85,3 +91,14 @@ func applyParamToResourceTemplate(param pipelinev1.Param, rt json.RawMessage) js
...
+
+// Uid generates a postfix like the k8s apiserver does for the generateName field.
+func Uid() string {
+	return rand.String(5)
+}
+
+// ApplyUIDToResourceTemplate returns the TriggerResourceTemplate after uid replacement
+// The same uid should be used per trigger to properly address resources throughout the TriggerTemplate.
+func ApplyUIDToResourceTemplate(rt json.RawMessage, uid string) json.RawMessage {
+	return bytes.Replace(rt, uidMatch, []byte(uid), -1)
```
ncskier: nit: since we're using `bytes.Replace()` here, can you please change the above
```
return []byte(strings.Replace(string(rt), paramVariable, param.Value.StringVal, -1))
```
to also use `bytes.replace()` for consistency?

pkg/template/resource_test.go
```diff
@@ -320,3 +321,13 @@ func Test_ResolveBinding_error(t *testing.T) {
 		})
 	}
 }
+
+func Test_ApplyUIDToResourceTemplate(t *testing.T) {
+	uid := Uid()
+	rt := json.RawMessage(`{"name": "something-$(uid)"}`)
```
ncskier: Can you please add more test cases? You can keep most of this code the same by using the table driven tests in the rest of this file. It would be great if you could add test cases for:
- no `$(uid)` (a regular string)
- multiple `$(uid)`s in one string

pkg/template/resource_test.go
```diff
@@ -320,3 +321,13 @@ func Test_ResolveBinding_error(t *testing.T) {
...
+	expectedRt := bytes.Replace(rt, uidMatch, []byte(uid), -1)
```
ncskier: Can you please write the `expectedRt` as a string like this:
```
json.RawMessage(`{"name": "something-dpms2"}`)
or maybe this:
json.RawMessage(fmt.Sprintf(`{"name": "something-$(%s)"}`, uid))
```
You probably shouldn't be verifying that your test works by copying the function's implementation code in your test case 😄  (`return bytes.Replace(rt, uidMatch, []byte(uid), -1)`)

pkg/template/resource_test.go
```diff
@@ -320,3 +321,13 @@ func Test_ResolveBinding_error(t *testing.T) {
...
+	actualRt := ApplyUIDToResourceTemplate(rt, uid)
+	if diff := cmp.Diff(string(expectedRt), string(actualRt)); diff != "" {
+		t.Error(diff)
```
ncskier: Please follow the logging format of the other tests explaining what the Diff means:
```
t.Errorf("ApplyUIDToResourceTemplate(): -want +got: %s", diff)
```
Without the `-want +got`, it can be confusing to understand what the diff string is showing.

ncskier:
Also, can you please update your commit message to be a little more descriptive? 😁 Maybe add some stuff from your PR description

pkg/template/event_test.go
vtereso: The library is only imported once and its internal state is seeded as specified. Seeding it again this way is probably a better practice though since the second value generated would be unknown (if `$(uid)` was used multiple times in the test).

pkg/template/resource_test.go
vtereso: For sure, this was before I seeded it. Forgot to change it.

pkg/template/event.go
vtereso: We need to update the extension in our linter. We have an issue tracking this, but I couldn't get it working for some reason.

ncskier: Ah, here it is https://github.com/tektoncd/plumbing/issues/64

pkg/template/resource.go
```diff
@@ -83,5 +88,16 @@ func ApplyParamsToResourceTemplate(params []pipelinev1.Param, rt json.RawMessage
 func applyParamToResourceTemplate(param pipelinev1.Param, rt json.RawMessage) json.RawMessage {
 	// Assume the param is valid
 	paramVariable := fmt.Sprintf("$(params.%s)", param.Name)
-	return []byte(strings.Replace(string(rt), paramVariable, param.Value.StringVal, -1))
+	return bytes.Replace(rt, []byte(paramVariable), []byte(param.Value.StringVal), -1)
+}
+
+// Uid generates a postfix like the k8s apiserver does for the generateName field.
```
ncskier: nit: this function generates a random string, not a postfix; the random string can be used as a postfix, but that's up to the user's discretion

maybe should be something like: `Uid generates a random string, like the postfix generated by the Kubernetes generateName metadata field.`

ncskier:
Thanks for making the changes Vincent!
/lgtm

Issue #45, opened 2019-07-30
Enable TriggerBindings to validate requests (e.g. GitHub)


vtereso:
# Expected Behavior
Events received by the `EventListener` addressable service should only be handled if the requester and request are authenticated. 

We can use GitHub as an initial use case.

From @khrm :

> Github computes the hash of payload using the above token and then pass the computed hash to the header.
> And on triggers side, if we want to be sure, we compute the hash of payload using that token(needs to be stored) and match it with the one received in the header.

# Actual Behavior
Currently, nothing does this.

We have discussed doing this via header matching:

> As per the WG discussion, add a `http` field with an underlying `headersMatch` field (as showcased in the [proposal](https://github.com/tektoncd/triggers/issues/42)) to the `TriggerBindingSpec`. The [deployment](https://github.com/tektoncd/triggers/commit/a4ff3311741341deb520ce98ff1c3c7c88c2381d) could intelligently map the received events to the correct `TriggerBinding(s)`, but simply iterating over the bindings is also good as a first pass. 

However as mentioned above, more is required than just header matching.

# Additional Info

@akihikokuroda included some example code at https://github.com/tektoncd/triggers/issues/78#issuecomment-523676106 to implement this.

vtereso:
Actually, not. Someone else feel free to tackle this. :)
/unassign

vtereso:
This probably should have been blocked since https://github.com/tektoncd/triggers/issues/47 and https://github.com/tektoncd/triggers/issues/55 are still open. @khrm how are you handling it in the meantime?

khrm:
I assign myself to work on this. Yet to start. Just following the discussions atm.

bobcatfish:
Hey @vtereso !

I'm late to the `headersMatch` discussion - can we add some details here (or even in a separate design doc) about what use cases we need the `headerMatch` field for and how it would work.

From the example in #42 it looked like the main use case was for filtering based on the type of event, e.g.:

```
  headerMatch:
    - X-GitHub-Event: issues
```

But then I asked @dlorenc and he said the most important use case was about verifying that requests actually come from GitHub i.e. matching a token in the header against a previously known token - it's not clear to me from the designs I've seen how we would do that (esp if the matching is just a simple regex as we are discussing in #49) so I'd love to see more details about that if possible.

vtereso:
[...]

PR #89, opened 2019-09-01
WIP: Enable TriggerBindings to validate requests

khrm:
This PR resolves the issue #45.
It assumes that a task has been defined which can validate requests.
That task will receive header and payload as params.
Before the creation of resources, task will be called alongwith serviceaccount
which has github-secret used to create webhook.
Assumption:
1. Task is defined in such a way that it can use headers and payload received as params.
2. Apart from serviceaccount, payload and headers, task doesn't need anything else.
3. Task gives us non zero exit if validation failed. A sample task and main.go is provided.

PS: This is still a work in progress.
# Changes

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```


cmd/securetrigger/main.go
```diff
@@ -0,0 +1,40 @@
+package main
```
khrm: Is this a correct location for this? Where do I put dockerfile for this? Maybe:
cmd/securetrigger/github/main.go  is better?

akihikokuroda:
Would you add functions that set params into the event listener and pass them to the validation task?  Otherwise the param values must be hard coded into the task and task is not reusable.

Use case: I specify the secret name (that has the secret key of the github) when I create the eventlistener and the secret name is passed to the validation task as one of the params so that the task can mount the secret.

khrm:
Good idea, will add that.

pkg/sink/sink.go
```diff
@@ -136,3 +155,64 @@ func createRequestURI(apiVersion, namePlural, namespace string) string {
...
+	params = append(params, trigger.TriggerValidate.Params...)
+	params = append(params, pipelinev1.Param{
+		Name: "Payload",
+		Value: pipelinev1.ArrayOrString{
+			Type:      pipelinev1.ParamTypeArray,
+			StringVal: string(payload),
+		},
+	})
+
+	for key := range headers {
```
dibyom: Are we adding each header as a param? If so, won't the validation task need to declare each header it expects as a declared parameter? If not, the task run will fail with a
`invalid input params: didn''t need these params but they were provided
        anyway: [header1]'`

pkg/sink/sink.go
```diff
@@ -136,3 +155,64 @@ func createRequestURI(apiVersion, namePlural, namespace string) string {
...
+		tr, err := r.PipelineClient.TektonV1alpha1().TaskRuns(r.EventListenerNamespace).Get(tr.Name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		if tr.IsSuccessful() {
+			break
+		}
+
+		time.Sleep(validateTaskWait)
```
dibyom: Hmmm, this seems a bit brittle. We are calling `secureEndpoint` for each trigger in a event listener spec in a for loop and then here we loop forever until the taskrun is done. 
So, if one of these validation taskruns takes a long time, other triggers will not be processed. Also, we are completely dependent on the TaskRun to behave properly (i.e. it either finishes successfully or times out properly). 

Some ideas:

* We process each trigger in its own goroutine so that one validation taskrun being slow does not block the others.

*  Instead of a for loop that goes on for ever, we should have some number of retries or give up after a timeout

* Add number of retries and a reasonable timeout value to these taskruns 

* In the future, instead of polling at fixed intervals, we could consider a informer/reconciliation based approach?

pkg/sink/sink.go
```diff
@@ -136,3 +155,64 @@ func createRequestURI(apiVersion, namePlural, namespace string) string {
...
+
+		if tr.IsDone() && !tr.IsSuccessful() {
+			return errors.New("validation taskrun failed")
```
dibyom: Would it be useful to add a reference to the taskrun somehow so that its possible to debug why the validation failed?

pkg/sink/sink.go
```diff
@@ -36,10 +42,15 @@ type Resource struct {
 	TriggersClient         triggersclientset.Interface
 	DiscoveryClient        discoveryclient.DiscoveryInterface
 	RESTClient             restclient.Interface
+	PipelineClient         pipelineclientset.Interface
 	EventListenerName      string
 	EventListenerNamespace string
 }
 
+const (
+	validateTaskWait = 20 * time.Second
```
dibyom: would something like `taskRunPollingInterval` be a more descriptive name for this value? 
(Also, this seems like a thing that we would want to make configurable in the future)

pkg/sink/sink.go
```diff
@@ -136,3 +155,64 @@ func createRequestURI(apiVersion, namePlural, namespace string) string {
 	uri = path.Join(uri, namePlural)
 	return uri
 }
+
+func (r Resource) secureEndpoint(trigger triggersv1.Trigger, headers http.Header, payload []byte) error {
```
dibyom: Nit. Consider renaming to `validateEvent` or something?

pkg/sink/sink.go
```diff
@@ -136,3 +155,64 @@ func createRequestURI(apiVersion, namePlural, namespace string) string {
...
+		params = append(params, pipelinev1.Param{
+			Name: key,
+			Value: pipelinev1.ArrayOrString{
+				Type:      pipelinev1.ParamTypeArray,
+				StringVal: headers.Get(key),
+			},
+		})
+	}
+
+	tr := &pipelinev1.TaskRun{
```
dibyom: Testing idea: If this was a separate function that took the event and returned a TaskRun spec, it would be simple to add focused unit tests that test that we generate the TaskRun with the correct spec.

pkg/sink/sink.go
```diff
@@ -136,3 +155,64 @@ func createRequestURI(apiVersion, namePlural, namespace string) string {
...
+
+func (r Resource) secureEndpoint(trigger triggersv1.Trigger, headers http.Header, payload []byte) error {
+
+	params := []pipelinev1.Param{}
+
+	params = append(params, trigger.TriggerValidate.Params...)
+	params = append(params, pipelinev1.Param{
+		Name: "Payload",
+		Value: pipelinev1.ArrayOrString{
+			Type:      pipelinev1.ParamTypeArray,
```
dibyom: Shouldn't this be ParamTypeString since we are using StringVal in the next line?

docs/validate-webhook.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  name: validate-webhook
+spec:
+  inputs:
+    params:
+      - name: Payload
+        description: Payload of Event Received
+      - name: X-Hub-Signature
+        description: Hash of the Request Received
+      - name: Github-Secret
```
dibyom: I don't see where this param is being used.

docs/validate-webhook.yaml
```diff
@@ -0,0 +1,27 @@
...
+      env:
+        - name: Payload
+          value: $(inputs.params.Payload)
+        - name: X-Hub-Signature
+          value: $(inputs.params.X-Hub-Signature)
+        - name: Github-Secret
+          valueFrom:
+           secretKeyRef:
+             name: github
+             key: github-secret
```
dibyom: Should this be hard-coded? Or is this something that comes in via the `Github-Secret` param above? I like the param approach since that way (with a default set) so that a user can override it via `TriggerValidate.Params`

cmd/securetrigger/main.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	payload := os.Getenv("Payload")
+	if len(payload) == 0 {
+		log.Println("Err securing endpoint: no payload")
+		os.Exit(1)
+	}
+
+	mac := hmac.New(sha1.New, []byte(secret))
+	_, _ = mac.Write([]byte(payload))
```
dibyom: Should we be ignoring the error here?

cmd/securetrigger/main.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"crypto/hmac"
+	"crypto/sha1"
+	"encoding/hex"
+	"log"
+	"os"
+)
+
+func main() {
```
dibyom: While the logic here is fairly simple, it would be nice to have some unit tests for this :angel:

cmd/securetrigger/main.go
dibyom: Maybe `cmd/validate-github`, `cmd/validate-gitlab`

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -36,10 +37,18 @@ type EventListenerSpec struct {
...
 // +k8s:deepcopy-gen=true
 type Trigger struct {
+	TriggerValidate *TriggerValidate   `json:"validate,omitempty"`
 	TriggerBinding  TriggerBindingRef  `json:"binding"`
 	TriggerTemplate TriggerTemplateRef `json:"template"`
 }
 
+// TriggerValidate represents the image to run taskrun for validating that trigger comes from the source which is desired
+type TriggerValidate struct {
+	TaskRef        *pipelinev1.TaskRef `json:"taskRef"`
```
dibyom: We should add some documentation around this. What Tasks are allowed e.g. they need to take a payload parameter and should not consume any resources etc.
 
In the future, we should validation for the Task to ensure it is a proper validation task (maybe instead of a Task we specify a specialized type which is a subset of a Task such as a `Validation` or `Filter`)

docs/validate-webhook.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: tekton.dev/v1alpha1
```
akihikokuroda: Instead of create the custom image with the custom executable, would you use a task something like this?
I haven't debugged  this but it should work with  only minor changes if necessary.

```
kind: Task
metadata:
  name: validate-webhook
spec:
  inputs:
    params:
      - name: Payload
        description: Payload of Event Received
      - name: X-Hub-Signature
        description: Hash of the Request Received
      - name: Github-Secret
        description: Secret used to configure webhook
  steps:
    - name: validate
      image: python
      command: ["/bin/bash"]
      args:
      - -ce
      - |
        set -e
        cat <<EOF | python
        import hashlib, os, hmac, six
        secret = os.environ.get('GithubSecret') 
        payload = $(inputs.params.Payload)
        signature = $(inputs.params.X-Hub-Signature)
        expected = hmac.new(secret, payload, hashlib.sha1).hexdigest()
        if expected is not None:
          if not isinstance(expected, six.text_type):
            expected = six.text_type(expected)
          sig_parts = signature.split("=", 1)
          if not isinstance(sig_parts[1], six.text_type):
            sig_parts1 = six.text_type(sig_parts[1])
          else:
            sig_parts1 = sig_parts[1]
          if len(sig_parts) > 1 and sig_parts[0] == "sha1" and hmac.compare_digest(sig_parts1, expected):
            exit(0)
        exit(1)
        EOF 
      env:
        - name: GithubSecret
          valueFrom:
           secretKeyRef:
             name: $(inputs.params.Github-Secret)

```

cmd/securetrigger/main.go
khrm: I am changing to shell-based task as suggested by akihikokuroda.

docs/validate-webhook.yaml
```diff
@@ -0,0 +1,27 @@
...
+    params:
+      - name: Payload
+        description: Payload of Event Received
+      - name: X-Hub-Signature
+        description: Hash of the Request Received
+      - name: Github-Secret
+        description: Secret used to configure webhook
+  steps:
+    - name: validate
+      image: quay.io/khrm/trigger-validate
```
vtereso: Maybe we would want to use a `ko` path?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -36,10 +37,18 @@ type EventListenerSpec struct {
 //
 // +k8s:deepcopy-gen=true
 type Trigger struct {
+	TriggerValidate *TriggerValidate   `json:"validate,omitempty"`
```
vtereso: For consistency, I think it makes sense to use a struct rather than a pointer.

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -36,10 +37,18 @@ type EventListenerSpec struct {
 //
 // +k8s:deepcopy-gen=true
 type Trigger struct {
```
vtereso: Since the `Trigger` type is changing, could you also update the `/test/builder` package?

ncskier:
I'd like to expand on @dibyom's comment [here](https://github.com/tektoncd/triggers/pull/89#discussion_r321430874) about passing header values to the validation Task.

Since we'll need to pass header values to the validation Task, I think we will need to add the headers as an interpolation variable, right? For example, `$(headers.X-Hub-Signature)`

Should we create a new issue for this? I don't want to pile too much work onto this PR.

pkg/sink/sink.go
vtereso: Only the `TriggerValidate` needs to be passed here.

vtereso:
> I'd like to expand on @dibyom's comment [here](https://github.com/tektoncd/triggers/pull/89#discussion_r321430874) about passing header values to the validation Task.
> 
> Since we'll need to pass header values to the validation Task, I think we will need to add the headers as an interpolation variable, right? For example, `$(headers.X-Hub-Signature)`
> 
> Should we create a new issue for this? I don't want to pile too much work onto this PR.

I think if we used an array type variable, it is sort of addressed?

docs/validate-webhook.yaml
khrm: It should come from GIthub-Secret above.

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -36,10 +37,18 @@ type EventListenerSpec struct {
...
+	TriggerValidate *TriggerValidate   `json:"validate,omitempty"`
 	TriggerBinding  TriggerBindingRef  `json:"binding"`
 	TriggerTemplate TriggerTemplateRef `json:"template"`
 }
 
+// TriggerValidate represents the image to run taskrun for validating that trigger comes from the source which is desired
+type TriggerValidate struct {
+	TaskRef        *pipelinev1.TaskRef `json:"taskRef"`
+	ServiceAccount string              `json:"serviceAccount,omitempty"`
```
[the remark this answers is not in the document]
khrm: OK. I name it that way due to Pipeline naming, will change it.

pkg/sink/sink.go
khrm: Yes.

docs/validate-webhook.yaml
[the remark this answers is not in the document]
khrm: These tasks would be specific to event sources like github, gitlab, etc. Should we keep them all in config/ directory?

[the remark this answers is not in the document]
dibyom: That's a good point. Maybe putting it in docs/ or examples/ is sufficient for now as long as the docs mention that the user has to install these!

pkg/apis/triggers/v1alpha1/event_listener_types.go
khrm: Sure.

pkg/apis/triggers/v1alpha1/event_listener_types.go
khrm: I did this because this is optional. I was using the pointer to check whether we want to validate or not.

docs/validate-webhook.yaml
khrm: I created this task with some minor changes and tested it. It works.

pkg/sink/sink.go
khrm: @dibyom What do you think of using a watch? For now, I have done using retries and taskRunPollingInterval.

khrm: Oh. And I created a function to execute trigger which runs under a goroutine.

pkg/sink/sink.go
khrm: I am using params from task and header to check this now.

pkg/sink/sink.go
```diff
@@ -53,25 +67,44 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
...
-		if err != nil {
-			log.Print(err)
-			continue
-		}
-		err = createResources(resources, r.RESTClient, r.DiscoveryClient)
-		if err != nil {
-			log.Print(err)
+		go r.executeTrigger(event, header, trigger, &wg)
+	}
+	wg.Wait()
```
dibyom: Why do we need to wait for each trigger to finish before we return from the handleEvent function? Can it not just call executeTrigger for each trigger and then return?

pkg/sink/sink.go
khrm: Did this.

docs/validate-github-webhook.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Github-Secret-Key
+        description: Secret key used to configured webhook
+  steps:
+    - name: validate
+      image: python:alpine
+      command: ["/bin/sh"]
+      args:
+      - -ce
+      - |
+        set -e
```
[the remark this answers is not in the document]
khrm: I do have written a taskrun to test this task. Maybe would push that also.

pkg/sink/sink.go
khrm: It does make more sense. The only reason I did that was because of the advice that I read that "Never start a goroutine without knowing how it will stop".  I feel we should quickly try to respond to the request but at the same time, I feel we should log somewhere that request came.

pkg/apis/triggers/v1alpha1/event_listener_types.go
khrm: I would create a separate pr may be for this. It has become too big.

docs/validate-github-webhook.yaml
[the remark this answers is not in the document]
dibyom: Sounds good!

pkg/sink/sink.go
dibyom: That seems like a good approach. Given that we retry for a fixed number of times in the goroutine, I'd argue that we know when it'd stop :smile: And yeah, responding to the request quickly was the main concern. 
In the far off future, we'd probably have to accept the event, queue it up for processing, and then process events off a queue!

pkg/sink/sink.go
```diff
@@ -53,25 +67,42 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
 		return
 	}
 
+	log.Printf("EventListener: %s in Namespace: %s handling event with payload: %s and header: %v",
+		r.EventListenerName, r.EventListenerNamespace, string(event), request.Header)
```
dibyom: nit: there is a log.Printf and a fmt.Printf a couple of lines below with the same info. We should remove one.

pkg/sink/sink.go
```diff
@@ -136,3 +167,96 @@ func createRequestURI(apiVersion, namePlural, namespace string) string {
...
+					Type:     v.Type,
+					ArrayVal: h,
+				},
+			})
+		} else if v.Type == pipelinev1.ParamTypeString {
+			params = append(params, pipelinev1.Param{
+				Name: v.Name,
+				Value: pipelinev1.ArrayOrString{
+					Type:      v.Type,
+					StringVal: h[0],
```
vtereso: Since the header key can be specified [multiple times](https://stackoverflow.com/questions/3096888/standard-for-adding-multiple-values-of-a-single-http-header-to-a-request-or-resp), maybe we should enforce the header to be passed an array always. Or at least conditionally check if there are multiple values?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -36,10 +37,18 @@ type EventListenerSpec struct {
 //
 // +k8s:deepcopy-gen=true
 type Trigger struct {
+	TriggerValidate *TriggerValidate   `json:"validate,omitempty"`
 	TriggerBinding  TriggerBindingRef  `json:"binding"`
 	TriggerTemplate TriggerTemplateRef `json:"template"`
 }
 
+// TriggerValidate represents the image to run taskrun for validating that trigger comes from the source which is desired
```
vtereso: nit: I feel like this could be more clear. At least at this level, there is no image.

pkg/sink/sink.go
```diff
@@ -136,3 +167,96 @@ func createRequestURI(apiVersion, namePlural, namespace string) string {
...
+	if err != nil {
+		return nil
+	}
+
+	tr, err = r.PipelineClient.TektonV1alpha1().TaskRuns(r.EventListenerNamespace).Create(tr)
+	if err != nil {
+		return err
+	}
+
+	for i := taskRunPollingRetries; i >= 0; i-- {
```
vtereso: As an improvement, we could probably use a watcher here, which has a deadline rather than polling.

pkg/sink/sink.go
```diff
@@ -136,3 +167,96 @@ func createRequestURI(apiVersion, namePlural, namespace string) string {
...
+		Name: "Payload",
+		Value: pipelinev1.ArrayOrString{
+			Type:      pipelinev1.ParamTypeString,
+			StringVal: string(payload),
+		},
+	})
+
+	for _, v := range task.Spec.Inputs.Params {
+		h, ok := headers[v.Name]
+		if !ok {
```
vtereso: I think we should check to see if there were default values within the task. Otherwise, this parameter will be never be passed and fail.

pkg/sink/sink.go
```diff
@@ -136,3 +167,96 @@ func createRequestURI(apiVersion, namePlural, namespace string) string {
...
+	}
+
+	if task.Spec.Inputs == nil {
+		return nil, errors.New("parameters are mandatory for validate task")
+	}
+
+	params := []pipelinev1.Param{}
+	params = append(params, triggerValidate.Params...)
+	params = append(params, pipelinev1.Param{
+		Name: "Payload",
```
vtereso: To further safeguard against parameter naming collisions/duplication, maybe it would be better to name this `TriggerPayload` or something like that?

pkg/sink/sink_test.go
```diff
@@ -359,28 +349,259 @@ func Test_HandleEvent(t *testing.T) {
...
+		EventListenerNamespace: "foo",
+	}
+
+	triggerValidate := &triggersv1.TriggerValidate{
+		TaskRef: pipelinev1.TaskRef{
+			Name: "bar",
+		},
+		ServiceAccountName: "foo",
+	}
+	task := pipelinetb.Task("bar", "foo", pipelinetb.TaskSpec(
```
vtereso: Thanks for using the pipeline builders 😎

pkg/sink/sink_test.go
```diff
@@ -359,28 +349,259 @@ func Test_HandleEvent(t *testing.T) {
...
+		return
+	}
+	wantPayload := `EventListener: my-eventlistener in Namespace: foo handling event with payload: {"head_commit": {"id": "testrevision"}, "repository": {"url": "testurl"}}`
+
+	if !strings.Contains(string(body), wantPayload) {
+		t.Errorf("Diff response body: %s, should have: %s", string(body), wantPayload)
+	}
+}
+
+func TestResource_validateEvent(t *testing.T) {
```
vtereso: I wonder if we should also be calling the [`TaskRun` validation](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1alpha1/taskrun_validation.go#L37) rather than just using reactor mocks.

pkg/sink/sink_test.go
```diff
@@ -359,28 +349,259 @@ func Test_HandleEvent(t *testing.T) {
...
+				ctx, _ := rtesting.SetupFakeContext(t)
+				clients, _ := pipelinetest.SeedTestData(t, ctx, pipelinetest.Data{
+					Tasks: []*pipelinev1.Task{pipelinetb.Task("bar", "foo", pipelinetb.TaskSpec(
+						pipelinetb.TaskInputs(
+							pipelinetb.InputsParamSpec("X-Hub-Signature", v1alpha1.ParamTypeString, pipelinetb.ParamSpecDescription("mydesc"), pipelinetb.ParamSpecDefault("default")),
+							pipelinetb.InputsParamSpec("X-Hub-Array", v1alpha1.ParamTypeArray, pipelinetb.ParamSpecDescription("desc"), pipelinetb.ParamSpecDefault("array", "values")))))},
+				})
+				return clients.Pipeline
+			}(),
+			want: &pipelinev1.TaskRun{
```
vtereso: I think it's fine to using test builders here as well. They have already been tested themselves as producing the same output as stubs.

pkg/sink/sink.go
dibyom: How about EventBody?

pkg/sink/sink.go
dibyom: I like the array idea though it will restrict where the header values can be used. Taskruns have validation to ensure arrays can only be used in certain fields so this might make the header params less usable: https://github.com/tektoncd/pipeline/blob/7a0454b27e9966ac6063b4c5320c46864fe33fc6/pkg/apis/pipeline/v1alpha1/task_validation.go#L205

vtereso: Joining the array on space or some other delimiter was what I meant behind flattening. If we are limited by arrays, maybe this is preferable?

khrm:
@dibyom I feel second option as better. I want to avoid resources. I am changing it to that.

dibyom:
@khrm cool, I like @vtereso 's idea as well:

> Joining the array on space or some other delimiter was what I meant behind flattening. If we are limited by arrays, maybe this is preferable?

I do agree at the moment, resources are probably not we want but later when we are done with all the cool resource extensibility stuff, it is something we could consider using

khrm:
> I do agree at the moment, resources are probably not we want but later when we are done with all the cool resource extensibility stuff, it is something we could consider using

Yes, let's wait for resource extensibility. For now, I have done json encoding. Will update the docs with that instruction.

pkg/sink/sink.go
khrm: log is being used to log here and fmt.Fprintf is being used to send a response to the event provider.

pkg/sink/sink.go
khrm: Yes. I would do that along with a way to configure timeout in a separate pr.

pkg/sink/sink_test.go
```diff
@@ -433,28 +421,267 @@ func Test_HandleEvent(t *testing.T) {
...
-		if diff := cmp.Diff(wantPipelineResourceURLPath, request.URL.Path); diff != "" {
-			t.Errorf("Diff request uri: -want +got: %s", diff)
-		}
-		body, err := ioutil.ReadAll(request.Body)
-		if err != nil {
-			return nil, err
-		}
-		if diff := cmp.Diff(string(wantPipelineResourceBytes), string(body)); diff != "" {
-			t.Errorf("Diff request body: -want +got: %s", diff)
-		}
```
ncskier: @khrm 
Sorry, this is pretty late, but why was this block of code removed? Now there's nothing checking that the correct payload was sent to the kubernetes api.

khrm: It wasn't thread safe. So I have removed it. And forgot to add a thread safe version. Sending a pr soon.
I had added response to check atleast that it reached till the end.

ncskier: Thanks! https://github.com/tektoncd/triggers/pull/121

PR #90, opened 2019-09-03
Recover deployment and SVC if deleted by accident

vincent-pli:
# Changes
Will recover `deployment` and `SVC` create by controller of `EventListener` if been deleted by accident.

# Submitter Checklist

review them:

### DIFF
--- a/pkg/reconciler/v1alpha1/eventlistener/controller.go
+++ b/pkg/reconciler/v1alpha1/eventlistener/controller.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	pipelineclient "github.com/tektoncd/pipeline/pkg/client/injection/client"
+	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	triggersclient "github.com/tektoncd/triggers/pkg/client/injection/client"
 	eventlistenerinformer "github.com/tektoncd/triggers/pkg/client/injection/informers/triggers/v1alpha1/eventlistener"
 	"github.com/tektoncd/triggers/pkg/reconciler"
@@ -28,6 +29,8 @@ import (
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/injection/clients/kubeclient"
+	deployinformer "knative.dev/pkg/injection/informers/kubeinformers/appsv1/deployment"
+	serviceinformer "knative.dev/pkg/injection/informers/kubeinformers/corev1/service"
 	"knative.dev/pkg/logging"
 )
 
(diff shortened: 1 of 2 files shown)


test/controller.go
```diff
@@ -50,6 +52,8 @@ type Clients struct {
 	Kube     *fakekubeclientset.Clientset
 	Triggers *faketriggersclientset.Clientset
 	Pipeline *fakepipelineclientset.Clientset
+	// deploy   *fakedeployinformer.Clientset
```
vtereso: These are commented, but created below?

pkg/reconciler/v1alpha1/eventlistener/controller.go
```diff
@@ -64,5 +68,17 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 		DeleteFunc: impl.Enqueue,
 	})
 
+	deploymentInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    impl.EnqueueControllerOf,
```
vtereso: This is using the same implementation as the EventListener? If that is the case, then any deployment/service created would trigger the Reconcile loop, but only the generated/children deployment/service should be reconciled?

vtereso: I suppose it works since the top of the reconciler has the following logic: 
```go
// Get the Event Listener resource with this namespace/name
	original, err := c.eventListenerLister.EventListeners(namespace).Get(name)
	if errors.IsNotFound(err) {
		// The resource no longer exists, in which case we stop processing.
		c.Logger.Infof("EventListener %q in work queue no longer exists", key)
		return nil
	} else if err != nil {
		c.Logger.Errorf("Error retreiving EventListener %q: %s", name, err)
		return err
	}
```

vincent-pli: Thanks for comments.
There is a mistake here, should use `cache.FilteringResourceEventHandler`, that the `deployment/svc` with incorrect owner will not go into the queue.
I have correct it, thanks again.

test/controller.go
vincent-pli: Sorry, forget to clear them.

vtereso:
As per dibyo's comment https://github.com/tektoncd/triggers/pull/70#issuecomment-527484417, we will need to create a reconciler for both the service and deployment to correcting update the status. Since this PR is already open, would you be able to make distinct reconcilers for both of these? I wouldn't want to override your work.

vincent-pli:
@vtereso 
A little confuse, could we plan to display the status of related `deployment` and `service` in status of `EventListener`? Any strong reason to do that? 
And we plan to reconciler `deployment` and `Service` in controller of `EventListener`? correct me thanks.

dibyom:
hi @vincent-pli I commented at https://github.com/tektoncd/triggers/pull/70#issuecomment-526287906 but the general idea is that the event listener status should reflect the status of the created deployment and service. We don't necessarily have to expose the entire status fields of deployments/services but some of them especially the status.Conditions can be useful. 

Say we create an event listener, and then we update it -- the event listener status should show that the we are moving to a new deployment. Otherwise, its hard for the user to know if the update is complete for not (they'd have to find the deployment, then track its status). Similarly, the ip address of the created service is something that the user might need in order to setup TLS/create the webhook from Github etc.

Hope that helps!

vincent-pli:
@dibyom 
Very helpful, thanks.
Could you merge #70 and this PR, I need work based on #70 and I don't want to put to much stuff in this PR, thank.

PR #100, opened 2019-09-06
Update WebhookTask

vtereso:
# Changes
Add links to Tasks
Break up the webhook task into an ingress task and webhook task for resuability
Remove static values within the ingress test
Add param to configure the GitHub webhook events
[Markdown Preview](https://github.com/vtereso/triggers/tree/LinkDocs/docs)

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```


docs/createwebhook.md
```diff
@@ -1,16 +1,163 @@
...
+      description: "Enables/disables the creation of the EventListener"
+      default: "true"
+    - name: CertificateKeyPassphrase
+      description: "Phrase that protects private key. This must be provided when the self-signed certificate is created."
+      default: "phrase1"
+    - name: CertificateSecretName
+      description: "Secret name for ingress certificate. The secret should not exist if the self-signed certificate creation is enabled."
+    - name: ExternalURL
+      description: "The external access URL for the EventListener."
+      default: "https://$(inputs.params.EventListenerName).<PROXYIP>.nip.io"
```
ncskier: Please add to the description that `ExternalURL` needs to be a valid IP address. Also, the `default` value here is not valid; plus, I don't think you can reference `$(inputs.params)` from inside the default param (but I could be wrong)

vtereso: Yeah, I think this is the only instance where the default is a valid default since `<PROXYIP>` wouldn't be known? I could push this into the description?

ncskier: The `<PROXYIP>` part is fine, but it shouldn't have `https://`. In the Task, it's used as `https://$(inputs.params.ExternalURL)`, so we don't want a double `https://`.

docs/README.md
```diff
@@ -2,8 +2,13 @@
...
-# Learn more
-See the following topics for more on each of the resources involved:
+
+## Learn More
+See the following links for more on each of the resources involved:
 - [`TriggerTemplate`](triggertemplates.md)
 - [`TriggerBinding`](triggerbindings.md)
 - [`EventListener`](eventlisteners.md)
+
+## Getting Started Tasks
```
iancoffey: Seems like we can remove this list, and just link to the getting started doc here instead here in that PR. What do you think?

iancoffey:
/lgtm 

Thank you for doing this, Ive begun adding these steps to the guide branch, will push that once this lands.

docs/README.md
vtereso: I agree!

Issue #101, opened 2019-09-09
TriggerTemplate resource should not need metadata.namespace


ncskier:
# Expected Behavior
When a templated resource in a TriggerTemplate is created without specifying a namespace, I expect the resource to be created in the same namespace as the EventListener which triggered this creation.

Currently, one needs to specify the namespace as a parameter in the TriggerTemplate for resources to be created properly.

# Actual Behavior
When a templated resource is created without specifying a namespace, the resource fails to be created with the following error:
```
PipelineResource.tekton.dev "git-resource-6chwh" is invalid: metadata.namespace: Required value
```

Example resource template:
```
- apiVersion: tekton.dev/v1alpha1
  kind: PipelineResource
  metadata:
    name: git-resource-$(uid)
    labels:
      triggers: "true"
      triggersUid: "$(uid)"
  spec:
    type: git
    params:
    - name: url
      value: $(params.url)
    - name: revision
      value: $(params.revision)
```

# Steps to Reproduce the Problem

1. Create a TriggerTemplate containing a resource template like the example above without a `metadata.namespace` filed.
2. Create a TriggerBinding and EventListener to use the TriggerTemplate
3. Send an POST request to the EventListener sink pod

PR #102, opened 2019-09-09
Default namespace for resource templates if not provided

vtereso:
# Changes
Fixes https://github.com/tektoncd/triggers/issues/101
This also allows us to create cluster scoped resources as prior the URI was incorrectly formed.

# Submitter Checklist

review them:


docs/triggertemplates.md
```diff
@@ -51,6 +51,8 @@ spec:
 
 Similar to [Pipelines](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md),`TriggerTemplates` do not do any actual work, but instead act as the blueprint for what resources should be created.
 Any parameters defined in the [`TriggerBinding`](triggerbindings.md) are passed into to the `TriggerTemplate` before resource creation.
+If the namespace is omitted, it will be resolved to the `TriggerTemplate` namespace.
```
ncskier: In my opinion, it makes most sense if this is the `EventListener`'s namespace (even though at the moment these are all in the same namespace)

ncskier: Also, thanks for keeping the docs up-to-date! 💯

ncskier:
Sorry, I think that I underspecified the issue, because we also have to take into account the case of cluster-wide resources; for example the [ClusterTask](https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md#clustertask). I don't think that this PR addresses cluster-scoped resources because it will add `metadata.namespace` to any resource without a `namespace`.

I am curious if there is a solution to this problem that does not involve changing the JSON/YAML at all. Looking through the Kubernetes documentation [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/), it looks like it should be possible to specify the `namespace` in the api path rather than in the POST data (but I could be misinterpreting this):
Screen Shot 2019-09-09 at 4 19 22 PM

vtereso:
Good point I will add this too /hold

pkg/sink/sink_test.go
```diff
@@ -188,13 +240,21 @@ func Test_createResource(t *testing.T) {
 	tests := []struct {
 		name        string
 		resource    json.RawMessage
+		namespace   string
```
ncskier: nit: please rename `namespace` this to `eventListenerNamespace`, because that is the name of the parameter in the `createResource` function.

Also, do you think it would make more sense to name the parameter `defaultNamespace` instead of `eventListenerNamespace`?

pkg/sink/sink_test.go
```diff
@@ -188,13 +240,21 @@ func Test_createResource(t *testing.T) {
...
 		wantURLPath string
 	}{
 		{
 			name:        "PipelineResource",
 			resource:    json.RawMessage(pipelineResourceBytes),
+			namespace:   pipelineResource.Namespace,
```
ncskier: nit: If I understand correctly, this namespace should probably be a different namespace than the one in the PipelineResource, because this namespace should be the namespace of the EventListener.

Can you also please add a test case for a resource created without its `metadata.namespace` field?

pkg/sink/sink.go
```diff
@@ -85,16 +85,20 @@ func createResources(resources []json.RawMessage, restClient restclient.Interfac
...
 // TriggerResourceTemplate and returns any errors with this process
-func createResource(rt json.RawMessage, restClient restclient.Interface, discoveryClient discoveryclient.DiscoveryInterface) error {
+func createResource(rt json.RawMessage, restClient restclient.Interface, discoveryClient discoveryclient.DiscoveryInterface, eventListenerNamespace string) error {
 	// Assume the TriggerResourceTemplate is valid (it has an apiVersion and Kind)
 	apiVersion := gjson.GetBytes(rt, "apiVersion").String()
 	kind := gjson.GetBytes(rt, "kind").String()
 	namespace := gjson.GetBytes(rt, "metadata.namespace").String()
-	namePlural, err := findAPIResourceNamePlural(discoveryClient, apiVersion, kind)
+	// Default the resource creation to the EventListenerNamespace if not found in the resource template
+	if namespace == "" {
```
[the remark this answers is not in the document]
vtereso: Namespace is an [omit empty value](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L147) where it should be the same. Also, this namespace value is just passed to `createRequestURI()`, where for cluster resources it will be ignored.

pkg/sink/sink_test.go
vtereso: Either one works for me. I'll use `eventListenerNamespace` for now since `defaultNamespace` seems less immediately clear/information lossy.

pkg/sink/sink.go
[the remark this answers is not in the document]
vtereso: Spoke with @ncskier more on this. Unless we check for this specifically in the `TriggerTemplate` validation, both approaches have pros and cons. Currently, if the resource template explicitly declares an empty namespace, this will also be overridden to the `EventListenerNamespaces`, otherwise we would be create a malformed URI.

pkg/sink/sink_test.go
```diff
@@ -186,14 +245,28 @@ func Test_createResource(t *testing.T) {
...
+		name                   string
+		resource               json.RawMessage
+		eventListenerNamespace string
+		wantURLPath            string
 	}{
 		{
-			name:        "PipelineResource",
-			resource:    json.RawMessage(pipelineResourceBytes),
-			wantURLPath: "/apis/tekton.dev/v1alpha1/namespaces/foo/pipelineresources",
+			name:                   "PipelineResource",
```
ncskier: quick nit: please update the test name, because now there're two named `PipelineResource`. One should probably say that it has a namespace, the other that it does not 👍

ncskier:
Thanks Vincent!! 🙂 
/lgtm

Issue #83, opened 2019-08-27
Restrict creation of Resources to just Pipeline Resources?


bobcatfish:
# Expected Behavior

After discussing with @IMJasonH and @dibyom I'm thinking it would make sense to restrict the types of Resources a user can create with TriggerTemplates to just types that we know are needed for executing Tekton Pipelines (e.g. `PipelineResources`, `Conditions`, `PipelineRuns`, etc.).

I think this makes sense because:

- It's easier to expand the scope later than it is to restrict it
- I don't think we have any concrete use cases for creating other Resources (correct me if I'm wrong); we know we want to start `PipelineRuns` (maybe `TaskRuns` too) but it's not clear we want to create anything else - and afaik we don't want to replace other generic event triggering solutions such as knative eventing or argo events
- If we do restrict the types, we can validate the structure of `TriggerTemplates` before waiting for the `EventListener` to be invoked

# Actual Behavior

We explicitly decided that it should be possible for a user to create _any_ Resource in a kube cluster, e.g. they could create a Deployment (or a Role!) as long as the service account had permission to.

vtereso:
I believe we already have the full scope since that is how it is implemented as per https://github.com/tektoncd/triggers/commit/c9537c0636218b51766dfc5dbab4db5bdbd0389d. If we wanted to restrict this to core tekton objects for the initial release, we could do so at the apiGroup/resource level.

ncskier:
Sorry, I'm a little late adding my opinion to this issue.

One benefit I see of not restricting TriggerTemplates to only Tekton resources, is that it decouples the Triggers project from the Pipelines project. So, if Pipelines creates new resources, or modifies its resources (which will happen in the future since it's still in alpha), the Triggers project might become broken from these changes, and/or Triggers might need to be updated to support these changes. I think the benefits of keeping the Triggers project decoupled or loosely coupled to the Pipelines project might outweigh the costs stated above.

vtereso:
From the WG discussion, we seem to be aligned that restricting resource creation to the `tekton.dev` apigroup makes sense for a first release. TBD on implementation. Should we add a `ConfigMap` whitelisting the permissible apigroups for creation? Use an environment variable?

dibyom:
I think a `ConfigMap` makes sense! I'll take a stab at this

/assign

vtereso:
@dibyom The root README specifies:
[...]

Issue #35, opened 2019-07-23
All existing types should have validation


bobcatfish:
# Expected Behavior

When the webhook notices that one of our trigger CRD types is created, it will invoke each type's validation. When that validation is invoked, it should look at the fields of each CRD and make sure they are valid.

# Actual Behavior

We have the skeleton webhook added, but our validation functions are empty, e.g.

https://github.com/tektoncd/triggers/blob/357fee5b70356dfd25eb26422310f9108c92fb85/pkg/apis/triggers/v1alpha1/trigger_template_validation.go#L24-L30

# Additional Info

This issue might not make any sense yet sine none of the types actually do anything, and we're still working on adding arbitrary event payloads in #19 so feel free to close this if I'm totally off base XD

vincent-pli:
@bobcatfish  
Can you assign this issue to me, I will add basic validation based on current design.

vtereso:
From the WG discussion:

With the typing structure of Triggers, following references will likely be necessary for validation. For example, parameter validation of an `EventListener` trigger will need to cross reference the corresponding `TriggerTemplate` and `TriggerBinding`. Our validation function does not have access to a client, but it would likely be the best place to store this logic. The admission controller would fast fail if the resources don't exist, which seems like the better option instead of having to parse some status/logs.  We can do this in the reconciler for a first pass if modifying the admission controller becomes cumbersome.

vincent-pli:
@vtereso 
😓 Missing the WG meeting.
Right, validation function is the right place to do that, I will check it and back here later.

vincent-pli:
To restrict that:
One `TriggerTemplate` should include at least one `Pipelinerun`, is that make sense?

vtereso:
@vincent-pli I don't think any verification of the params being used in the templates is necessary. In this way, I don't think we need to check for particular resources in the templates (e.g `PipelineRun`). What restrictions are you referring to?

vincent-pli:
Actually I want to find a signal which can express that: 
`The process triggered by the coming event has done, then please collect the garbage if necessary.`
I think the status of `Pipelinerun` could play the role, so I think the restriction is needed.

ncskier:
I agree with @vtereso. I don't think that we should force users to trigger PipelineRuns (event though in most cases they will be doing this).

[...]

PR #105, opened 2019-09-10
Restrict TriggerResourceTemplates to only Tekton Types

dibyom:
# Changes

This change restricts valid `TriggerResourceTemplate` to only Tekton types as discussed in #83. To do this,  the first commit adds validation for `TriggerTemplates` that validates that the `Spec` contains at least one `TriggerResourceTemplate` and that the `resourcetemplate` has the `apiVersion` and `Kind` fields set to one of the allowed types..

(At the moment, the allowed types are in code and will be made configurable via a configMap in a follow up PR)

This PR addresses #83 and #35 

# Submitter Checklist

review them:

# Release Notes

```
`TriggerResourceTemplate` can now be used to template only Tekton resources.
```


pkg/apis/triggers/v1alpha1/trigger_template_types.go
```diff
@@ -64,3 +65,10 @@ type TriggerTemplateList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []TriggerTemplate `json:"items"`
 }
+// GetApiVersionAndKind returns the apiVersion and Kind for the resourceTemplate
+// Missing fields are represented by empty strings
+func (trt *TriggerResourceTemplate) getApiVersionAndKind() (string, string) {
```
vtereso: Maybe we could use this within the the sink package `createResource()` as well?

dibyom: Yeah, definitely!

pkg/apis/triggers/v1alpha1/trigger_template_types.go
```diff
@@ -18,12 +18,26 @@ package v1alpha1
...
 )
 
+// apiVersion for Tekton core types
+const tektonApiVersion = GroupName + "/v1alpha1"
+
+// allowedTemplate types are the resource types identified by apiVersion + kind
+// that can be templated using TriggerResourceTemplates
+// TODO: Make this configurable using a ConfigMap entry
+var allowedTemplateTypes = map[string][]string{
+	tektonApiVersion: {"pipelineresource", "pipelinerun", "taskrun", "pipeline", "task", "condition"},
```
vtereso: There is no permission to create Triggers resources, which might be okay, but this is one reason the e2e tests are currently failing as we have a `TriggerTemplate` creating another template. Also note that creating a `TriggerTemplate` within a `TriggerTemplate` with the below validation (at least one resource) is a recursive problem. Also, the kinds within this value are not capitalized appropriately (the other reason validation is failing). We could use the strict capitalization or use `ToLower()`.

pkg/apis/triggers/v1alpha1/trigger_template_types.go
dibyom: gonna do this in a follow up. Since the function is on the resourcetemplate and not the triggerTemplate itself, we have to modify a `template.NewResources` as well as `createResource`

pkg/apis/triggers/v1alpha1/trigger_template_types.go
dibyom: used strings.ToLower() :man_facepalming:

So, the allowedKinds do not include TriggerTemplates at the moment which I guess is one way of avoiding the recursion problem

vtereso:
@dibyom Just adding another note here as well. https://github.com/dibyom/triggers/blob/65be1ab74a613d3341697a92bf1832b54ebad2b5/pkg/apis/triggers/v1alpha1/trigger_template_types.go#L36 did not specify the ability to create `ClusterTask`

Issue #114, opened 2019-09-12
EventListener Service type  should not be hardcoded


dibyom:
# Expected Behavior

Currently, the service object created by the EventListener controller is of type LoadBalancer. This is really nice for getting started quickly since (at least for GCP) you get a public IP that you can use to talk to the Event Listener sink very easily.

However, this may not be the behavior that everyone wants. Cluster operators/Tekton admins may have different requirements on how to expose their services e.g. use Ingress instead of a LoadBalancer Service. For instance, the webhook task in docs/ creates a ingress with self-signed certs.

Some options:

* We use the default Service type i.e. `clusterIP`. Users can their preferred  way to expose the service. (We'd update the docs and the webhook tasks in this case)

* Make this a configurable option. Users can select which Service type they want. Though this might be exposing too much of an implementation detail?

# Actual Behavior

Services are always of type LoadBalancer. 

# Additional Info

Service types: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types

dibyom:
Closed via #116

PR #116, opened 2019-09-12
Use clusterIP as the default service type

dibyom:
# Changes

Using `LoadBalancer` as the default service type
can have unintended effects e.g. in GCP it will
automatically expose a public IP.

Addresses #114


ncskier:
Thanks @dibyom !
/lgtm

PR #123, opened 2019-09-16
Move params into EventListener

### LINKED ISSUE #88: MakeTriggerBinding more reusable
# Expected Behavior

The TriggerBinding should be reusable for the same event and the same pipeline.

# Actual Behavior
> It should (maybe!) be possible for users to provide values to TriggerTemplate from sources other than events

from https://github.com/tektoncd/triggers/issues/87

Users can provide **static** values to the TriggerTemplate by putting them to the params in the TriggerBinding instead of the attribute values in the event.

This method makes the TriggerBindings non-reusable in the same cases.

- same task but the different credentials are required to upload the result to the repository.
- same task but the the result is uploaded in the different repository

For these cases, the user need to create different event EventListener and TriggerBinding for each credential or repository.  

# One Idea

Adding the capability to pass the param values from the EventListener to the TriggerBinding.  With this, the EventListener still have to be created for each case but the same TriggerBinding can be used for the both cases.

ncskier:
Fixes #88 Make TriggerBindings more reusable

This is the first pass at moving static parameters out of the
TriggerBinding. This implementation moves static parameters into the
EventListener where the user currently specifies all of the other static
information.

# Changes
- Move static parameters out of TriggerBinding and into EventListener
- Updated/created a parameters section in the EventListener, TriggerBinding & TriggerTemplate docs.
- Updated `examples/` with a working example of a static parameter in the EventListener
- Updated EventListener and TriggerBinding types with the new param workflow
  - Updated `builder` functions & tests to fit the updated types
- Updated EventListener in the `create-webhook.yaml` Task

New EventListener yaml (from `examples/`):
```YAML
kind: EventListener
metadata:
  name: listener
spec:
  triggers:
    - binding:
        name: pipeline-binding
        params:
          - name: message
      template:
        name: pipeline-template

```
New TriggerBinding yaml (from `examples/`):
```YAML
kind: TriggerBinding
metadata:
  name: pipeline-binding
spec:
  inputParams:
    - name: message
      description: The message to print
      default: This is the default message
  outputParams:
    - name: gitrevision
    - name: gitrepositoryurl
    - name: message
      value: $(inputParams.message)
```

If interested, please look over https://github.com/tektoncd/triggers/issues/88 for the discussion with proposed solutions to make TriggerBindings more reusable.

# Submitter Checklist

review them:


docs/triggerbindings.md
```diff
@@ -10,11 +11,37 @@ kind: TriggerBinding
 metadata:
   name: pipeline-binding
 spec:
-  params:
+  inputParams:
+    - name: message
+      description: The message to print
+      default: This is the default message
```
vtereso: Do you think it makes sense to have default within the `TriggerTemplate`? Since these _should_ be specified in the `EventListener`, maybe better illustrates the behavior that these should be passed into the `TriggerBinding`.

docs/triggerbindings.md
```diff
@@ -10,11 +11,37 @@ kind: TriggerBinding
...
     - name: gitrepositoryurl
       value: $(event.repository.url)
+    - name: message
+      value: $(inputParams.message)
 ```
 
 `TriggerBinding`s are connected to `TriggerTemplate`s within an [`EventListener`](eventlisteners.md), which is where the pod is actually instantiated that "listens" for the respective events.
+
+## Input Parameters
+`TriggerBinding`s can declare input parameters that must be supplied to the
```
vtereso: This "must be supplied" portion is sort of my philosophy behind my above comment.

docs/triggertemplates.md
```diff
@@ -15,6 +16,9 @@ spec:
       default: master
     - name: gitrepositoryurl
       description: The git repository url
+    - name: message
+      description: The message to print
+      default: This is the default message
```
vtereso: Maybe we want to remove default values here too?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -28,18 +28,23 @@ var _ apis.Validatable = (*EventListener)(nil)
...
-// +k8s:deepcopy-gen=true
-type Trigger struct {
-	TriggerValidate *TriggerValidate   `json:"validate,omitempty"`
-	TriggerBinding  TriggerBindingRef  `json:"binding"`
-	TriggerTemplate TriggerTemplateRef `json:"template"`
+// EventListenerTrigger represents a connection between TriggerBinding and TriggerTemplate;
+// TriggerBinding provides extracted values for TriggerTemplate to then create resources from.
+type EventListenerTrigger struct {
+	TriggerValidate    *TriggerValidate     `json:"validate,omitempty"`
+	Binding            EventListenerBinding `json:"binding"`
```
vtereso: Seems heterogeneous to have `templateRef.name` and `binding.bindingRef.name`. Could this be flattened one level?

pkg/apis/triggers/v1alpha1/trigger_binding_validation.go
```diff
@@ -30,11 +30,29 @@ func (t *TriggerBinding) Validate(ctx context.Context) *apis.FieldError {
...
+	}
+	return nil
+}
+
+func validateInputParams(params []v1alpha1.ParamSpec) *apis.FieldError {
+	// Ensure there aren't multiple params with the same name.
+	seen := map[string]struct{}{}
+	for _, p := range params {
+		if _, ok := seen[strings.ToLower(p.Name)]; ok {
+			return apis.ErrMultipleOneOf("spec.params")
```
vtereso: Should this be different between input and output params?

pkg/apis/triggers/v1alpha1/trigger_binding_validation.go
```diff
@@ -30,11 +30,29 @@ func (t *TriggerBinding) Validate(ctx context.Context) *apis.FieldError {
...
+	if err := validateInputParams(s.InputParams); err != nil {
+		return err
+	}
+	if err := validateOutputParams(s.OutputParams); err != nil {
+		return err
+	}
+	return nil
+}
+
+func validateInputParams(params []v1alpha1.ParamSpec) *apis.FieldError {
```
vtereso: This _could_ be recycled for validating input and output?

test/builder/eventlistener.go
```diff
@@ -60,19 +64,47 @@ func EventListenerServiceAccount(saName string) EventListenerSpecOp {
...
+					StringVal: value,
+					Type:      pipelinev1.ParamTypeString,
 				},
 			})
 	}
 }
+
+// Trigger creates an EventListenerTrigger. Any number of EventListenerTriggerOp
+// modifiers can be passed to create/modify it.
+func Trigger(tbName, ttName, apiVersion string, ops ...EventListenerTriggerOp) v1alpha1.EventListenerTrigger {
```
vtereso: Before the trigger was created within the `EventListenerTrigger()` function. This seems slightly confusing to me according to that function's docstring. Could you do the below:
```go
 		for _, op := range ops {
		op(&t)
	}
```
within that function?

docs/triggerbindings.md
ncskier: > Do you think it makes sense to have default values for inputParams within the TriggerTemplate?

(I'm assuming this meant `TriggerBinding`) This is a good point; I didn't really think about whether the user would need a `default` value, I used them for the sake of consistency. I worry that omitting `default` might be confusing if these parameter specs behave differently than the other parameter specs in Tekton.

Also, since the `default` value is optional, I think it is alright to include it.

docs/triggerbindings.md
ncskier: I wasn't sure how to word this, because the param is not necessary if there's a `default` value.

Should I rephrase it to:
`TriggerBinding`s can declare input parameters that must be supplied to the
`TriggerBinding` from an `EventListener` if no `default` value is specified.

docs/triggertemplates.md
ncskier: I think that you raise an important point to consider. It might not make sense to have a `default` value for Triggers parameters. However, I think that we should raise a separate issue to explore this idea. Under this issue, we might also want to consider creating our own Triggers Parameter types 👍

pkg/apis/triggers/v1alpha1/event_listener_types.go
ncskier: There needs to be one more layer for the TriggerBinding, because of the parameters being passed to the TriggerBinding at `binding.params`. I couldn't think of a good way to get around this asymmetry; do you have any ideas?

Flattening to one level makes it unclear if the parameters are passed to the TriggerBinding or TriggerTemplate:
```
bindingRef:
  name: pipeline-binding
templateRef:
  name: pipeline-template
params:
- name: message
```

vtereso: I was thinking 
```yaml
kind: EventListener
metadata:
  name: listener
spec:
  triggers:
  - binding:
      name: pipeline-binding
      params:
      - name: message
    template:
      name: pipeline-template
```

test/builder/eventlistener.go
ncskier: The `EventListenerTrigger()` function returned (and still does return) an `EventListenerSpecOp`. However, I wanted to use the builder functions to create an `EventListenerTrigger` in some of the lower-level unit tests that don't use an `EventListener` directly. So, I created the `Trigger()` function which is used as a helper in `EventListenerTrigger()`, and can also create `EventListenerTrigger` objects in the unit tests.

pkg/apis/triggers/v1alpha1/trigger_binding_validation.go
ncskier: I don't think it should be different because right now we're only checking that there aren't duplicates.

pkg/apis/triggers/v1alpha1/trigger_binding_validation.go
ncskier: I created a helper function to use for both

Issue #126, opened 2019-09-18
Params without defaults


vtereso:
As per discussion on [this PR](https://github.com/tektoncd/triggers/pull/123), perhaps it is more explicit/better separation to _only_ allow for static parameters to be passed from the top level `EventListener`. Should we find sensible, remove the default field within the `TriggerBinding` input parametes and/or the `TriggerTemplate` params.

ncskier:
@vtereso does https://github.com/tektoncd/triggers/pull/127 close this issue? Or is there still more discussion that you want to have about removing `default` from TriggerTemplate parameters?

vtereso:
Yeah, I would keep it open for discussion about `TriggerTemplate` defaults.

vtereso:
As per WG, the `TriggerTemplate` author should be able to well define the parameters such that  `TriggerTemplate` could be valid without bindings. Closing.

PR #127, opened 2019-09-19
Pass params directly from EventListener to TriggerTemplate

ncskier:
Pass params directly from the EventListener to the TriggerTemplate instead of going though the TriggerBinding. This way, the EventListener params will be merged with the TriggerBinding params; an error is reporting if there are multiple params with the same name.

This simplifies the TriggerBinding, so we no longer need an additional interpolation variable in it. This also leads the way to potentially making the TriggerBinding optional if a user does not care about information in the event.

This also addresses #126 

# Changes
New EventListener yaml (from `examples/`):
```yaml
kind: EventListener
metadata:
  name: listener
spec:
  triggers:
    - binding:
        name: pipeline-binding
      template:
        name: pipeline-template
      params:
        - name: message
```
New TriggerBinding yaml (from `examples/`):
```yaml
kind: TriggerBinding
metadata:
  name: pipeline-binding
spec:
  params:
    - name: gitrevision
    - name: gitrepositoryurl
```

# Submitter Checklist

review them:


docs/eventlisteners.md
```diff
@@ -25,13 +25,13 @@ spec:
...
         name: pipeline-template
+      params:
+        - name: message
+          value: Hello from the Triggers EventListener!
 ```
 
 ## Parameters
-`EventListener`s can provide `params` which are passed to the `TriggerBinding`
-specified by the `binding.name`. Each parameter has a `name` and a `value`.
+`EventListener`s can provide `params` which are passed to the `TriggerTemplate`
```
vtereso: This is slightly confusing? The params are passed to the `TriggerTemplate` as specified, but it also sort of sounds like the params should be under the template. Maybe just me?

pkg/template/resource.go
```diff
@@ -117,3 +97,31 @@ func Uid() string {
 func ApplyUIDToResourceTemplate(rt json.RawMessage, uid string) json.RawMessage {
 	return bytes.Replace(rt, uidMatch, []byte(uid), -1)
 }
+
+// MergeParams merges two param arrays. An error is returned if there are
+// multiple params with the same name.
+func MergeParams(params1 []pipelinev1.Param, params2 []pipelinev1.Param) ([]pipelinev1.Param, error) {
```
vtereso: IIUC, this is the same as `MergeInDefaultParams()`, but does not allow for duplicates. These could be combined (maybe), but I think they are distinct enough as well.

ncskier: `MergeParams` merges two lists of `Param`s, whereas `MergeInDefaultParams` takes one list of `Param`s, and one list of `ParamSpec`s, then adds the `default` values for params not provided in the list

docs/eventlisteners.md
dibyom: The resource templates declare the parameters they need while the params here specify the values for those declared params. Usually, we'd pass those values in from the event via the binding or via default values.

pkg/template/resource.go
ncskier: Do you think there's a more descriptive name that I can use to make this distinction more clear?

docs/eventlisteners.md
ncskier: I rephrased it to hopefully make this more clear.

Issue #124, opened 2019-09-18
Propagate an eventid label to resources created in response to an event 


dibbles:
Proposal to add a label to all resources created in response to an event with the same label, for example:

labels:
  triggers-eventid: 12345

this would enable items to lookup other artefacts created in response to the same event.  For example, a github webhook triggers the event listener causing 2 pipelineruns to launch (2 separate trigger templates) and a monitor taskrun to launch (a 3rd triggertemplate).

The monitor task could now use the downwars api to set the eventid as an environment variable it could use to find which pipelineruns it needs to monitor and report on.  This task could then collate the status and report it once.

@ncskier  has a good understanding of the use case here.

ncskier:
Thanks for creating this issue @dibbles! I think that this is an important use case for us to consider where we're orchestrating work across multiple TriggerTemplates.

This `eventUID` sounds very similar to the `$(uid)` interpolation variable we currently have in the [TriggerTemplates](https://github.com/tektoncd/triggers/blob/master/docs/triggertemplates.md). One suggestion I have is to treat the `eventUID` like we do the existing `uid`. I think it could be helpful to give users the option of using `$(eventUID)` as an interpolation variable in their TriggerTemplates for increased flexibility; for example, `$(eventUID)` could be used as a label or as a name postfix to orchestrate work across TriggerTemplates.

Also, we should include clear documentation from this issue about the difference between the existing `uid` and this proposed `eventUID`.

akihikokuroda:
PR: https://github.com/tektoncd/triggers/pull/133

PR #133, opened 2019-09-23
Add triggers-eventid label for all resources created for each event

akihikokuroda:
# Changes

This PR is for https://github.com/tektoncd/triggers/issues/124

Add triggers-eventid label for all resources created for each event.

When creating resource in the EventLisner for the processing of the event, an additional label is added on created resources with the following value: `triggers-eventid`: `<EventID>`.  The `<EventID>` is assigned for each event processed by the EventListener. 
Mark all resources created for a certain event processing, which helps with event process tracking.

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```


pkg/sink/sink.go
```diff
@@ -68,20 +68,21 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
...
-		r.EventListenerName, r.EventListenerNamespace, string(event), request.Header)
+	eventId := template.Uid()
+	log.Printf("EventListener: %s in Namespace: %s handling event (EventID: %s) with payload: %s and header: %v",
+		r.EventListenerName, r.EventListenerNamespace, eventId, string(event), request.Header)
 	// Execute each Trigger
 	for _, trigger := range el.Spec.Triggers {
-		go r.executeTrigger(event, request.Header, trigger)
+		go r.executeTrigger(event, request.Header, trigger, eventId)
 	}
 	fmt.Fprintf(response, "EventListener: %s in Namespace: %s handling event with payload: %s and header: %v",
```
khrm: Can you add Event ID here also and remove the header and payload? I was thinking of removing the header. I realise that they are not a useful field to receive.

akihikokuroda: I make changes.

pkg/apis/triggers/v1alpha1/register.go
```diff
@@ -25,6 +25,7 @@ import (
 const (
 	GroupName              = "tekton.dev"
 	EventListenerLabelKey  = "/eventlistener"
+	EventIdLabelKey        = "/triggers-eventid"
```
khrm: Nit: It should EventIDLabelKey

Check here:
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/sink/sink.go
```diff
@@ -68,20 +68,21 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
 		return
 	}
 
-	log.Printf("EventListener: %s in Namespace: %s handling event with payload: %s and header: %v",
-		r.EventListenerName, r.EventListenerNamespace, string(event), request.Header)
+	eventId := template.Uid()
```
khrm: Same here for ID vs Id.

test/eventlistener_test.go
```diff
@@ -38,6 +38,7 @@ import (
 )
 
 const resourceLabel = triggersv1.GroupName + triggersv1.EventListenerLabelKey
+const eventIdLabel = triggersv1.GroupName + triggersv1.EventIdLabelKey
```
khrm: Same here. eventIDLabel

pkg/apis/triggers/v1alpha1/register.go
akihikokuroda: OK  I update.

pkg/sink/sink.go
akihikokuroda: I update.

test/eventlistener_test.go
akihikokuroda: I update.

pkg/sink/sink.go
```diff
@@ -134,6 +135,10 @@ func createResource(rt json.RawMessage, restClient restclient.Interface, discove
 	if err != nil {
 		log.Print(err)
 	}
+	rt, err = sjson.SetBytes(rt, "metadata.labels."+triggersv1.LabelEscape+triggersv1.EventIDLabelKey, eventId)
+	if err != nil {
+		log.Print(err)
```
dibyom: Should we return here instead of logging and continuing on?

test/eventlistener_test.go
```diff
@@ -275,6 +276,11 @@ func TestEventListenerCreate(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Error getting ResourceTemplate: %s: %s", wantPr.Name, err)
 		}
+		if gotPr.Labels[eventIdLabel] == "" {
+			t.Fatalf("Instantiated ResourceTemplate missing EventId")
+		} else {
+			delete(gotPr.Labels, eventIdLabel)
```
vtereso: This seems strange. I think it would make more sense if the expected resources had this label already since there is already a below deep comparison. I think comparing the whole objects rather than just the labels is probably good too.

pkg/sink/sink_test.go
```diff
@@ -325,13 +330,33 @@ func Test_createResource(t *testing.T) {
 					return nil, err
 				}
 
+				if tt.name != "Namespace" {
```
vtereso: I don't think it is necessary to make make this check. Technically, we are limited in what we create, but conceptually `createResource()` handles anything.

test/eventlistener_test.go
akihikokuroda: We can not know the expected value for the label so this checks if the label has a value and the rest of the response matches to the expected.

pkg/sink/sink_test.go
```diff
@@ -466,6 +491,11 @@ func Test_HandleEvent(t *testing.T) {
 		if err = json.Unmarshal(body, &gotPipelineResource); err != nil {
 			t.Errorf("Error unmarshalling body as pipelineResource: %s \n%s", string(body), err)
 		}
+		if gotPipelineResource.ObjectMeta.Labels["tekton.dev/triggers-eventid"] == "" {
```
vtereso: At this point, the API server takes over and there is no way for us to validate this.

pkg/sink/sink_test.go
akihikokuroda: Without this check, the Marshal(gotPipelineResource) in the else clause creates additional items like "spec".

pkg/sink/sink_test.go
```diff
@@ -480,7 +510,11 @@ func Test_HandleEvent(t *testing.T) {
 		t.Fatalf("Error reading response body: %s", err)
 		return
 	}
-	wantPayload := `EventListener: my-eventlistener in Namespace: foo handling event with payload: {"head_commit": {"id": "testrevision"}, "repository": {"url": "testurl"}}`
+	wantPayload := `EventListener: my-eventlistener in Namespace: foo handling event`
```
vtereso: Can you clean this section to mirror the [original commit](https://github.com/tektoncd/triggers/commit/c9537c0636218b51766dfc5dbab4db5bdbd0389d#diff-aa07fc23750dee7511c478c5311b0d15)? We don't currently return a response.

pkg/sink/sink.go
vtereso: I think returning probably makes most sense too. If we expect the label and it isn't there we won't ever know, but we could alternatively not create resources, which seems safer.

akihikokuroda: I'm not sure.   This is same way as the event listener name label.  I leave it for now.

test/eventlistener_test.go
vtereso: Right, I forgot this was an e2e where we couldn't seed the generator.

pkg/sink/sink_test.go
vtereso: I see the issue. The labels aren't guaranteed to be ordered. It works if we marshal into an interface (map[string]interface{} underlying type):
```go
var actualResource, expectedResource interface{}
				if err := json.Unmarshal(body, &actualResource); err != nil {
					t.Error(err)
				}
				if err := json.Unmarshal(tt.wantResource, &expectedResource); err != nil {
					t.Error(err)
				}
				if diff := cmp.Diff(expectedResource, actualResource); diff != "" {
					t.Errorf("Diff request body: -want +got: %s", diff)
				}
```

pkg/sink/sink_test.go
akihikokuroda: That's true.

pkg/sink/sink_test.go
akihikokuroda: Actually we do.  The response is put at the end of the "HandleEvent" in sink.go.

pkg/sink/sink_test.go
akihikokuroda: OK  I update.

pkg/sink/sink.go
akihikokuroda: OK  I add return.

pkg/sink/sink_test.go
vtereso: Ah, good point. I didn't realize we were writing to the response. Can we seed the eventid so that we can compare the body with an exact wantPayload?

docs/triggertemplates.md
```diff
@@ -64,6 +64,9 @@ One instance where there is useful is when resources in a `TriggerTemplate` have
 When creating resource, an additional label is added on created resources with the following value: `event-listener-name`: `<EventListenerName>`.
 Mark all resources created by a certain EventListener, which helps with housekeeping.
 
+When creating resource, another additional label is added on created resources with the following value: `triggers-eventid`: `<EventID>`.  The `<EventID>` is assigned for each event processed by the EventListener.
```
vtereso: Can we update the key here to be `tekton.dev/triggers-eventid`? Also, the above key should be `tekton.dev/eventlistener`.

akihikokuroda: I make changes

pkg/sink/sink_test.go
akihikokuroda: No, I can seed it.  It is generated by the eventlistener in this test.

Issue #132, opened 2019-09-23
EventListenerTrigger should have a Name property


dibbles:
# Expected Behavior

I expect to be able to easily identify different triggers within an eventlistener.   I would propose that each trigger have a Name, thus I can easily find an entry I want to delete in the situation where only a parameter value differs between multiple entries

# Actual Behavior

If I have multiple entries in an eventlistener using the same triggerbinding and triggertemplate but maybe one different parameter value i have to quite clumsy programmatically check all the values for each trigger until I find the one matching my requirements.

khrm:
@dibbles Are you looking for something like this: https://github.com/tektoncd/triggers/pull/134

PR #134, opened 2019-09-23
Add name to EventlistenerSpec Trigger

khrm:
Add Name for EventlistenerSpec Trigger.
Also, add it to test builder.
Also, documented in the example.
Resolves https://github.com/tektoncd/triggers/issues/132

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/apis/triggers/v1alpha1/event_listener_types.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_types.go
@@ -39,6 +39,8 @@ type EventListenerTrigger struct {
 	Binding         EventListenerBinding  `json:"binding"`
 	Template        EventListenerTemplate `json:"template"`
 	Params          []pipelinev1.Param    `json:"params,omitempty"`
+	// +optional
+	Name string `json:"name,omitempty"`
 }
 
 // EventListenerBinding refers to a particular TriggerBinding resource.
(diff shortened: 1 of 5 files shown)


dibbles:
That would probably suffice ... yes..... however my concern would be around whether you would want all the additional properties that ObjectMeta allows?  It would however resolve my problem I believe.

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -39,6 +39,8 @@ type EventListenerTrigger struct {
 	Binding         EventListenerBinding  `json:"binding"`
 	Template        EventListenerTemplate `json:"template"`
 	Params          []pipelinev1.Param    `json:"params,omitempty"`
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
dibyom: Agree with @dibbles 's comment here -- does supporting all the other fields in ObjectMeta make sense here? e.g. namespaces, generateName, selfLinks etc? If not, maybe a name of type string should suffice?

Also if we are going to use objectMeta, the Trigger type is starting to look a lot like its own CustomResource. In that case, the other fields like Binding, Template, Params etc should be within a Spec?

khrm: Yea. Should we have Trigger CRD? Will change to just string `name` for now and create an issue regarding this so that we can discuss this properly.

khrm:
@dibbles @dibyom I have changed it to `string` type.

PR #135, opened 2019-09-24
Getting started guide

iancoffey:
# Changes

Adds a getting started guide, which allows new operators to "get started" with a real CI system.

Update: This makes use of the new ingress and webhook Tasks now, which is awesome.

There is def some overlap between the documentation that exists and this guide, but this guide still a nice way to get started building your first pipelines and seeing results. Totally up for whatever direction the WG wants to do with the docs organization.

# Submitter Checklist

review them:


docs/create-webhook.yaml
```diff
@@ -33,7 +33,7 @@ spec:
       description: The github repo name
       default: "repo"
     - name: GithubSecretName
-      description: The secret name for github
+      description: The secret name for github access token
```
dibyom: access token, and optionally the webook secret

docs/getting-started/pipeline.yaml
```diff
@@ -0,0 +1,134 @@
...
+spec:
+  resources:
+    - name: source-repo
+      type: git
+  check:
+    image: golang
+    command: ["go"]
+    args: ['test', "$(inputs.resources.source-repo.path)/..."]
+---
+# Finally, we need something to receive our cloudevent announcing success!
```
dibyom: Does this mean that users following getting started will need to have knative serving installed?

docs/getting-started/rbac/admin-role.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - tekton.dev
+  resources:
+  - eventlisteners
+  - triggerbindings
+  - triggertemplates
+  verbs:
+  - get
+- apiGroups:
+  - tekton.dev
+  resources:
```
dibyom: probably need create on pipelineresources as well?

docs/getting-started/pipeline.yaml
iancoffey: Good point - ill make this a generic corev1 service instead.

docs/create-webhook.yaml
iancoffey: I removed these non-getting-started changes from this PR, ty!

docs/getting-started/pipeline.yaml
iancoffey: I will open a dedicated create-webhook PR to gather and focus on these fixes to that Task.

docs/getting-started/rbac/admin-role.yaml
iancoffey: Yes, good call, ty

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
+# Getting started with Triggers
+
+To get started with Triggers, lets put it to work building and deploying a real image. In the following guide we will use `Triggers` to handle a real Github webhook request to kickoff a PipelineRun.
```
vtereso: Nit: let's

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+Before we can use the Triggers project, we need to get some dependencies out of the way.
+
+  - [A Kubernetes Cluster](https://kubernetes.io/docs/setup/)
+    - For our purposes any publicly available k8s cluster will do, as long as it can produce a Service Loadbalancer.
+  - [Install Tekton Pipelines](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#installing-tekton-pipelines)
+    - Pipelines is the backbone of Tekton and will allow us to accomplish the work we plan to do.
+  - [Install Triggers](https://github.com/tektoncd/triggers/blob/master/DEVELOPMENT.md#install-triggers)
+    - Of course we need to install our project as well, so we can accept and process events into PipelineRuns!
+  - Pick a Github repo with a Dockerfile as your build object (or you can fork [this one](https://github.com/iancoffey/ulmaceae)).
```
vtereso: Nit: GitHub

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
...
+Now we have to install the Pipeline we plan to use and also our Triggers resources.
+
+Our Pipeline will do a few things:
+- Retrieve the source code
+- Build and push the source code into a docker image
+- Push the image to the specified repository
+- Run the image locally
+<!-- -->
+- [Install the Pipeline](./pipeline.yaml)
+  - The Pipeline will build a docker image with img and deploy it locally via kubectl image.
```
vtereso: I didn't know about `img`. Linking to the source seems nice: https://github.com/genuinetools/img

vtereso: Nit: Docker

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
...
+  - `kubectl apply -f ./docs/getting-started/pipeline.yaml`
+<!-- -->
+Our Triggers project will pickup from there:
+- We will setup an EventListener to accept and process Github Push events
+- A TriggerTemplate to create templates Resources and PipelineRun per event received.
```
vtereso: Nit: Backticks on the Triggers resources?
I feel like it's probably better for the sake of consistency with the rest of the docs

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
...
+  -  First, **edit** the `triggers.yaml` file to reflect
+    - The Docker repository to push the image blob
+      - You will the `DOCKERREPO-REPLACEME` string everywhere it is needed.
+  - Once you have updated the triggers file, you can apply it!
+  - `kubectl apply -f ./docs/getting-started/triggers.yaml`
+  - If that succeeded, your cluster is ready to start handling Events.
+
+## Configure Github Webhook
+
+Now we need some events to work with. We need to configure Github to send events to our Cluster, which is quick but has a few tricky parts.
```
vtereso: Nit: cluster

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+- Establish your inbound EXTERNAL-URL hostname:
+  - In order to accept Github webhook payloads, our listener must be reachable from the internet.
+  - Our EventListener will create a [Loadbalancer Service](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) for us, so we just need to determine its hostname. How the loadbalancer gets created will depend on how you are deploying this getting-started - AWS will behave differently than GCP, and so on.
```
vtereso: The service now uses ClusterIP

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
...
+  - Inspect the output of `k get service/getting-started-listener -n getting-started -o yaml` - there should be a hostname associated with the ingress now.
```
vtereso: Not everyone has `k` as an alias to `kubectl`. Also, `kubectl` is used above and simplifies copy-pasting.

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
...
+- [Install the Pipeline](./pipeline.yaml)
+  - The Pipeline will build a docker image with img and deploy it locally via kubectl image.
+  - `kubectl apply -f ./docs/getting-started/pipeline.yaml`
+<!-- -->
+Our Triggers project will pickup from there:
+- We will setup an EventListener to accept and process Github Push events
+- A TriggerTemplate to create templates Resources and PipelineRun per event received.
+<!-- -->
+- [Install the TriggerTemplate, TriggerBinding and EventListener](./triggers.yaml)
+  -  First, **edit** the `triggers.yaml` file to reflect
```
vtereso: Was this line supposed to be connected to the below one?

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
...
+![Github Webhook Push](images/trigger-webhook.png)
+
+Make sure to set `push` events as the type of events you want to send - push is the default.
+
+![Github Webhook Setup](images/trigger-webhook2.png)
+
+## Watch it work!
+
+- Commit and push an empty commit to your development repo:
+  - `git commit -a -m"build commit" --allow-empty && git push origin mybranch`
```
vtereso: Nit: Space after the `-m` flag

docs/getting-started/triggers.yaml
```diff
@@ -0,0 +1,101 @@
...
+        params:
+          - name: targetURI
+            value: http://event-display.getting-started.svc.cluster.local
+    - apiVersion: tekton.dev/v1alpha1
+      kind: PipelineRun
+      metadata:
+        name: getting-started-pipeline-run-$(uid)
+        namespace: $(params.namespace)
+        labels:
+          triggertemplated: "true"
```
vtereso: All resources now have the labels [specified here](https://github.com/tektoncd/triggers/blob/master/docs/triggertemplates.md). As a result we can remove this label.

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
...
+- Now, you can follow the Task output in `kubectl logs`:
+  - First the image builder task:
+    - `kubectl logs -l somelabel=somekey --all-containers`
```
vtereso: We could add a label on the builder so that this could be well defined.

docs/getting-started/README.md
```diff
@@ -0,0 +1,95 @@
...
+  - Then our deployer task:
+    - `k logs -l tekton.dev/pipeline=getting-started-pipeline -n getting-started --all-containers`
```
vtereso: Same as above.

docs/getting-started/README.md
iancoffey: ty, will update.

docs/getting-started/README.md
iancoffey: 14 places! thanks 😄

docs/getting-started/README.md
iancoffey: makes sense, will update

docs/getting-started/README.md
iancoffey: Right, I hadnt noticed the change. No easy solution to it. To allow this sort of workflow to function, I added a new bool field on `EventListener.Spec.Public` in this branch. When set to true the service will be a LB, else its a ClusterIP.

docs/getting-started/triggers.yaml
iancoffey: sgtm 👍

docs/getting-started/README.md
iancoffey: Yeah good idea +1

iancoffey: Fixed! ty

docs/getting-started/README.md
iancoffey: Very interested in opinions on this part

docs/getting-started/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Watch it work!
+
+- Commit and push an empty commit to your development repo:
+  - `git commit -a -m "build commit" --allow-empty && git push origin mybranch`
+- Now, you can follow the Task output in `kubectl logs`:
+  - First the image builder task:
+    - `kubectl logs -l somelabel=somekey --all-containers`
+  - Then our deployer task:
+    - `k logs -l tekton.dev/pipeline=getting-started-pipeline -n getting-started --all-containers`
```
khrm: Can we `kubectl logs` instead of `k`? While it's quite common alias but some might not have that enabled.

iancoffey: Yeah def. these `k` refs are all typos, ty

docs/getting-started/README.md
iancoffey: UPDATE, this will become an ingress instead.

akihikokuroda:
Would you add a link to https://github.com/tektoncd/triggers/blob/master/docs/createwebhook.md
at the beginning of "Configure GitHub Webhook" section in this getting-started README.md for the secure webhook eventing?

iancoffey:
@dibyom I had forgotten to add the TaskRun files, ty for noticing.

I actually really like the Task approach @vtereso created, and this should work well enough for now imo.

vtereso:
@dibyom The simple service ingress seems interesting. I think that would be a good addition to the docs, but I can't really say I know how it works? Since there is no domain, what is the callback url for the webhook. Also, how does it handle TLS?

dibyom:
I really like the tasks too! My main concern is that it requires that the users have a readily available domain available that they can use in order to complete the getting started guide.

@vtereso I'm going off of https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer#step_3_create_an_ingress_resource

Not sure how this works with other cloud providers or if you are using minikube. To be clear, I'm not saying we replace the task -- it does the right things e.g. TLS, certs. etc. This is more of an alternative to the `LoadBalancer` service type that we were using initially and just for the getting started guide.

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -54,6 +54,7 @@ type EventListener struct {
 type EventListenerSpec struct {
 	ServiceAccountName string                 `json:"serviceAccountName"`
 	Triggers           []EventListenerTrigger `json:"triggers"`
+	Public             bool                   `json:"public"`
```
dibyom: This is not necessary anymore since we are using Ingress right?

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -146,12 +146,16 @@ func reconcileObjectMeta(oldMeta *metav1.ObjectMeta, newMeta metav1.ObjectMeta)
 }
 
 func (c *Reconciler) reconcileService(el *v1alpha1.EventListener) error {
+	serviceType := corev1.ServiceTypeClusterIP
```
dibyom: Same as above.

dibyom:
I don't think we should hold up this PR for the Ingress suggestion I made. We can always add that in later (once I confirm it works :stuck_out_tongue: )

I think this still needs a rebase + some small typos (k vs kubectl), and taking out the `public` field from the EventListener type

pkg/apis/triggers/v1alpha1/event_listener_types.go
iancoffey: Great catch!

docs/getting-started/triggers.yaml
```diff
@@ -0,0 +1,94 @@
...
+      value: $(event.head_commit.url)
+---
+apiVersion: tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: getting-started-listener
+  namespace: getting-started
+spec:
+  public: true # this is new
+  serviceAccountName: task-admin
```
dibyom: ` tekton-triggers-admin`? `task-admin` does not exist

docs/getting-started/ingress-run.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: create-ingress-run
+  namespace: getting-started
+spec:
+  taskRef:
+    name: create-ingress
+  inputs:
+    params:
+    - name: CreateCertificate
+      value: "true"
+    - name: CreateEventListener
```
dibyom: This param does not exist in the latest version of the task

docs/getting-started/rbac/webhook-role.yaml
```diff
@@ -0,0 +1,55 @@
...
+  namespace: getting-started
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: tekton-triggers-createwebhook
+  namespace: getting-started
+subjects:
+  - kind: ServiceAccount
+    name: tekton-triggers-webhook
```
dibyom: typo: should be tekton-triggers-createwebhook

docs/getting-started/rbac/webhook-role.yaml
```diff
@@ -0,0 +1,55 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: tekton-createwebhook
```
dibyom: typo: should be tekton-triggers-createwebhook

iancoffey:
I believe this is ready for another review! Ty in advance 😸

docs/getting-started/README.md
```diff
@@ -0,0 +1,137 @@
+# Getting started with Triggers
+
+To get started with Triggers, let's put it to work building and deploying a real image. In the following guide we will use `Triggers` to handle a real GitHub webhook request to kickoff a PipelineRun.
+
+## Install dependencies
+
+Before we can use the Triggers project, we need to get some dependencies out of the way.
+
+  - [A Kubernetes Cluster](https://kubernetes.io/docs/setup/)
+    - For our purposes any publicly available k8s cluster will do, as long as it can produce a Service Loadbalancer.
```
vtereso: The service is no longer a LoadBalancer, but ClusterIP. When speaking with @dibyom, it should also be documented that ingress does not work out of the box on GKE. See: https://github.com/kubernetes/kubernetes/issues/26508#issuecomment-222376962

docs/getting-started/README.md
```diff
@@ -0,0 +1,137 @@
...
+- Build and push the source code into a Docker image
+- Push the image to the specified repository
+- Run the image locally
+
+#### What does it do?
+
+The Pipeline will build a Docker image with [img](https://github.com/genuinetools/img) and deploy it locally via kubectl image.
+
+### [Install the TriggerTemplate, TriggerBinding and EventListener](./triggers.yaml)
+The Triggers project will pickup from there:
```
vtereso: Here we use colon, but in other similar structures, it is a period. Whatever it be, I think we should be consistent.

docs/getting-started/README.md
```diff
@@ -0,0 +1,137 @@
...
+- A `TriggerTemplate` to create templated PipelineResource and PipelineRun resources per event received by the `EventListener`.
+  -  First, **update** the `triggers.yaml` file to reflect the Docker repository you wish to push the image blob to.
+    - You will need to replace the `DOCKERREPO-REPLACEME` string everywhere it is needed.
+  - Once you have updated the triggers file, you can apply it!
+    - `kubectl apply -f ./docs/getting-started/triggers.yaml`
+  - If that succeeded, your cluster is ready to start handling Events.
+
+## Add Ingress and Github-Webook Tasks
+
+We will need an ingress to handle incoming webhooks and we will make use of our new ingress by configuring Github with our GitHub Task.
```
vtereso: Nit: Github -> GitHub

docs/getting-started/README.md
```diff
@@ -0,0 +1,137 @@
+# Getting started with Triggers
+
+To get started with Triggers, let's put it to work building and deploying a real image. In the following guide we will use `Triggers` to handle a real GitHub webhook request to kickoff a PipelineRun.
```
vtereso: Nit: Maybe add a comma? In the following guide,

docs/getting-started/README.md
```diff
@@ -0,0 +1,137 @@
...
+- Push the image to the specified repository
+- Run the image locally
+
+#### What does it do?
+
+The Pipeline will build a Docker image with [img](https://github.com/genuinetools/img) and deploy it locally via kubectl image.
+
+### [Install the TriggerTemplate, TriggerBinding and EventListener](./triggers.yaml)
+The Triggers project will pickup from there:
+- We will setup an EventListener to accept and process GitHub Push events
```
vtereso: Nit: Backticks

docs/getting-started/README.md
dibyom: We need a better solution long term (e.g. maybe let users specify the service type) but for now, we could probably use the nginx ingress: https://kubernetes.github.io/ingress-nginx/deploy/#gce-gke

(I can add this bit in later once this PR is merged)

docs/getting-started/ingress-run.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - name: CertificateKeyPassphrase
+      value: asecretphrase
+    - name: CertificateSecretName
+      value: ingresssecret
+    - name: ExternalDomain
+      value: getting-started.iancoffey.com
+    - name: Service
+      value: getting-started
+    - name: ServicePort
+      value: 8080
```
dibyom: "8080" else I get a validation error

docs/getting-started/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
```
dibyom: Nice! Should fix the integration tests!

docs/getting-started/rbac/admin-role.yaml
```diff
@@ -0,0 +1,41 @@
...
+  namespace: getting-started
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: tekton-triggers-admin-binding
+  namespace: getting-started
+subjects:
+  - kind: ServiceAccount
+    name: tekton-triggers-admin-sa
```
dibyom: This should be tekton-triggers-admin

docs/getting-started/namespace.yaml
dibyom: Hmm, looks like we have to ensure this is is the first YAML that gets applied.
An alternative might be to explicitly add the create namespace in `test/e2e-tests-ingress.sh`

iancoffey: I renamed the ns yaml to `01-namespace.yaml`, I wonder if it will get ordered first.

docs/getting-started/rbac/admin-role.yaml
iancoffey: great catch

docs/getting-started/README.md
iancoffey: Added a bit for this, ty. Im not super familiar with the ingress/GKY situation so any additional suggestions welcome! 😄

iancoffey:
Most (all?) feedback addressed, thanks! How is this looking?

docs/getting-started/namespace.yaml
dibyom: That works too!

dibyom:
I think this looks great. Probably just need to squash up the commits

iancoffey:
The create-webhook task is failing. One thats figured out, I think itl all be good.
```
triggers (getting-started) $ k logs -l tekton.dev/taskRun=create-webhook-run -n getting-started 
Create Webhook
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
{
  "message": "Problems parsing JSON",
  "documentation_url": "https://developer.github.com/v3/repos/hooks/#create-a-hook"
}
```

dibyom:
hmmm, should we be creating a webhook for each PR?

iancoffey:
Rebased, squashed and a link added to the main readme.

Issue #125, opened 2019-09-18
Pod proliferation problem using Triggers with many webhooks


vtereso:
# Expected Behavior
ValidationTasks allow events to be properly validated, which gates `EventListener` Triggers

# Actual Behavior
This works "as intended" assuming low traffic, but with potentially many Triggers on an `EventListener`, this will start a validate `Task` for every single Trigger on the the `EventListener`, which can currently be the source of multiple event sources resulting in a lot of wasted work.

For example, with just two different GitHub repositories (they require separate validation), one may be receiving many events, while the other receives none. Now, whenever an event comes in, the validation `Task` for this inactive repository will be run. 

This could easily result in spikes of thousands of pods being created if someone were to put all their Triggers in a single `EventListener`. With a forced single source approach, N listener pods are created for N event sources (compared to 1), but this allows for native IP tables to route to the underlying services to these pods (with a single validate) rather than it being resolved internally through many much slower `Task`s.

# Additional Info
[Here](https://github.com/tektoncd/triggers/issues/125#issuecomment-533171392) are some questions to evaluate under this issue.

dibbles:
@vtereso I'm not sure I exactly follow what is it you are specifying here - but I think this in relation to conversations we were having yesterday about "pod proliferation" 

I think what you are saying is "only start one instance of the validate task/pod" and use an underlying service from that validate pod to report back for each of the triggers?  Is that right?
 
Do you know if the plan is to allow different validate tasks to be specified for different triggers in an event listener?

Do you think the event listener should have validate tasks listed at the top level and each trigger refers to one of those defined tasks?

khrm:
> Do you know if the plan is to allow different validate tasks to be specified for different triggers in an event listener?
At present, you could have Github and Gitlab combined in a single one. 

> Do you think the event listener should have validate tasks listed at the top level and each trigger refers to one of those defined tasks?
I don't think this would solve. Won't during validation you will have to create pod anyway?


I think what @vtereso want is just one source for an event listener. So that we have just one 'pod' which communicates with EventListener during validation.  Not using tasks or conditionals. 

We could use a side car for validation?

vtereso:
[...]

PR #140, opened 2019-10-01
Event Interceptors for pre processing Events

dibyom:
# Changes
Adds an optional `Interceptor` field to Triggers that takes an ObjectRef. Currently only k8s Services are supported. If an EventInterceptor is specified, the EL will forward incoming requests to the Interceptor service which is expected to process the event and return a response back. The status code determines if the processing is successful and the returned body is used as the new event payload that is used by the trigger binding.

An example interceptor service that validates Github requests is present.

**TODO**:
- [x] Update task for github webhook validation: #144 
- [ ] Consider renaming Interceptors to `EventProcessor`
- [ ] Add an integration test

Addresses #125 
/cc @khrm @vtereso 

# Submitter Checklist

review them:

# Release Notes

```
Adds an optional `interceptor` field to `Triggers` that can be used to pre-process events before passing them on to the `TriggerBindings`
```


dibyom:
Looks like we need go 1.13 for Prow

cmd/interceptors/gh-validate/main.go
```diff
@@ -0,0 +1,55 @@
...
+		log.Printf("Handling Github Event with delivery ID: %s", github.DeliveryID(request))
+		payload, err := github.ValidatePayload(request, []byte(secretToken))
+		if err != nil {
+			log.Printf("Error handling Github Event with delivery ID %s : %q", github.DeliveryID(request), err)
+			http.Error(writer, fmt.Sprint(err), http.StatusBadRequest)
+		}
+		n, err := writer.Write(payload)
+		fmt.Printf("%s", payload)
+		if err != nil {
+			log.Printf("Failed to write response for github event ID: %s. Bytes writted: %d. Error: %q", "TODO", n, err)
```
vtereso: Nit: The above log uses `Github Event`

pkg/sink/interceptors.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	defer resp.Body.Close()
+
+	respBody, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, xerrors.Errorf("Failed to parse response body")
+	}
+
+	// Wrap error and return
+	if resp.StatusCode != http.StatusOK {
```
vtereso: Seems fine, but we should document whatever status codes will be accepted

dibyom: Yeah, I think just 200 OK should be it?

pkg/sink/sink.go
```diff
@@ -59,27 +62,48 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
...
+func (r Resource) executeTrigger(payload []byte, request *http.Request, trigger triggersv1.EventListenerTrigger, eventId string) {
+	header := request.Header
+	if trigger.Interceptor != nil {
+		interceptorUrl, err := GetURI(trigger.Interceptor.ObjectRef, r.EventListenerNamespace) // TODO: Cache this result or do this on initialization
+		if err != nil {
+			log.Printf("Could not resolve Interceptor Service URI: %w", err)
+			return
+		}
+
+		modifiedPayload, err := r.processEvent(interceptorUrl, request, payload)
```
vtereso: I suppose an interceptor can intercept and augment the payload however they see fit. However, it seems to enforce that a payload should be returned?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -66,6 +66,22 @@ type EventListenerTrigger struct {
...
+}
+
+// EventInterceptor provides a hook to intercept and pre-process events
+type EventInterceptor struct {
+	// ObjectRef is a reference to an object that will resolve to a cluster DNS
+	// name to use as the EventInterceptor. Either objectRef or url can be specified
+	// +optional
+	ObjectRef *corev1.ObjectReference `json:"objectRef,omitempty"`
+
+	// Url is a full URL for the interceptor service. Also, the only supported
```
vtereso: Only supported option for now?

dibyom: Incorrect docs...will fix

pkg/sink/interceptors.go
dibyom: Will add docs

pkg/sink/sink.go
dibyom: Right now, the contract is the service always has to return a body (adding docs). If the body should remain unmodified, simply return it back. We could potentially make this optional I guess?
1. If a body is returned use that as the event Payload
2. If not, just use the original body as the event Payload

pkg/sink/interceptors.go
dibyom: Added docs

pkg/sink/sink.go
vtereso: I think either approach makes sense. 👍

pkg/sink/interceptors.go
```diff
@@ -0,0 +1,87 @@
...
+	"golang.org/x/xerrors"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Retrieve ObjectReference to URI
+func GetURI(objRef *corev1.ObjectReference, ns string) (*url.URL, error) {
+	// TODO: This should work for any Adressable.
+	// Use something like https://github.com/knative/eventing-contrib/blob/7c0fc5cfa8bd44da0767d9e7b250264ea6eb7d8d/pkg/controller/sinks/sinks.go#L32
+	if objRef.Kind == "Service" && objRef.APIVersion == "v1" {
+		// TODO: Also assuming port 80 and http here. Use DNS/or the env vars?
```
vtereso: Can we add an additional line to the documentation that the interceptor service should specify port 80?

docs/eventlisteners.md
```diff
@@ -14,6 +14,35 @@ references a listener pod. This listener pod accepts the incoming events and
...
+
+## Parameters
+`EventListener`s can provide `params` which are merged with the `TriggerBinding`
+`params` and passed to the `TriggerTemplate`. Each parameter has a `name` and a
+`value`.
+
+
+## Event Interceptors
+
+Triggers within an EventListener can optionally specify an interceptor field
```
vtereso: Nit: Backticks within this paragraph on the EventListener type.

pkg/sink/interceptors_test.go
```diff
@@ -0,0 +1,168 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			ts := httptest.NewServer(http.HandlerFunc(tc.handler))
+			defer ts.Close()
+
+			req, err := http.NewRequest(http.MethodPost, ts.URL, bytes.NewBuffer(reqBody))
+			if err != nil {
+				t.Fatalf("Error trying to create request: %q", err)
+			}
+			payload, err := makeRequest(http.DefaultClient, req)
+			expectedPayload := tc.expectedPayload
```
vtereso: Could this check the status code rather than the body? It makes sense that the body is not the one we expect in this test, but that should be the source of truth?

docs/eventlisteners.md
```diff
@@ -14,6 +14,35 @@ references a listener pod. This listener pod accepts the incoming events and
...
+which contains an `ObjectReference` to a Kubernetes Service. If an interceptor
```
vtereso: Maybe a link to the docs on ObjectReference?

docs/eventlisteners.md
```diff
@@ -14,6 +14,35 @@ references a listener pod. This listener pod accepts the incoming events and
...
+`params` and passed to the `TriggerTemplate`. Each parameter has a `name` and a
+`value`.
+
+
+## Event Interceptors
+
+Triggers within an EventListener can optionally specify an interceptor field
+which contains an `ObjectReference` to a Kubernetes Service. If an interceptor 
+is specified, the `EventListener` sink will forward incoming events to the 
+service referenced by the interceptor over HTTP. The service is expected to
```
vtereso: I didn't even catch the HTTP part when reading this initially. Maybe still be explicit about port 80?

test/builder/eventlistener.go
```diff
@@ -204,3 +204,16 @@ func EventListenerTriggerValidateParam(name, value string) EventListenerTriggerV
 		)
 	}
 }
+
+// EventListenerTriggerValidateParam adds a param name to the TriggerValidate.
```
vtereso: Nit: Docstring

test/builder/eventlistener.go
```diff
@@ -204,3 +204,16 @@ func EventListenerTriggerValidateParam(name, value string) EventListenerTriggerV
...
+func EventListenerTriggerInterceptor(name, version, kind string) EventListenerTriggerOp {
```
vtereso: I don't know if the intent is to restrict interceptors to the same namespace since there is no parameter for it?

docs/eventlisteners.md
dibyom: Yeah, the port 80 thing is true at the moment but should change once we support all Addressable types. Will add to the docs for now

test/builder/eventlistener.go
dibyom: No, you can specify a namespace, but it defaults to the namespace of the EL. 
Added a namespace param to the builder.

pkg/sink/interceptors_test.go
dibyom: Yes, makeRequest returns an error if the status code is not 200. The test case tests that it does indeed return an error if the service returns a 400 status code.

dibyom:
Thoughts on the `EventInterceptor` name?  Alternatives:
* `EventProcessor`
* `EventPreProcessor`
* `TriggerAdmissionController`?

vtereso:
Thanks @dibyom!
/lgtm

PR #144, opened 2019-10-04
Adds a Tekton Task to create github validation interceptor

dibyom:
# Changes

Along with the task, I added some additional documentation.

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/docs/create-gh-validate-interceptor.yaml
+++ b/docs/create-gh-validate-interceptor.yaml
@@ -1,0 +1,77 @@
+# This Task deploys the example github validation webhook (cmd/gh-validate) and
+# exposes it as a ClusterIP service.
+# See examples/event-interceptors and docs/eventlisteners.md#event-interceptors
+# for more information.
+
+apiVersion: tekton.dev/v1alpha1
+kind: Task
+metadata:
+  name: create-github-validate-interceptor
+spec:
+  inputs:
+    params:
+      - name: Name
+        description: "The name of the interceptor Service."
+      - name: GitHubSecretName
+        description: "The Secret name for GitHub webhook secret. This is always mounted and must exist"
+      - name: GitHubSecretStringKey
+        description: "Key name within the Secret for the Github webhook secret."
+      - name: ContainerImage
+        description: "Container image of the Github Validation Interceptor. Defaults to the latest release and can be overridden with your own image"
+        default: "gcr.io/tekton-releases/github.com/tektoncd/triggers/cmd/gh-validate@sha256:98e4f1ac48fbad76f51d5eae68147c9800dc4c9a2e95a60532c92fc4c8d204b1"
+  steps:
+    - name: create-service
+      image: lachlanevenson/k8s-kubectl:latest
+      command:
+        - sh
+      args:
+        - -ce
+        - |
+          set -e
+          cat <<EOF | kubectl create -f -
+          apiVersion: v1
+          kind: Service
+          metadata:
+            name: $(inputs.params.Name)
+          spec:
+            type: ClusterIP
+            selector:
+              app: $(inputs.params.Name)
+            ports:
+              - protocol: TCP
+                port: 80
+                targetPort: 8080
+          EOF
+    - name: create-deployment
+      image: lachlanevenson/k8s-kubectl:latest
+      command:
+        - sh
+      args:
+        - -ce
+        - |
+          set -e
+          cat <<EOF | kubectl create -f -
+          apiVersion: apps/v1
+          kind: Deployment
+          metadata:
+            name: $(inputs.params.Name)
+          spec:
+            replicas: 1
+            selector:
+              matchLabels:
+                app: $(inputs.params.Name)
+            template:
+              metadata:
+                labels:
+                  app: $(inputs.params.Name)
+              spec:
+                containers:
+                  - name: validate
+                    image: $(inputs.params.ContainerImage)
+                    env:
+                      - name: GITHUB_SECRET_TOKEN
+                        valueFrom:
+                          secretKeyRef:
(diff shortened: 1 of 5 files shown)


examples/event-interceptors/README.md
```diff
@@ -0,0 +1,25 @@
+## Sample Event Interceptors
+
+This folder contains some sample event interceptors:
+
+* [Github Validation](#github-validation)
+
+
+### Github Validation
+
+Validates Github webhook payloads using a supplied webhook secret token. Assumes
```
vtereso: Nit: Github -> GitHub

examples/event-interceptors/gh-validate.yaml
```diff
@@ -0,0 +1,53 @@
+# Copyright 2019 The Tekton Authors
```
vtereso: Doesn't this already exist?

dibyom: Indeed it does

examples/event-interceptors/README.md
dibyom: Added a comment to change this. The task does allow the parameter to be changed.

dibyom: Is it worth it to have both a task as well as s deployment? A user could choose to use either. Maybe having both and that too in two different places is confusing? What do you think @a-roberts ?

a-roberts: I like the idea of a Tekton Task yeah, users get to see the logs e.g. through the dashboard or kubectl trivially, they have a "record" (in the form of a TaskRun CRD) of the interceptor being created (which might be useful on a prod system where serviceability is an important factor) and it serves as another example of how a good Tekton Task definition should look, so I'm all for having both

docs/create-gh-validate-interceptor.yaml
```diff
@@ -0,0 +1,74 @@
...
+spec:
+  inputs:
+    params:
+      - name: Name
+        description: "The name of the interceptor Service."
+      - name: GitHubSecretName
+        description: "The Secret name for GitHub webhook secret. This is always mounted and must exist"
+      - name: GitHubSecretStringKey
+        description: "Key name within the Secret for the Github webhook secret."
+      - name: ContainerImage
```
a-roberts: I'm following these docs implementing a validator for our move to use Triggers with our Tekton Webhooks Extension, it's not clear what this container image might do exactly or if this should be replaced to use, say, my own Docker registry if I go on to ko apply the provided Task

Should the container image, say, run some script perhaps that would return a non-zero exit code if some condition fails? 

What would say script look like (I see https://github.com/tektoncd/triggers/blob/6e81d0f3dea0497f70e32cd5dc30285703d451e7/examples/event-interceptors/github-validate.yaml, but I'm used to validation being done through the Task itself like with https://github.com/tektoncd/experimental/pull/258/files#diff-16e2249a36211431f525dd0f47c1703c)

I know this is a WIP so perhaps that's coming or elsewhere!

The use cases we're looking at specifically include:
- Checking valid X-GitHub signature (checking secretToken used in the validation service with what's set on the webhook)
- Checking a repository URL matches an input parameter (so templates/bindings only get used to trigger PipelineRuns/PipelineResource generation for a given repository, not everything we have webhooks for)
- Eventually checking the type of event coming in (e.g. a push or pull request being created, not just anything)

examples/event-interceptors/README.md
a-roberts: An example TaskRun would be super useful here too btw

docs/create-gh-validate-interceptor.yaml
```diff
@@ -0,0 +1,74 @@
...
+          spec:
+            type: ClusterIP
+            selector:
+              app: $(inputs.params.Name)
+            ports:
+              - protocol: TCP
+                port: 80
+                targetPort: 8080
+          EOF
+          fi
```
a-roberts: I'm pretty sure the fi's cause errors (we don't have any ifs after all)

```
adams-mbp:triggers_prototype aroberts$ k logs github-validate-run-89mhl-pod-195926 --all-containers
{"level":"warn","ts":1570624241.7803333,"logger":"fallback-logger","caller":"logging/config.go:69","msg":"Fetch GitHub commit ID from kodata failed: \"ref: refs/heads/master\" is not a valid GitHub commit ID"}
{"level":"info","ts":1570624241.781536,"logger":"fallback-logger","caller":"creds-init/main.go:40","msg":"Credentials initialized."}
service/my-interceptor-service created
sh: syntax error: unexpected "fi"
```

Same with the other

docs/create-gh-validate-interceptor.yaml
```diff
@@ -0,0 +1,74 @@
...
+              matchLabels:
+                app: $(inputs.params.Name)
+            template:
+              metadata:
+                labels:
+                  app: $(inputs.params.Name)
+              spec:
+                containers:
+                  - name: validate
+                    image: $(input.params.ContainerImage)
```
a-roberts: inputs* everywhere

docs/create-gh-validate-interceptor.yaml
dibyom: So, I tried adding some docs in https://github.com/tektoncd/triggers/blob/6e81d0f3dea0497f70e32cd5dc30285703d451e7/examples/event-interceptors/README.md

but yeah having the task in docs/ and the actual container code in examples/ looks confusing. Open to ideas here. The other tasks are in docs/ so I was following the pattern. Maybe we should have a folder for these tasks somewhere?

vtereso:
Closing as per https://github.com/tektoncd/triggers/pull/160.

PR #152, opened 2019-10-08
Document exposing EL with nginx ingress

dibyom:
# Changes

Part of #153 

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -9,14 +9,15 @@ with.
 When an `EventListener` is successfully created, a service is created that
 references a listener pod. This listener pod accepts the incoming events and
 does what has been specified in the corresponding
-`TriggerBinding`s/`TriggerTemplate`s.
-
+`TriggerBinding`s/`TriggerTemplate`s. The service created is a `ClusterIP` service 
+which means any other pods running in the same Kubernetes cluster can access it via the service's 
+cluster DNS. For external services to connect to your cluster (e.g. GitHub 
+sending webhooks), check out the guide on [exposing eventlisteners](./exposing-eventlisteners.md) 
 
 ## Parameters
 `EventListener`s can provide `params` which are merged with the `TriggerBinding`
 `params` and passed to the `TriggerTemplate`. Each parameter has a `name` and a
 `value`.
-
 
 ## Event Interceptors
 
--- a/docs/exposing-eventlisteners.md
+++ b/docs/exposing-eventlisteners.md
@@ -1,0 +1,54 @@
+# Exposing EventListeners Externally
+
+By default, `ClusterIP` services such as the EventListener sink are accessible 
+within the cluster. There are a few ways of exposing it so that external services
+can talk to it:
+
+
+## Using an Ingress
+
+You can use an Ingress resource to expose the EventListener. The [`create-ingress`](./create-ingress.yaml)
+Tekton task can help setup an ingress resource using self-signed certs.
+
+**Note**: If you are using a cloud hosted Kubernetes solution such as GKE, the built-in ingress will not work
+with `ClusterIP` services. Instead, you can use the Nginx Ingress based approach below.
+
+
+## Using Nginx Ingress
+
+The following instructions have been tested on GKE cluster running version `1.13.7-gke.24`.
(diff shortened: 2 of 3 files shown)


dibyom:
@a-roberts thanks for trying this out. Definitely a WIP but what'a bit messy here is that the instructions differ based on your k8s installation. It works on GKE and it *should* work on docker/minikube  
 
We need this if you are using a cloud provider hosted k8s like GKE, you'll have to install and use nginx ingress. The reason being `ClusterIP` services like the EventListenerSinks are not supported by the default cloud provider Ingresses: https://github.com/kubernetes/kubernetes/issues/26508#issuecomment-222376962

For other envrionments like Docker Desktop/minikube, the ingress task might be simpler to use.

a-roberts:
> @a-roberts thanks for trying this out. Definitely a WIP but what'a bit messy here is that the instructions differ based on your k8s installation. It works on GKE and it _should_ work on docker/minikube
> 
> We need this if you are using a cloud provider hosted k8s like GKE, you'll have to install and use nginx ingress. The reason being `ClusterIP` services like the EventListenerSinks are not supported by the default cloud provider Ingresses: [kubernetes/kubernetes#26508 (comment)](https://github.com/kubernetes/kubernetes/issues/26508#issuecomment-222376962)
> 
> For other envrionments like Docker Desktop/minikube, the ingress task might be simpler to use.

Yeah, my comment was more: let's provide some known good configuration, e.g. "make sure you're running with *X* version of Kubernetes and Docker - a known good configuration is Docker 2.1.3.0 and Kubernetes 1.14.6 for example - more testing and documentation welcome!" - it took me ages to figure out why I had no address and it's because I was running old versions of both

Would be nice to know the versions of those (Docker and Kubernetes) on GKE too which I assume is the test environment you've been using 😄

docs/exposing-eventlisteners.md
```diff
@@ -0,0 +1,58 @@
...
+Instructions for installing nginx Ingress on other Kubernetes services can be found [here](https://kubernetes.github.io/ingress-nginx/deploy/).
+
+1. First, install Nginx ingress controller:
+
+```yaml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/mandatory.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/cloud-generic.yaml
+```
+
+1. Find the service name expose by the GitHub service:
```
vtereso: Nit: This is still correct, but the service name is known ahead of time now since we have a static prefix.

docs/eventlisteners.md
```diff
@@ -9,15 +9,16 @@ with.
 When an `EventListener` is successfully created, a service is created that
 references a listener pod. This listener pod accepts the incoming events and
 does what has been specified in the corresponding
-`TriggerBinding`s/`TriggerTemplate`s.
-
+`TriggerBinding`s/`TriggerTemplate`s. The service created is a `ClusterIP` service
```
vtereso: Spoke with you on this, will be updated if/when the service type configuration PR is merged.

docs/exposing-eventlisteners.md
```diff
@@ -0,0 +1,58 @@
...
+**Note**: If you are using a cloud hosted Kubernetes solution such as GKE, the built-in ingress will not work
+with `ClusterIP` services. Instead, you can use the Nginx Ingress based approach below.
+
+
+## Using Nginx Ingress
+
+The following instructions have been tested on GKE cluster running version `1.13.7-gke.24`.
+Instructions for installing nginx Ingress on other Kubernetes services can be found [here](https://kubernetes.github.io/ingress-nginx/deploy/).
+
+1. First, install Nginx ingress controller:
```
vtereso: Nit: Indentation is off? Everything is step 1.

docs/eventlisteners.md
dibyom: ack, adding it once we have #175 merged

docs/exposing-eventlisteners.md
dibyom: done!

Issue #112, opened 2019-09-12
Add logging for handling events 


a-roberts:
# Expected Behavior
We should provide useful logging in https://github.com/tektoncd/triggers/blob/master/pkg/sink/sink.go#L43 (in my case I had a PipelineRun and PipelineResources created, it just wasn't immediately obvious to me - and tailing the listener logs didn't give me any clues

# Actual Behavior
I expect to see, for example: created PipelineRun named x with PipelineResources (a, b, c)

# Steps to Reproduce the Problem

1. Create an event listener and successfully create a PipelineRun and PipelineResources using the example code for this repository (I'm planning to add some docs to make this easy)

# Additional info
This would be an awesome first issue

akihikokuroda:
I added log entry for each resource generation.  If it fails, the error log entry follows the log entry.
```
2019/10/30 19:38:11 Generating resource: kind: PipelineResource, name: source-repo-jgcpl 
2019/10/30 19:38:11 Generating resource: kind: PipelineResource, name: image-source-jgcpl 
2019/10/30 19:38:11 Generating resource: kind: Task, name: devopsascode-task-jgcpl 
2019/10/30 19:38:11 Generating resource: kind: TaskRun, name: devopsascode-taskrun-1 
2019/10/30 19:38:11 taskruns.tekton.dev "devopsascode-taskrun-1" already exists
```

akihikokuroda:
PR: https://github.com/tektoncd/triggers/pull/197

PR #157, opened 2019-10-09
Access headers within TriggerBinding

ncskier:
# Changes

- Added `$(header)` and `$(header.X)` as interpolation variables that will be
substituted for header values from the event.
- Changed `event` interpolation variable keyword to `body`.
  - `$(event)` and `$(event.X.Y.Z)` interpolation variables are now `$(body)` and `$(body.X.Y.Z)`.
- Escape `"` with `\"` in `body` and `header` interpolation variables.
  - This fixes a bug I found where resource templates failed to be created when JSON objects were substituted for interpolation variables; for example, when `$(event)` is used.
  - Added an e2e test to cover passing JSON objects as parameters.

# Submitter Checklist

review them:

# Release Notes

```
Add $(header.X) variable interpolation for TriggerBindings.
$(event) and $(event.X.Y.Z) interpolation variables are now $(body) and $(body.X.Y.Z)
```

### DIFF
--- a/docs/triggerbindings.md
+++ b/docs/triggerbindings.md
@@ -13,9 +13,11 @@ metadata:
 spec:
   params:
   - name: gitrevision
-    value: $(event.head_commit.id)
+    value: $(body.head_commit.id)
   - name: gitrepositoryurl
-    value: $(event.repository.url)
+    value: $(body.repository.url)
+  - name: contenttype
+    value: $(header.Content-Type)
 ```
 
 `TriggerBinding`s are connected to `TriggerTemplate`s within an [`EventListener`](eventlisteners.md), which is where the pod is actually instantiated that "listens" for the respective events.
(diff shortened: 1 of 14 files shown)


docs/triggerbindings.md
```diff
@@ -13,13 +13,51 @@ metadata:
...
 `TriggerBinding`s are connected to `TriggerTemplate`s within an [`EventListener`](eventlisteners.md), which is where the pod is actually instantiated that "listens" for the respective events.
 
 ## Parameters
 `TriggerBinding`s can provide `params` which are passed to a
 `TriggerTemplate`. Each parameter has a `name` and a `value`.
+
+
+## Event Variable Interpolation
+
+### Body
```
dibyom: Nice docs! Quick question: Does it work with arrays e.g. can I do $(body.key[1].key2)
Also, is this syntax different from JSONPath?

ncskier: Thanks for taking a look @dibyom!
I haven't changed the event/body path functionality, so we currently can't do `$(body.key[1].key2)`. We're using the [gjson](https://github.com/tidwall/gjson) library, so we could technically do anything supported by gjson, however, right now we restrict body paths to [this regex](https://github.com/tektoncd/triggers/blob/master/pkg/template/event.go#L30) `\$\(body(.[0-9A-Za-z_-]+)*\)`.

ncskier: I don't think that gjson uses the JSONPath syntax though.

dibyom: cool cool! as long as it is some well known syntax :+1:

ncskier:
Thanks for the review Adam! 🙂 
Also, I forgot to squash the commits, so I just did that 👍

docs/triggerbindings.md
```diff
@@ -13,13 +13,51 @@ metadata:
...
+
+### Header
+HTTP Post request header data can be referenced using variable interpolation.
+Text in the form of `$(header.X)` is replaced by the event's header named `X`.
+
+`$(header)` is replaced by all of the headers from the event.
+
+The following are some example variable interpolation replacements:
+```
+$(header) -> {"One":["one"], "Three":["one","two","three"]}
```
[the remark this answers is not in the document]
ncskier: Thanks, good catch!

pkg/sink/sink.go
```diff
@@ -157,6 +157,8 @@ func createResource(rt json.RawMessage, restClient restclient.Interface, discove
 		return err
 	}
 
+	log.Printf("Create resource: \n%s", string(rt))
```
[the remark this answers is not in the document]
ncskier: Good point. I removed the log. Plus, we should handle logging in issue https://github.com/tektoncd/triggers/issues/112

docs/triggerbindings.md
ncskier: At the moment, I don't think that we should advertise the gjson syntax, because we currently don't allow users to use the full gjson syntax. We only allow `$(body.X.Y.Z)` right now.

PR #172, opened 2019-10-11
Fix interceptor headers to be a deepcopy

ncskier:
# Changes

The interceptor code was doing a shallowcopy of the headers, so when a
header parameter was added to the interceptor, it would persist in the
request processed by the sink after the interceptor.

I added a test to verify that the original request headers are not
modified by the interceptor parameters.

# Submitter Checklist

review them:

### DIFF
--- a/pkg/sink/interceptors.go
+++ b/pkg/sink/interceptors.go
@@ -49,17 +49,18 @@ func createOutgoingRequest(ctx context.Context, original *http.Request, url *url
 	r.RequestURI = ""
 	r.URL = url
 
-	headers := make(map[string][]string)
+	headers := make(map[string][]string, len(original.Header))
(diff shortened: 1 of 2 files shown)


pkg/sink/interceptors.go
```diff
@@ -49,17 +49,18 @@ func createOutgoingRequest(ctx context.Context, original *http.Request, url *url
...
 		copy(v2, v)
 		headers[k] = v2
 	}
+	r.Header = headers
 
 	if s := original.TransferEncoding; s != nil {
 		s2 := make([]string, len(s))
 		copy(s2, s)
-		original.TransferEncoding = s
+		r.TransferEncoding = s
```
dibyom: 😬

PR #180, opened 2019-10-16
Add Go modules support.

wlynch:
# Changes

Adds Go modules support for triggers repo.

This follows a similar pattern as https://github.com/tektoncd/pipeline/pull/1607. Vendor directory in place for now - would like to remove it, but requires further discussion.

# Submitter Checklist

review them:

# Release Notes

n/a


go.mod
```diff
@@ -0,0 +1,84 @@
+module github.com/tektoncd/triggers
+
+go 1.13
```
dibyom: Hmm, does this need go 1.13? I don't think our test runner works with 1.13 yet

wlynch: Not sure. This was auto generated by the `go mod init`. We can try bumping this down, but I suspect we might be fighting the tool to keep this out.

Any reason we can't bump the test runner to 1.13?

.gitignore
```diff
@@ -14,7 +14,6 @@
 *.dll
 
 # Fortran module files
-*.mod
```
bobcatfish: image

go.mod
bobcatfish: looks like @dibyom went ahead and bumped it? https://github.com/tektoncd/plumbing/pull/88 :D :tada:

.gitignore
bobcatfish: ooooooh i get it (for posterity: it looked like we were randomly changing the **Fortran** gitignore and i didnt understand why - but of course, the go module file ends with `.mod` so we need to stop ignoring it. (that must have been an interesting one to track down @wlynch !!)

but i really liked this cat so i dont want to delete this comment... :innocent:

wlynch:
> * Would it make sense to include a documentation component to this PR as well, like something in [DEVELOPMENT.md](https://github.com/tektoncd/triggers/blob/master/DEVELOPMENT.md)? (e.g. i dont know how to interact with this `go.mod` file at all)

Done.

> * What are the next steps before we can update [update-deps.sh](https://github.com/tektoncd/triggers/blob/master/hack/update-deps.sh) to use go modules instead of `go dep`, and remove the files we maintain for `go dep`

Done.

> * Can we totally switch over right now or should we make it so that  [update-deps.sh](https://github.com/tektoncd/triggers/blob/master/hack/update-deps.sh) ensures that we are keeping both lists of dependencies up to date?

Done.

go.mod
vdemeester: I think it's fine having 1.13 in there :wink:

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "vendor/github.com/tektoncd/plumbing"]
```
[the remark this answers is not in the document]
khrm: We also need to be consistent in whether having vendor folder or not. We are still using vendor folder in cli.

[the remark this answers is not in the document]
bobcatfish: hm interesting! I think this highlights something very odd (and historical) about our plumbing repo - we were using go to vendor in bash scripts basically :thinking: (is there anything besides the bash scripts we need from plumbing?)

(My experiences with git submodules in the past were painful but I can't remember the specifics so maybe it's okay?)

I think no matter what we should work toward _not_ sharing scripts from plumbing like this, so this is more motivation to migrate from the bash scripts to some combo of
- Tekton Tasks (shared via one of @vdemeester 's catalog designs? https://github.com/tektoncd/pipeline/issues/964)
- Code (I thought python made a lot of sense but then we're going to have to start publishing python packages? maybe sticking to Go makes more sense)
- Images

Couple ideas:
1. Use git submodule for now
2. Update whatever actually needs the plumbing scripts to fetch them (probably our .sh e2e scripts)
3. Hold off on this until we migrate away from the bash scripts

I'm partial to (2) tho it's more work but would settle for (1) (can always do 2 as a next step :))

wlynch:
/hold

Putting this on hold to figure out the vendoring of tektoncd/plumbing.

As @bobcatfish mentioned, the only usage of the vendor directory for this repo is vendoring for the bash scripts for testing, as well as some complications around LICENSE checks. You can `go build ./...` and `go test ./...` this repo via go modules with without the vendor directory, which is great!

I'll see what I can do with the presubmit scripts. If it proves to be too difficult I'll bring back the vendor directory, with [`-mod=vendor`](https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away)

.gitmodules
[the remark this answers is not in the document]
chmouel: dunno about bash or python (althought I rather do the later than the former) but I usually just black-out everytime there is a mention of go and modules or even just go and managing dependencies...

PR #183, opened 2019-10-21
Support more JSON array query syntax for the body

afrittoli:
Extend the supported filtering syntax to include queries on arrays,
which can be useful when dealing with JSON that describes
k8s resources.

This adds the bare minimum from GJSON syntax to support array
queries, trying to avoid exposing too much of GJSON syntax.

Out of the full syntax defined in https://github.com/tidwall/gjson/blob/master/SYNTAX.md
we would have the following support list:
- Basic: already supported
- Queries: support added here, but no nested queries
- Wildcards: support added here, but only for array filtering (needed for the % operator)
- Escape character: not supported
- Arrays: not supported
- Dot vs Pipe: not supported (only pipe)
- Modifiers: not supported
- Multipaths: not supported

# Changes

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/docs/triggerbindings.md
+++ b/docs/triggerbindings.md
@@ -29,12 +29,12 @@ spec:
 
 ## Event Variable Interpolation
 
-In order to parse generic events as efficiently as possible, [GJSON](https://github.com/tidwall/gjson) 
+In order to parse generic events as efficiently as possible, [GJSON](https://github.com/tidwall/gjson)
 is used internally. As a result, the binding [path syntax](https://github.com/tidwall/gjson#path-syntax)
 differs slightly from standard JSON. As of now, the following patterns are
 supported within `TriggerBinding` parameter value interpolation:
-- `$(body(.[0-9A-Za-z_-]+)*)`
-- `$(header(.[0-9A-Za-z_-]+)?)`
+- `\$\(body(\.[[:alnum:]/_\-\.\\]+|\.#\([[:alnum:]=<>%!"\*_-]+\)#??)*\)`
+- `\$\(header(\.[[:alnum:]_\-]+)?\)`
 
 ### Body
 HTTP Post request body data can be referenced using variable interpolation.
--- a/pkg/template/event.go
+++ b/pkg/template/event.go
@@ -27,8 +27,14 @@ import (
 )
 
 // bodyPathVarRegex determines valid body path variables
-var bodyPathVarRegex = regexp.MustCompile(`\$\(body(.[0-9A-Za-z_-]+)*\)`)
-var headerVarRegex = regexp.MustCompile(`\$\(header(.[0-9A-Za-z_-]+)?\)`)
+// The body regular expression allows for a subset of GJSON syntax, the mininum
+// required to navigate through dictionaries, query arrays and support
+// namespaced label names e.g. tekton.dev/eventlistener
+var bodyPathVarRegex = regexp.MustCompile(`\$\(body(\.[[:alnum:]/_\-\.\\]+|\.#\([[:alnum:]=<>%!"\*_-]+\)#??)*\)`)
+
+// The headers regular expression allows for simple navigation down a hierarchy
+// of dictionaries
+var headerVarRegex = regexp.MustCompile(`\$\(header(\.[[:alnum:]_\-]+)?\)`)
 
 // getBodyPathFromVar returns the body path given an body path variable
 // $(body.my.path) -> my.path
--- a/pkg/template/event_test.go
+++ b/pkg/template/event_test.go
@@ -41,6 +41,11 @@ func Test_BodyPathVarRegex(t *testing.T) {
 		"$(body.a1)",
 		"$(body.a.b)",
 		"$(body.a.b.c)",
+		"$(body.1.b.c\\.e/f)",
+		"$(body.#(a==b))",
+		"$(body.#(a>1)#)",
+		"$(body.#(a%\"D*\")#.c)",
+		"$(body.#(a!%\"D*\").c)",
 	}
 	for _, bodyPathVar := range tests {
 		t.Run(bodyPathVar, func(t *testing.T) {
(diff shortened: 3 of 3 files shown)


docs/triggerbindings.md
```diff
@@ -29,12 +29,12 @@ spec:
...
 
-In order to parse generic events as efficiently as possible, [GJSON](https://github.com/tidwall/gjson) 
+In order to parse generic events as efficiently as possible, [GJSON](https://github.com/tidwall/gjson)
 is used internally. As a result, the binding [path syntax](https://github.com/tidwall/gjson#path-syntax)
 differs slightly from standard JSON. As of now, the following patterns are
 supported within `TriggerBinding` parameter value interpolation:
-- `$(body(.[0-9A-Za-z_-]+)*)`
-- `$(header(.[0-9A-Za-z_-]+)?)`
+- `$(body([.|][[:alnum:]*?_-]+|[.|]#|[.|]#\([[:alnum:]=<>%!*?"_-]+\)#??)*)`
+- `$(header([.|][:alnum:]+)?)`
```
ncskier: This could be a naive question, but what does `[.|]` mean?

Also, do you think it might be more clear to include the actual regex
```
regexp.MustCompile(`\$\(body(\.[[:alnum:]/_\-\.\\]+|\.#\([[:alnum:]=<>%!"\*_-]+\)#??)*\)`)
```
in the docs instead of using a pseudo regex?

pkg/template/event.go
```diff
@@ -27,8 +27,13 @@ import (
...
 
 // bodyPathVarRegex determines valid body path variables
-var bodyPathVarRegex = regexp.MustCompile(`\$\(body(.[0-9A-Za-z_-]+)*\)`)
-var headerVarRegex = regexp.MustCompile(`\$\(header(.[0-9A-Za-z_-]+)?\)`)
+// The body regular expression allows for a subset of GJSON syntax, the mininum
+// required to navigate through dictionaries, query arrays and support
+// namespaced label names e.g. tekton.dev/eventlistener
+var bodyPathVarRegex = regexp.MustCompile(`\$\(body(\.[[:alnum:]/_\-\.\\]+|\.#\([[:alnum:]=<>%!"\*_-]+\)#??)*\)`)
+
+// The headers regular expression allows for simple
```
ncskier: I think that this comment got cut off or something?

pkg/template/event_test.go
```diff
@@ -57,8 +62,10 @@ func Test_BodyPathVarRegex_invalid(t *testing.T) {
 		"$[body]",
 		"${body}",
 		"$(body.)",
-		"$(body..)",
```
ncskier: Why was this test case removed? I think `$(body..)` would be invalid, right?

pkg/template/event.go
```diff
@@ -27,8 +27,13 @@ import (
 )
 
 // bodyPathVarRegex determines valid body path variables
-var bodyPathVarRegex = regexp.MustCompile(`\$\(body(.[0-9A-Za-z_-]+)*\)`)
-var headerVarRegex = regexp.MustCompile(`\$\(header(.[0-9A-Za-z_-]+)?\)`)
+// The body regular expression allows for a subset of GJSON syntax, the mininum
+// required to navigate through dictionaries, query arrays and support
+// namespaced label names e.g. tekton.dev/eventlistener
+var bodyPathVarRegex = regexp.MustCompile(`\$\(body(\.[[:alnum:]/_\-\.\\]+|\.#\([[:alnum:]=<>%!"\*_-]+\)#??)*\)`)
```
ncskier: I think that the only metacharacters that we need to escape are `*+?()|` (https://github.com/google/re2/wiki/Syntax)

docs/triggerbindings.md
afrittoli: Nice catch, I removed [.|] from the actual regex but forgot to remove it from the docs.
GJSON allows you to traverse dicts using `.` or `|`, but I removed `|` to minimize the surface of GJSON syntax exposed.
I will fix the docs with the actual regex.

pkg/template/event.go
afrittoli: Good point, thanks!

pkg/template/event_test.go
afrittoli: Yes indeed, however if we want to support having a `.` in the key of a dict then we cannot easily avoid that, unless we split the regex to say what the first character may or may not be and then every other character from the second onward. I though it would not be worth it.
I added more used cases though that used to be considered valid before and are now not valid anymore.

pkg/template/event.go
afrittoli: It is not accurate. It is now more accurate than it was before but a regex is not guarantee that the body is valid GJSON. It would be helpful if GJSON provided a validation function.

dibyom: Yeah, validating using regex is not ideal though it does help with some use cases. So my vote is to keep it though it is not a guarantee that the GJSON is correct.  Ideally, like @afrittoli said, the library should provide a validation function!

pkg/template/event_test.go
ncskier: I don't think `gjson` works with `.` as a key of a dict, because it messes up the syntax; here's a go playground link: https://play.golang.org/p/OTQkYyFSJaM. Let me know if I messed up the syntax or am missing something here.

afrittoli: The "." is a valid character as long as it's escaped.
See the syntax reference https://github.com/tidwall/gjson/blob/master/SYNTAX.md#escape-character

The "." is important to filter labels, that often are namespaced and the namespace includes a ".", for instance:

```
spec:
  params:
  - name: pipelineRun
    value: $(body.Data.taskRun.metadata.labels.tekton\.dev/pipelineRun)
```

ncskier: So, that might be a  bug in the gjson library. I don't think it works when I try to escape the `.` character: https://play.golang.org/p/8f67D3_JTrU

afrittoli: heh, you need to escape the back-slash for go, else go thinks you're trying to define a string an escape sequence "\." rather than a string with the "\" character followed by the "." one. 
https://play.golang.org/p/16L6fO6LHMT

ncskier: Ohhhhh, lol, thanks! 😅

PR #190, opened 2019-10-28
Update cron example with new EventListener prefix.

wlynch:
# Changes

The EventListener prefix was updated in https://github.com/tektoncd/triggers/commit/ecd97356c856ad73499fb383ae4ed8010fe48ee0, which broke this example. This updates the service name to match the change.

# Submitter Checklist

review them:

### DIFF
--- a/examples/cron/cronjob.yaml
+++ b/examples/cron/cronjob.yaml
@@ -14,5 +14,5 @@ spec:
             args:
               - wget
               - --spider
-              - el--cron-listener.default.svc.cluster.local:8080
+              - el-cron-listener.default.svc.cluster.local:8080
           restartPolicy: Never


wlynch:
> Sorry, that was my PR that broke this 😬
> 
> Thanks for catching it!
> /lgtm

No worries! We should find a way to add a test for these examples to make sure they stay up to date. Things like cron may be tricky though since we need to wait around for a run to happen. 😓

ncskier:
Yeah, I agree that it would be great if we could get some e2e tests going for our examples. In particular the cron and the GitHub webhook (maybe with validation) examples should be tested. There are going to be some obstacles with setting up these tests, but I think that it will be beneficial in the long-run.

PR #191, opened 2019-10-28
Allow multiple bindings in triggers.

### LINKED ISSUE #176: Proposal: Allow multiple TriggerBindings in Triggers
# Expected Behavior

Allow triggers to be composed from multiple trigger bindings.

Motivating example: Have one trigger binding that extracts relevant parameters from the incoming event and provide a second binding that provides environment specific parameters (e.g. where to run/deploy).

Example:

```yaml
kind: TriggerBinding
metadata:
  name: event-binding
spec:
  params:
  - name: gitrevision
    value: $(body.head_commit.id)
  - name: gitrepositoryurl
    value: $(body.repository.url)
---
kind: TriggerBinding
metadata:
  name: prod-env
spec:
  params:
  - name: environment
    value: prod
---
kind: EventListener
metadata:
  name: listener
spec:
  triggers:
    - name: prod-trigger
      binding:
        - name: event-binding
        - name: prod-env
      template:
        name: pipeline-template
    - name: staging-trigger
      binding:
        - name: event-binding
        - name: staging-env
      template:
        name: pipeline-template
```

# Actual Behavior

Triggers only allow specifying one TriggerBinding.

# Additional Info

This is similar to #87, but focuses on just allowing multiple bindings without introducing a new ParamBinding type.

Things to think about:
* What happens if multiple trigger bindings specify the same param?

wlynch:
# Changes

Allow multiple bindings in triggers.

BREAKING CHANGE: Changes field name of TriggerBinding "binding" =>
"bindings" and changes the underlying type from a single object to a
list.

Fixes #176.

# Submitter Checklist

review them:

# Release Notes

```
Triggers now support multiple bindings.

This enables multiple bindings to be specified in a trigger, enabling
users to break bindings into composable pieces (e.g. one binding for
extracting data from the incoming payload, another for specifying
environment information).


BREAKING CHANGE: Changes field name of TriggerBinding "binding" =>
"bindings" and changes the underlying type from a single object to a
list.
```


pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -61,9 +61,9 @@ type EventListenerSpec struct {
 // and TriggerTemplate; TriggerBinding provides extracted values for
 // TriggerTemplate to then create resources from.
 type EventListenerTrigger struct {
-	Binding  *EventListenerBinding `json:"binding"`
-	Template EventListenerTemplate `json:"template"`
-	Params   []pipelinev1.Param    `json:"params,omitempty"`
+	Bindings []*EventListenerBinding `json:"bindings"`
+	Template EventListenerTemplate   `json:"template"`
+	Params   []pipelinev1.Param      `json:"params,omitempty"`
```
ncskier: Should we remove `params` from the EventListener? I think that this change to allow multiple TriggerBindings has overlapping functionality with the EventListener params. Now a user can use a TriggerBinding instead of the EventListener params, and I think that it might be confusing to have both options.

The direction that this works seems to be heading towards is embedding TriggerBindings directly in an EventListener yaml (which would completely replace the params).

wlynch: As mentioned in the Triggers WG meeting, I think this is a good idea!

I think that should be done in a separate change though, just to minimize size and the breaking changes of this PR. Happy to file an issue for this once this PR is submitted! ^_^

dibyom: +1 to both  -- let's remove Params in a diff PR!

pkg/template/event_test.go
```diff
@@ -920,6 +935,37 @@ func Test_NewResources(t *testing.T) {
...
+								bldr.TriggerBindingParam("param2", "$(header.one)"),
+							),
+						),
+					},
+				},
+			},
+			want: []json.RawMessage{
+				json.RawMessage(`{"rt1": "bar-1"}`),
+			},
+		},
```
bobcatfish: maybe not something you want to tackle now @wlynch but i think unit testing at the level of `NewResources` is leading to the test cases being a long list of pretty involved setup + expected state; those kinda tests are hard to read and maintain over time, so i think we should start preferring more focused tests - e.g. test an example or end to end test that uses multiple bindings, then have unit tests for individual functions such as `mergeBindingParams`

pkg/template/resource.go
```diff
@@ -45,14 +45,13 @@ type getTriggerTemplate func(name string, options metav1.GetOptions) (*triggersv
 // ResolveBinding takes in a trigger containing object refs to bindings and
 // templates and resolves them to their underlying values.
 func ResolveBinding(trigger triggersv1.EventListenerTrigger, getTB getTriggerBinding, getTT getTriggerTemplate) (ResolvedBinding, error) {
```
bobcatfish: should this be called `ResolveBindings` now?

pkg/template/event_test.go
wlynch: > * Could the docs talk about this feature, and why you'd want to use it?

Done.

> * The commit message (and probably the relase note too), I think they could include [more details](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages) around _why_ we want to allow multiple trigger bindings (e.g. use cases we want to support) (the phrase "multiple bindings" may not mean much to folks without more info)

Done

> And one question: do you know what behavior is if someone tried to use an event listener still using `binding` instead of `bindings`? I'm just thinking about our dogfooding, even tho we're still very very alpha, i wonder if it would make sense to raise an error in that case so we dont just get unexpected behavior we need to debug (maybe this is how it works already?)

There's validation before the resources are accepted, so you get an error like:
```
Error from server (BadRequest): error when creating "eventlistener.yaml": admission webhook "triggers-webhook.tekton.dev" denied the request: mutation failed: invalid value: triggerbindings.tekton.dev "pipeline-binding" not found: spec.triggers[0].binding.name
```

pkg/template/resource.go
wlynch: Since this resolves Bindings + Template, I don't want it to seem like we'd also support multiple Templates. I think the verbage is fine with respect to handling multiple bindings, but I did go ahead and change `ResolvedBinding.TriggerBinding` -> `ResolvedBinding.TriggerBindings`.

wlynch:
> * Could the docs talk about this feature, and why you'd want to use it?

Done.

> * The commit message (and probably the relase note too), I think they could include [more details](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages) around _why_ we want to allow multiple trigger bindings (e.g. use cases we want to support) (the phrase "multiple bindings" may not mean much to folks without more info)

Done

> And one question: do you know what behavior is if someone tried to use an event listener still using `binding` instead of `bindings`? I'm just thinking about our dogfooding, even tho we're still very very alpha, i wonder if it would make sense to raise an error in that case so we dont just get unexpected behavior we need to debug (maybe this is how it works already?)

There's validation before the resources are accepted, so you get an error like:
```
Error from server (BadRequest): error when creating "eventlistener.yaml": admission webhook "triggers-webhook.tekton.dev" denied the request: mutation failed: invalid value: triggerbindings.tekton.dev "pipeline-binding" not found: spec.triggers[0].binding.name
```

docs/triggerbindings.md
```diff
@@ -20,55 +21,116 @@ spec:
...
+    value: staging
+---
+apiVersion: tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: listener
+spec:
+  triggers:
+    - name: prod-trigger
+      binding:
```
dibbles: shouldn't this say bindings now?

docs/triggerbindings.md
```diff
@@ -20,55 +21,116 @@ spec:
...
+        - name: event-binding
+        - name: prod-env
+      template:
+        name: pipeline-template
+    - name: staging-trigger
+      binding:
```
dibbles: here too ..... shouldn't this say bindings now?

docs/triggerbindings.md
ncskier: Yep, good point 😁 I'll create a PR to fix it.

ncskier: https://github.com/tektoncd/triggers/pull/210

PR #193, opened 2019-10-29
Update EventListener Addressable to include port

Fabian-K:
# Changes
Previously the EventListener did not include the port information for its addressable contract. As a result, the EventListener could not be used as a Sink for Knative Eventing as the EventListener pod listens on 8080 however Knative assumed 80.

# Release Notes

```
Update the EventListener Addressable to include the correct port information for use as a Knative Sink using the `Status.Address.URL field`  

```

### DIFF
--- a/pkg/apis/triggers/v1alpha1/event_listener_types.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_types.go
@@ -227,12 +227,11 @@ func (els *EventListenerStatus) SetAddress(hostname string) {
 		els.Address = &duckv1alpha1.Addressable{}
 	}
 	if hostname != "" {
-		els.Address.Hostname = hostname
-		if u, err := apis.ParseURL(fmt.Sprintf("http://%s/", hostname)); err != nil {
-			els.Address.URL = u
+		els.Address.URL = &apis.URL{
+			Scheme: "http",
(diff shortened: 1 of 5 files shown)


Fabian-K:
I´m not sure why pull-tekton-triggers-build-tests fails. Is this related to the code changes?

Fabian-K:
hm... it looks like a timeout failed the test (`Timeout exceeded: try increase it by passing --timeout option`).

/retest

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -227,12 +227,11 @@ func (els *EventListenerStatus) SetAddress(hostname string) {
 		els.Address = &duckv1alpha1.Addressable{}
 	}
 	if hostname != "" {
-		els.Address.Hostname = hostname
-		if u, err := apis.ParseURL(fmt.Sprintf("http://%s/", hostname)); err != nil {
-			els.Address.URL = u
+		els.Address.URL = &apis.URL{
+			Scheme: "http",
```
dibyom: So, the status now only has Adress.URL and no Address.Hostname. I think this is totally fine. Just worth calling this out as a breaking change in the `Release Notes` section of the PR!

Issue #111, opened 2019-09-12
Implement EventListener response codes (e.g. bad request on empty)


a-roberts:
# Expected Behavior
If I curl my event listener with no data, I should get back an error that I've sent it a bad request - and be told what the format should be. Noticed this when working https://github.com/tektoncd/triggers/issues/109 (figuring out what data I should be sending and for fun, sending nothing to see what I'd get back)

# Actual Behavior
I get an empty response and a return code 200

# Steps to Reproduce the Problem

1. Apply example resources from this repository
2. `kubectl port-forward *listener pod name* 8082`
3. curl -X POST http://localhost:8082
4. Observe you get a 200 back with no data

In Postman:
![image](https://user-images.githubusercontent.com/12814972/64776930-0ab1c300-d551-11e9-9f86-f4dd737a7101.png)

# Additional Info
Fix to go in https://github.com/tektoncd/triggers/blob/2f2c62517fdc45a1c0bed3882a8d27d9a8cceec1/pkg/sink/sink.go#L43 I expect

akihikokuroda:
PR: https://github.com/tektoncd/triggers/pull/198

ncskier:
Thanks @akihikokuroda 👍 The PR was merged, so I'm going to close this issue.

PR #198, opened 2019-10-31
Eventlistener response code change

akihikokuroda:
# Changes

This PR is for issue https://github.com/tektoncd/triggers/issues/111

The eventlistener waits until all the trigger executions (up-to the creation of the resources) and only when at least one of the execution completed successfully, it returns response code 201(Accepted) otherwise it returns 202 (Created).

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -72,27 +72,41 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
 	log.Printf("EventListener: %s in Namespace: %s handling event (EventID: %s) with payload: %s and header: %v",
 		r.EventListenerName, r.EventListenerNamespace, eventID, string(event), request.Header)
 
+	result := make(chan int, 10)
 	// Execute each Trigger
 	for _, trigger := range el.Spec.Triggers {
-		go r.executeTrigger(event, request, trigger, eventID)
+		go r.executeTrigger(event, request, trigger, eventID, result)
+	}
+
+	//The eventlistener waits until all the trigger executions (up-to the creation of the resources) and
+	//only when at least one of the execution completed successfully, it returns response code 201(Accepted) otherwise it returns 202 (Created).
+	code := http.StatusAccepted
+	for i := 0; i < len(el.Spec.Triggers); i++ {
+		thiscode := <-result
+		if thiscode < code {
+			code = thiscode
+		}
 	}
 
 	// TODO: Do we really need to return the entire body back???
+	response.WriteHeader(code)
 	fmt.Fprintf(response, "EventListener: %s in Namespace: %s handling event (EventID: %s) with payload: %s and header: %v",
 		r.EventListenerName, r.EventListenerNamespace, string(eventID), string(event), request.Header)
 }
 
-func (r Resource) executeTrigger(payload []byte, request *http.Request, trigger triggersv1.EventListenerTrigger, eventID string) {
+func (r Resource) executeTrigger(payload []byte, request *http.Request, trigger triggersv1.EventListenerTrigger, eventID string, result chan int) {
 	if trigger.Interceptor != nil {
 		interceptorURL, err := GetURI(trigger.Interceptor.ObjectRef, r.EventListenerNamespace) // TODO: Cache this result or do this on initialization
 		if err != nil {
 			log.Printf("Could not resolve Interceptor Service URI: %q", err)
+			result <- http.StatusAccepted
 			return
 		}
 
 		modifiedPayload, err := r.processEvent(interceptorURL, request, payload, trigger.Interceptor.Header)
 		if err != nil {
 			log.Printf("Error Intercepting Event (EventID: %s): %q", eventID, err)
+			result <- http.StatusAccepted
 			return
 		}
 		payload = modifiedPayload
@@ -103,17 +117,21 @@ func (r Resource) executeTrigger(payload []byte, request *http.Request, trigger 
 		r.TriggersClient.TektonV1alpha1().TriggerTemplates(r.EventListenerNamespace).Get)
 	if err != nil {
 		log.Print(err)
+		result <- http.StatusAccepted
 		return
 	}
 	resources, err := template.NewResources(payload, request.Header, trigger.Params, binding)
 	if err != nil {
 		log.Print(err)
+		result <- http.StatusAccepted
 		return
 	}
 	err = createResources(resources, r.RESTClient, r.DiscoveryClient, r.EventListenerNamespace, r.EventListenerName, eventID)
 	if err != nil {
+		result <- http.StatusAccepted
 		log.Print(err)
 	}
+	result <- http.StatusCreated
 }
 
 func (r Resource) processEvent(interceptorURL *url.URL, request *http.Request, payload []byte, headerParams []pipelinev1.Param) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


pkg/sink/sink.go
```diff
@@ -74,26 +74,32 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
 
 	// Execute each Trigger
 	for _, trigger := range el.Spec.Triggers {
-		go r.executeTrigger(event, request, trigger, eventID)
+		resources, err := r.prepareExecuteTrigger(event, request, trigger, eventID)
+		if err != nil {
+			log.Print(err)
+			response.WriteHeader(http.StatusInternalServerError)
+			return
```
ncskier: I could be misinterpreting this, but I think that there will be an error if we return here. Consider the case where I have 2 triggers in my EventListener, `repo1` and `repo2`, with an interceptor for validation. When the EventListener receives an event from `repo2`, the first trigger's interceptor will fail because the request from `repo1` has a different secret than `repo2`. Then, this code will return with an error code. This behavior is faulty because the `repo2` will never be executed.

pkg/sink/sink.go
```diff
@@ -72,27 +72,39 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
...
+		go r.executeTrigger(event, request, trigger, eventID, result)
+	}
+	success := false
+	for i := 0; i < len(el.Spec.Triggers); i++ {
+		if <-result {
+			success = true
+		}
+	}
+	if !success {
+		response.WriteHeader(http.StatusInternalServerError)
```
dibyom: Are we sure we always want to return a 5xx if we fail to process any triggers? 
e.g. what if it is a event that the event listener is not interest in (e.g. push to a dev branch or something).
5xx usually implies that the server is having issues and retrying later should work which is not what we'd want here.

dibyom:
Going through Github's best practices, two of them sound relevant
* We need to respond within 10s: https://developer.github.com/v3/guides/best-practices-for-integrators/#favor-asynchronous-work-over-synchronous

* https://developer.github.com/v3/guides/best-practices-for-integrators/#use-appropriate-http-status-codes-when-responding-to-github
> codes like 201 or 202 to acknowledge receipt of payload that won't be processed (for example, a payload delivered by a branch that's not the default). Reserve the 500 error code for catastrophic failures.

pkg/sink/sink.go
akihikokuroda: Thanks for the reference.  I'll make changes to return 201 for the good case and 202 for no resources are created. 

About the 10 seconds, it should not take such long time to process the triggers.  We can document to reduce the number of triggers if it happens.

pkg/sink/sink.go
```diff
@@ -72,27 +72,39 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
 	log.Printf("EventListener: %s in Namespace: %s handling event (EventID: %s) with payload: %s and header: %v",
 		r.EventListenerName, r.EventListenerNamespace, eventID, string(event), request.Header)
 
+	result := make(chan bool, 10)
 	// Execute each Trigger
 	for _, trigger := range el.Spec.Triggers {
-		go r.executeTrigger(event, request, trigger, eventID)
+		go r.executeTrigger(event, request, trigger, eventID, result)
```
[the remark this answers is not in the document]
akihikokuroda: It needs to get the status code from each thread.   The thread safe variable may be necessary if we use the sync.WaitGroup.  I want to keep the way.

pkg/sink/sink.go
```diff
@@ -72,27 +72,39 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
 	log.Printf("EventListener: %s in Namespace: %s handling event (EventID: %s) with payload: %s and header: %v",
 		r.EventListenerName, r.EventListenerNamespace, eventID, string(event), request.Header)
 
+	result := make(chan int, 10)
```
dibyom: Why is the buffer size 10?

pkg/sink/sink.go
dibyom: Sounds good to me. My main concern is around the interceptor http requests hanging though I think we can fix that by setting low request timeout values on the http client we use.

pkg/sink/sink.go
akihikokuroda: There is no reason :-).  Any number should be OK.  It should work fine with 1 but but it's a little nice not to make some thread wait for the queue.

pkg/sink/sink.go
```diff
@@ -72,27 +72,39 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
...
-		go r.executeTrigger(event, request, trigger, eventID)
+		go r.executeTrigger(event, request, trigger, eventID, result)
+	}
+
+	code := http.StatusAccepted
+	for i := 0; i < len(el.Spec.Triggers); i++ {
+		thiscode := <-result
+		if thiscode < code {
+			code = thiscode
+		}
```
ncskier: Please add a comment before this logic explaining the logic behind what return code is used (essentially repeat what you explained in the PR comment).

pkg/sink/sink.go
```diff
@@ -103,17 +115,22 @@ func (r Resource) executeTrigger(payload []byte, request *http.Request, trigger
...
+		result <- http.StatusAccepted
 		return
 	}
 	err = createResources(resources, r.RESTClient, r.DiscoveryClient, r.EventListenerNamespace, r.EventListenerName, eventID)
 	if err != nil {
+		result <- http.StatusAccepted
 		log.Print(err)
 	}
+	result <- http.StatusCreated
+
```
ncskier: nit: please remove this extra line

pkg/sink/sink.go
ncskier: nit: If we don't want those goroutines to wait, shouldn't the buffer be `len(el.Spec.Triggers)`?

Also, I don't think it really matters if the `executeTrigger()` goroutines are blocked by the channel, because they all return/exit right after sending the message to the channel. That said, I guess it can only help to have the buffer.

pkg/sink/sink.go
```diff
@@ -72,27 +72,39 @@ func (r Resource) HandleEvent(response http.ResponseWriter, request *http.Reques
...
+	code := http.StatusAccepted
+	for i := 0; i < len(el.Spec.Triggers); i++ {
+		thiscode := <-result
+		if thiscode < code {
+			code = thiscode
+		}
 	}
 
 	// TODO: Do we really need to return the entire body back???
+	response.WriteHeader(code)
```
ncskier: Please add a test in [`sink_test.go`](https://github.com/tektoncd/triggers/blob/master/pkg/sink/sink_test.go#L489) to check the `statusCode` of the response.

I think it'll just be a quick check of [this](https://github.com/tektoncd/triggers/blob/master/pkg/sink/sink_test.go#L489) `resp.statusCode`.

pkg/sink/sink.go
akihikokuroda: fixed.

pkg/sink/sink.go
akihikokuroda: I added the comment.

pkg/sink/sink.go
akihikokuroda: I added the check.

PR #211, opened 2019-11-08
Rename sink.Resource to sink.Sink

dibyom:
# Changes
This is a cleanup PR!

Resource is too generic and Sink seems like a more apt name.
Also, fixes the indenting in a bunch of tests to be more consistent
with they style from pipelines.

### DIFF
--- a/cmd/eventlistenersink/main.go
+++ b/cmd/eventlistenersink/main.go
@@ -70,8 +70,8 @@ func main() {
 		logger.Fatal(err)
 	}
 
-	// Create sink Resource
-	r := sink.Resource{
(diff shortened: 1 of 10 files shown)


vtereso:
Did it quick look and seems fine. However, I believe `sink.Sink` would be a stutter as mentioned here https://blog.golang.org/package-names? We do this in a few places, but if we turned on harsher linting it could be an issue.

dibyom:
Interesting point! Did not really consider that but the `pkg.Pkg` naming pattern seems pretty common across core golang e.g. `time.Time`, `context.Context`, `list.List` etc.

Issue #207, opened 2019-11-05
FR: Suport core/built-in event interceptors


dibyom:
We should add a set of "core" built-in interceptors e.g. for for validating Github/Gitlab payloads. This will simplify things for the user as they do not have to deploy their own services to use them. 

Previously mentioned in the [interceptors design doc](https://docs.google.com/document/d/1sw6R-DHmGzfHpD6itMr1AgmNdsjkFdMIUZhv-L6OQLw/edit#heading=h.z3hraejv7bmg) 

In terms of syntax, we could either allow a `type` and set of key value pairs:
```
type EventInterceptor struct {
	Header []pipelinev1.Param json:"header,omitempty"
	ObjectRef *corev1.ObjectReference json:"objectRef,omitempty"
	InterceptorType string
	InterceptorParams map[string]string
}
```

Or we could have a more "typed" approach:

```
interceptor:
  githubValidation:
    repoName: "foo/bar"
    secretToken: ""  
  ```

cc @dlorenc

**Comment by dlorenc** [comment]
Nice!

**Comment by dlorenc** [comment]
The proposed schema is:

```yaml
interceptor:
  webhook:
    ...       
  github:
    secretRef:
      secretName: foo
      secretKey: bar
  gitlab:
    ...
```

The existing `objectRef` and `header` fields will move under `webhook`, and then other core interceptors will appear at the top level next to `webhook`.

bigkevmcd:
I'd be interested in taking on at least the GitHub interceptor (and probably the GitLab one too) once the refactoring lands...

dlorenc:
Ah nice! Sorry I missed this and sort of jumped the gun on getting the github one started. PR coming in a minute!

dlorenc:
https://github.com/tektoncd/triggers/pull/221

dibyom:
Closing this since we now have core interceptors both GitHub and GitLab. Thanks @dlorenc

PR #216, opened 2019-11-09
Core interceptor refactor.

dlorenc:
# Changes

This starts a refactor to support "core" event interceptors. This moves
the existing "ObjectRef" style interceptor to a new package called "webhook",
and embeds it as a "WebhookInterceptor" on the EventBinding type.

Other interceptors (Github, Gitlab, etc.) will be added alongside this. This
PR also extracts the beginnings of the "Interceptor" interface that these other
interceptors will implement.

Ref #207 

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```


vtereso:
An EL trigger can specify an event interceptor, which is usually a webhook event, but does not have to be. Do you think it makes sense to use a different name?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -77,6 +77,12 @@ type EventInterceptor struct {
 	// interceptor request headers. This allows the interceptor to make
 	// decisions specific to an EventListenerTrigger.
 	Header []pipelinev1.Param `json:"header,omitempty"`
```
dibyom: Does the Header field only apply to Webhook interceptors? If so, probably should be within the `WebhookInterceptor` struct?

vtereso:
I had some thoughts more thoughts regarding the interceptor schema. IIUC, the core interceptors for GitHub, GitLab, etc. would be simple validator interceptors for their respective webhooks. In this way, maybe we should explicitly separate the interceptors? For example, 
```
interceptors:
    validation:
    mutator:
```
where validators would likely only be the core interceptors since we would want to provide easy support out of the box? I don't think this is necessary and we really wouldn't have a way to validate an interceptor is a validation/mutating interceptor, but just thought I'd mention.

dlorenc:
> I had some thoughts more thoughts regarding the interceptor schema. IIUC, the core interceptors for GitHub, GitLab, etc. would be simple validator interceptors for their respective webhooks. In this way, maybe we should explicitly separate the interceptors? For example,

Great point! I think actually @wlynch had some ideas in mind where the github interceptor would also modify the incoming request. Use cases included adding a bit on whether or not the sender of the PR is a collaborator on the repo, which is a separate set of github API calls.

pkg/interceptors/interceptors.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"net/http"
+
+	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
+)
+
+// InterceptorInterface is the interface that all interceptors implement.
+type InterceptorInterface interface {
+	ExecuteTrigger([]byte, *http.Request, triggersv1.EventListenerTrigger, string) ([]byte, error)
```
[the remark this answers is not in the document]
dlorenc: D'oh, added names - hopefully they help.

pkg/apis/triggers/v1alpha1/event_listener_types.go
dlorenc: Yeah I think so. I misunderstood what it did earlier.

pkg/interceptors/interceptors.go
```diff
@@ -0,0 +1,28 @@
...
+package interceptors
+
+import (
+	"net/http"
+
+	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
+)
+
+// InterceptorInterface is the interface that all interceptors implement.
+type InterceptorInterface interface {
```
dibyom: Could this just be Interface instead of InterceptorInterface?

bigkevmcd: Or how about just `Interceptor`?

dlorenc: Done!

PR #221, opened 2019-11-13
Add an initial Github interceptor.

dlorenc:
# Changes

Right now this only supports validating a github webhook based on the X-Hub-Signature
header, with a shared secret stored in a Kubernetes secret.

# Submitter Checklist

review them:

# Release Notes

```
A new `Github` interceptor was created, which supports validating external webhooks from github.
```


pkg/interceptors/github/github.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return secret.Data[gh.SecretRef.SecretKey], nil
+}
+
+func (w *Interceptor) ExecuteTrigger(payload []byte, request *http.Request, _ *triggersv1.EventListenerTrigger, _ string) ([]byte, error) {
+	// No secret set, just continue
+	if w.Github.SecretRef == nil {
```
bigkevmcd: If the `GitHub` interceptor has been configured, shouldn't it require a `SecretRef`?

https://github.com/tektoncd/triggers/pull/221/files#diff-559661e0e83bb92fa19d37cc32078347R143

pkg/interceptors/github/github.go
```diff
@@ -0,0 +1,61 @@
...
+	KubeClientSet kubernetes.Interface
+	Logger        *zap.SugaredLogger
+	Github        *triggersv1.GithubInterceptor
+}
+
+func NewInterceptor(ei *triggersv1.EventInterceptor, k kubernetes.Interface, l *zap.SugaredLogger) interceptors.Interceptor {
+	return &Interceptor{
+		Logger:        l,
+		Github:        ei.Github,
+		KubeClientSet: k,
```
bigkevmcd: Maybe this should accept `GithubInterceptor` rather than `EventInterceptor`, to reduce the coupling, what do you think?

(Same for the WebhookInterceptor)

pkg/apis/triggers/v1alpha1/event_listener_validation.go
```diff
@@ -81,44 +81,68 @@ func (s *EventListenerSpec) validate(ctx context.Context, el *EventListener) *ap
...
-	// Optional explicit match
-	if len(w.ObjectRef.APIVersion) != 0 {
-		if w.ObjectRef.APIVersion != "v1" {
-			return apis.ErrInvalidValue(fmt.Errorf("Invalid apiVersion"), "interceptor.webhook.objectRef.apiVersion")
-		}
+
+	// Enforce oneof
+	numSet := 0
+	if i.Webhook != nil {
+		numSet++
```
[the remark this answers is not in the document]
dlorenc: The idea was to match the style of other k8s objects. The canonical example is something like:
https://godoc.org/k8s.io/api/core/v1#VolumeSource

where a Volume can come from a plethora of sources, but only one can be specified at once.

pkg/interceptors/github/github.go
dlorenc: Ah, done.

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -87,6 +88,18 @@ type WebhookInterceptor struct {
 	Header []pipelinev1.Param `json:"header,omitempty"`
 }
 
+// GithubInterceptor provides a webhook to intercept and pre-process events
+type GithubInterceptor struct {
+	SecretRef *SecretRef `json:"secretRef,omitempty"`
+}
+
+// SecretRef contains the information required to reference a single secret string
```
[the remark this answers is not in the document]
dlorenc: Done. That wasn't the actual reason, so clarified!

pkg/interceptors/github/github.go
dlorenc: Ah yeah, I think I was future proofing it a bit since there are other fields I want to add here.

pkg/apis/triggers/v1alpha1/event_listener_validation.go
```diff
@@ -81,44 +81,68 @@ func (s *EventListenerSpec) validate(ctx context.Context, el *EventListener) *ap
...
+			} else if len(header.Value.ArrayVal) == 0 {
 				return apis.ErrInvalidValue(fmt.Errorf("Invalid header value"), fmt.Sprintf("interceptor.webhook.header[%d].value", i))
 			}
-		} else if len(header.Value.ArrayVal) == 0 {
-			return apis.ErrInvalidValue(fmt.Errorf("Invalid header value"), fmt.Sprintf("interceptor.webhook.header[%d].value", i))
 		}
 	}
+
+	// No github validation required yet.
+	// if i.Github != nil {
```
dibyom: Should we validate that the SecretRef is set?

pkg/interceptors/github/github.go
```diff
@@ -0,0 +1,64 @@
...
+		Logger:        l,
+		Github:        gh,
+		KubeClientSet: k,
+	}
+}
+
+func getSecretToken(cs kubernetes.Interface, gh *triggersv1.GithubInterceptor) ([]byte, error) {
+	ns := gh.SecretRef.Namespace
+	if ns == "" {
+		ns = metav1.NamespaceDefault
```
dibyom: Would it make sense to default to the namespace that the event listener is on? We do this in other places: https://github.com/tektoncd/triggers/blob/fccb70830e5836bce882513e357bafa283cb6baa/pkg/sink/sink.go#L154

pkg/apis/triggers/v1alpha1/event_listener_validation.go
dlorenc: Purposefully leaving this out, since Github will have other fields in the future. We could add it for now then remove it when the next field is added if you prefer, though.

pkg/interceptors/github/github.go
dlorenc: Done, thanks!

docs/eventlisteners.md
```diff
@@ -45,8 +45,18 @@ to configure the underlying `Service` resource to make it reachable externally.
...
 ## Event Interceptors
 
-Triggers within an `EventListener` can optionally specify an interceptor field
-which contains an [`ObjectReference`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#objectreference-v1-core) to a Kubernetes Service. If an interceptor
+Triggers within an `EventListener` can optionally specify an interceptor field.
+
+Event Interceptors can take several different forms today:
+
+* Webhook Interceptors
+* Github Interceptors
```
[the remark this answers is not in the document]
dlorenc: Thanks!

PR #225, opened 2019-11-15
Simplify merging params in Bindings

dibyom:
# Changes

This commit also simplifies some of the tests in resource_test
as well as align the indenting in event_test.

# Submitter Checklist

review them:

### DIFF
--- a/pkg/template/event_test.go
+++ b/pkg/template/event_test.go
@@ -35,7 +35,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 )
 
-func Test_BodyPathVarRegex(t *testing.T) {
+func TestBodyPathVarRegex(t *testing.T) {
 	tests := []string{
 		"$(body)",
 		"$(body.a-b)",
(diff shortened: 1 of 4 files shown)


pkg/template/event_test.go
```diff
@@ -35,7 +35,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 )
```
dibyom: All changes in this file are just indenting changes

pkg/template/resource.go
```diff
@@ -106,30 +106,26 @@ func ApplyUIDToResourceTemplate(rt json.RawMessage, uid string) json.RawMessage
...
 	}
 	return params
 }
+
+func MergeBindingParams(bindings []*triggersv1.TriggerBinding) ([]pipelinev1.Param, error) {
+	params := []pipelinev1.Param{}
+	for _, b := range bindings {
+		params = append(params, b.Spec.Params...)
+	}
+	seen := make(map[string]struct{}, len(params))
```
wlynch: This might be simpler as map[string]bool. It avoids the need for the emptyVal struct, and you can simplify the existence check to:

```go
if seen[p.Name] {
	return nil, fmt.Errorf("duplicate param name: %s", p.Name)
}
```

dibyom: Yeah, struct{} is better for memory but that's premature optimization at this point 😄

wlynch: :astonished: I didn't know this! Thanks for teaching me something new! ^_^

Issue #178, opened 2019-10-16
Support JSONPath syntax for variable extraction in TriggerBindings


ncskier:
# Expected Behavior
A user should _intuitively_ be able to specify a [`body`](https://github.com/tektoncd/triggers/blob/master/docs/triggerbindings.md#body) parameter to point to a piece of data in their HTTP request JSON body. 

# Actual Behavior
Currently, the Triggers project is using the [gjson](https://github.com/tidwall/gjson) library because it makes it easy for us to search the JSON for a specified path. However, gjson does not use the conventional JSONPath syntax, instead it has its own [GJSON path syntax](https://github.com/tidwall/gjson/blob/master/SYNTAX.md). Since the gjson path syntax is different from the JSONPath syntax, it can be unintuitive, and Triggers does not have any documentation linking to an explanation of this syntax.

Furthermore, the [`bodyPathVarRegex`](https://github.com/tektoncd/triggers/blob/master/pkg/template/event.go#L30) limits the user to a subset of the gjson syntax.

# Additional Info
This concern originated in our Slack channel [here](https://tektoncd.slack.com/archives/CKUSJ2A5D/p1571198034014600).

### I hope that the following work will come from this issue:
Decide whether to keep the gjson path syntax, or switch to JSONPath
1. If we decide to keep the gjson library and the gjson path syntax
- add Triggers documentation linking to the gjson path syntax page
- update the `bodyPathVarRegex` to accept all valid gjson path syntax
- update/add tests for examples using gjson path syntax
2. If we decide to use JSONPath syntax
- find a library that we can substitute for gjson (or use with gjson?)
- update code & tests to use the JSONPath syntax
- add Triggers documentation linking to an explanation of the JSONPath syntax

dibyom:
JSONPath definitely seems more popular and is definitely used in other parts of k8s (e.g. kubectl supports it)

kav:
My inclination as a user would be to match standard json syntax  _unless_
1. The expressiveness of the querying capabilities is an important feature here and throughout Tekton (or even just Triggers)
1. gjson is standard elsewhere in the Kubernetes or even Tekton ecosystem 

Those are true perhaps updating the regex to catch `[` and `]` and report to the user that they might want to read about gjson.

[...]

PR #241, opened 2019-11-27
JSONPath Support for TriggerBindings

dibyom:
# Changes
There are two commits in this PR. The first commit adds functions for JSONPath and the second one uses those functions to extract values for EventBinding.

Commits:
This commit adds functions to support JSONPath
in `TriggerBindings` using the `k8s.io/client-go/util/jsonpath`
library.

There are a few deviations from the default behavior of the library:

1. The JSONPath expressions have to be wrapped in the Tekton variable
interpolation syntax i.e `$()` : `$(body)`.

2.  We  support the `RelaxedJSONPathExpresion` syntax used in
`kubectl get -o custom-columns`. This means that the curly braces `{}`
and the leading dot `.` can be omitted i.e we support both `$(body.key)`
as well as `$({.body.key})

3. We return valid JSON values when the value is a JSON array or
map. By default, the library returns a string containing an internal
go representation. So, if the JSONPath expression selects `{"a": "b"}`,
the library will return `map[a: b]` while we return the valid JSON
string i.e `{"a":"b"}`

This commit switches TriggerBinding params to use
JSONPath instead of GJSON.

Part of #178 

# Submitter Checklist

review them:

# Release Notes

```
TriggerBindings now support JSONPath expressions instead of GJSON. Check docs/triggerbindings.md for the syntax and examples

```

### DIFF
--- a/docs/triggerbindings.md
+++ b/docs/triggerbindings.md
@@ -33,46 +33,40 @@ Each parameter has a `name` and a `value`.
 
 ## Event Variable Interpolation
 
-In order to parse generic events as efficiently as possible,
-[GJSON](https://github.com/tidwall/gjson) is used internally. As a result, the
-binding [path syntax](https://github.com/tidwall/gjson#path-syntax) differs
-slightly from standard JSON. As of now, the following patterns are supported
-within `TriggerBinding` parameter value interpolation: -
-`\$\(body(\.[[:alnum:]/_\-\.\\]+|\.#\([[:alnum:]=<>%!"\*_-]+\)#??)*\)` -
-`\$\(header(\.[[:alnum:]_\-]+)?\)`
+TriggerBindings can access values from the HTTP JSON body and the headers using
+JSONPath expressions wrapped in `$()`.
 
-### Body
+These are all valid expressions:
+```shell script
+$(body.key1)
+$(.body.key)
+```
 
-HTTP Post request body data can be referenced using variable interpolation. Text
-in the form of `$(body.X.Y.Z)` is replaced by the body data at JSON path
-`X.Y.Z`.
+These are invalid expressions:
+```shell script
+.body.key1 # INVALID - Not wrapped in $()
+$({body) # INVALID - Ending curly brace absent
+```
+
+### Examples
+
+``` shell script
 
 `$(body)` is replaced by the entire body.
 
-The following are some example variable interpolation replacements:
-``` $(body)
--> "{\"key1\": \"value1\", \"key2\": {\"key3\": \"value3\"}, \"key4\":
-[\"value4\", \"value5\"]}"
+$(body) -> "{"key1": "value1", "key2": {"key3": "value3"}, "key4": ["value4", "value5"]}"
 
 $(body.key1) -> "value1"
 
-$(body.key2) -> "{\"key3\": \"value3\"}"
+$(body.key2) -> "{"key3": "value3"}"
 
 $(body.key2.key3) -> "value3"
 
-$(body.key4.0) -> "value4"
-```
+$(body.key4[0]) -> "value4"
 
-### Header
+# $(header) is replaced by all of the headers from the event.
 
-HTTP Post request header data can be referenced using variable interpolation.
-Text in the form of `$(header.X)` is replaced by the event's header named `X`.
-
-`$(header)` is replaced by all of the headers from the event.
-
-The following are some example variable interpolation replacements:
-```
-$(header) -> "{\"One\":[\"one\"], \"Two\":[\"one\",\"two\",\"three\"]}"
+$(header) -> "{"One":["one"], "Two":["one","two","three"]}"
 
 $(header.One) -> "one"
 
--- a/pkg/template/event.go
+++ b/pkg/template/event.go
@@ -18,177 +18,82 @@ package template
 
 import (
 	"encoding/json"
-	"regexp"
+	"fmt"
+	"net/http"
 	"strings"
 
 	pipelinev1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
-	"github.com/tidwall/gjson"
-	"golang.org/x/xerrors"
 )
-
-var (
-	// bodyPathVarRegex determines valid body path variables
-	// The body regular expression allows for a subset of GJSON syntax, the mininum
-	// required to navigate through dictionaries, query arrays and support
-	// namespaced label names e.g. tekton.dev/eventlistener
-	bodyPathVarRegex = regexp.MustCompile(`\$\(body(\.[[:alnum:]/_\-\.\\]+|\.#\([[:alnum:]=<>%!"\*_-]+\)#??)*\)`)
-
-	// The headers regular expression allows for simple navigation down a hierarchy
-	// of dictionaries
-	headerVarRegex = regexp.MustCompile(`\$\(header(\.[[:alnum:]_\-]+)?\)`)
-)
-
-// getBodyPathFromVar returns the body path given an body path variable
-// $(body.my.path) -> my.path
-// $(body) returns an empty string "" because there is no body path
-func getBodyPathFromVar(bodyPathVar string) string {
-	// Assume bodyPathVar matches the bodyPathVarRegex
-	if bodyPathVar == "$(body)" {
-		return ""
-	}
-	return strings.TrimSuffix(strings.TrimPrefix(bodyPathVar, "$(body."), ")")
-}
-
-// getHeaderFromVar returns the header given a header variable
-// $(header.example) -> example
-func getHeaderFromVar(headerVar string) string {
-	// Assume headerVar matches the headerVarRegex
-	if headerVar == "$(header)" {
-		return ""
-	}
-
-	return strings.TrimSuffix(strings.TrimPrefix(headerVar, "$(header."), ")")
-}
-
-// ApplyBodyToParams returns the params with each body path variable replaced
-// with the appropriate data from the body. Returns an error when the body
-// path variable is not found in the body.
-func ApplyBodyToParams(body []byte, params []pipelinev1.Param) ([]pipelinev1.Param, error) {
-	for i := range params {
-		param, err := applyBodyToParam(body, params[i])
-		if err != nil {
-			return nil, err
-		}
-		params[i] = param
-	}
-	return params, nil
-}
-
-// applyBodyToParam returns the param with each body path variable replaced
-// with the appropriate data from the body. Returns an error when the body
-// path variable is not found in the body.
-func applyBodyToParam(body []byte, param pipelinev1.Param) (pipelinev1.Param, error) {
-	// Get each body path variable in the param
-	bodyPathVars := bodyPathVarRegex.FindAllString(param.Value.StringVal, -1)
-	for _, bodyPathVar := range bodyPathVars {
-		bodyPath := getBodyPathFromVar(bodyPathVar)
-		bodyPathValue, err := getBodyPathValue(body, bodyPath)
-		if err != nil {
-			return param, err
-		}
-		param.Value.StringVal = strings.Replace(param.Value.StringVal, bodyPathVar, bodyPathValue, -1)
-	}
-	return param, nil
-}
-
-// getBodyPathValue returns the value of the bodyPath in the body. An error
-// is returned if the bodyPath is not found in the body.
-func getBodyPathValue(body []byte, bodyPath string) (string, error) {
-	var bodyPathValue string
-	if bodyPath == "" {
-		// $(body) has an empty bodyPath, so use the entire body as the bodyValue
-		bodyPathValue = string(body)
-	} else {
-		bodyPathResult := gjson.GetBytes(body, bodyPath)
-		if bodyPathResult.Index == 0 {
-			return "", xerrors.Errorf("Error body path %s not found in the body %s", bodyPath, string(body))
-		}
-		bodyPathValue = bodyPathResult.String()
-		if bodyPathResult.Type == gjson.Null {
-			bodyPathValue = "null"
-		}
-	}
-	return strings.Replace(bodyPathValue, `"`, `\"`, -1), nil
-}
-
-// ApplyHeaderToParams returns the params with each header variable replaced
-// with the appropriate header value. Returns an error when the header variable
-// is not found.
-func ApplyHeaderToParams(header map[string][]string, params []pipelinev1.Param) ([]pipelinev1.Param, error) {
-	for i := range params {
-		param, err := applyHeaderToParam(header, params[i])
-		if err != nil {
-			return nil, err
-		}
-		params[i] = param
-	}
-	return params, nil
-}
-
-// applyHeaderToParam returns the param with each header variable replaced
-// with the appropriate header value. Returns an error when the header variable
-// is not found.
-func applyHeaderToParam(header map[string][]string, param pipelinev1.Param) (pipelinev1.Param, error) {
-	// Get each header variable in the param
-	headerVars := headerVarRegex.FindAllString(param.Value.StringVal, -1)
-	for _, headerVar := range headerVars {
-		headerName := getHeaderFromVar(headerVar)
-		headerValue, err := getHeaderValue(header, headerName)
-		if err != nil {
-			return param, err
-		}
-		param.Value.StringVal = strings.Replace(param.Value.StringVal, headerVar, headerValue, -1)
-	}
-	return param, nil
-}
-
-// getHeaderValue returns a string representation of the headerName in the event
-// header. An error is returned if the headerName is not found in the header.
-func getHeaderValue(header map[string][]string, headerName string) (string, error) {
-	var headerValue string
-	if headerName == "" {
-		// $(header) has an empty headerName, so use all the headers in the headerValue
-		b, err := json.Marshal(&header)
-		if err != nil {
-			return "", xerrors.Errorf("Error marshalling header %s: %s", header, err)
-		}
-		headerValue = string(b)
-	} else {
-		value, ok := header[headerName]
-		if !ok {
-			return "", xerrors.Errorf("Error headerName %s not found in the event header %s", headerName, header)
-		}
-		headerValue = strings.Join(value, " ")
-	}
-	return strings.Replace(headerValue, `"`, `\"`, -1), nil
-}
 
 // ResolveParams takes a given trigger binding and produces the resulting
 // resource params.
-func ResolveParams(bindings []*triggersv1.TriggerBinding, body []byte, header map[string][]string, params []pipelinev1.ParamSpec) ([]pipelinev1.Param, error) {
+func ResolveParams(bindings []*triggersv1.TriggerBinding, body []byte, header http.Header, params []pipelinev1.ParamSpec) ([]pipelinev1.Param, error) {
 	out, err := MergeBindingParams(bindings)
 	if err != nil {
-		return nil, xerrors.Errorf("error merging trigger params: %v", err)
+		return nil, fmt.Errorf("error merging trigger params: %w", err)
 	}
-	out, err = ApplyBodyToParams(body, out)
+	event, err := newEvent(body, header)
 	if err != nil {
-		return nil, xerrors.Errorf("error applying body to trigger params: %s", err)
+		return nil, fmt.Errorf("failed to create Event: %w", err)
 	}
-	out, err = ApplyHeaderToParams(header, out)
+	out, err = applyEventValuesToParams(out, event)
 	if err != nil {
-		return nil, xerrors.Errorf("error applying header to trigger params: %s", err)
+		return nil, fmt.Errorf("failed to ApplyEventValuesToParams: %w", err)
 	}
-
 	return MergeInDefaultParams(out, params), nil
 }
 
+// ResolveResources resolves a templated resource by replacing params with their values.
 func ResolveResources(template *triggersv1.TriggerTemplate, params []pipelinev1.Param) []json.RawMessage {
 	resources := make([]json.RawMessage, len(template.Spec.ResourceTemplates))
-	uid := UID()
 	for i := range template.Spec.ResourceTemplates {
 		resources[i] = ApplyParamsToResourceTemplate(params, template.Spec.ResourceTemplates[i].RawMessage)
-		resources[i] = ApplyUIDToResourceTemplate(resources[i], uid)
+		resources[i] = ApplyUIDToResourceTemplate(resources[i], UID())
 	}
 	return resources
+}
+
+// event represents a HTTP event that Triggers processes
+type event struct {
+	Header map[string]string `json:"header"`
+	Body   interface{}       `json:"body"`
+}
+
+// newEvent returns a new Event from HTTP headers and body
+func newEvent(body []byte, headers http.Header) (*event, error) {
+	var data interface{}
+	if len(body) > 0 {
+		if err := json.Unmarshal(body, &data); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal request body: %w", err)
+		}
+	}
+	joinedHeaders := make(map[string]string, len(headers))
+	for k, v := range headers {
+		joinedHeaders[k] = strings.Join(v, ",")
+	}
+
+	return &event{
+		Header: joinedHeaders,
+		Body:   data,
+	}, nil
+}
+
+// applyEventValuesToParams returns a slice of Params with the JSONPath variables replaced
+// with values from the event body and headers.
+func applyEventValuesToParams(params []pipelinev1.Param, ec *event) ([]pipelinev1.Param, error) {
+	for idx, p := range params {
+		pValue := p.Value.StringVal
+		// Find all expressions wrapped in $() from the value
+		expressions := tektonVar.FindAllString(pValue, -1)
+		for _, expr := range expressions {
+			val, err := ParseJSONPath(ec, expr)
+			if err != nil {
+				return nil, fmt.Errorf("failed to replace JSONPath value for param %s: %s: %w", p.Name, p.Value, err)
+			}
+			pValue = strings.ReplaceAll(pValue, expr, val)
+		}
+		params[idx].Value = pipelinev1.ArrayOrString{Type: pipelinev1.ParamTypeString, StringVal: pValue}
+	}
+	return params, nil
 }
--- a/pkg/template/event_test.go
+++ b/pkg/template/event_test.go
@@ -18,7 +18,6 @@ package template
 
 import (
 	"encoding/json"
-	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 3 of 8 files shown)


pkg/template/jsonpath.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+
+// TektonJSONPathExpression returns a valid JSONPath expression. It accepts
+// a "RelaxedJSONPath" expression that is wrapped in the Tekton variable
+// interpolation syntax i.e. $(). RelaxedJSONPath expressions can optionally
+// omit the leading curly braces '{}' and '.'
+func TektonJSONPathExpression(expr string) (string, error) {
+	if !isTektonExpr(expr) {
+		return "", errors.New("Expression not wrapped in $()")
```
wlynch: Won't this error for cases of simple strings? We'll still want this for environment parameters. (not sure if we want to add this logic here, or above in ParseJSONPath)

pkg/template/event.go
```diff
@@ -169,26 +169,65 @@ func getHeaderValue(header map[string][]string, headerName string) (string, erro
...
+	return Event{
+		Header: joinedHeaders,
+		Body:   data,
+	}, nil
+}
+
+
+// ApplyEventValuesToParams returns a slice of Params with the JSONPath variables replaced
+// with values from the event body and headers.
+func ApplyEventValuesToParams(ec Event, params []pipelinev1.Param) ([]pipelinev1.Param, error) {
```
wlynch: super nit: my preference is for this to be (params []pipelinev1.Param, ec Event), since the params are the primary object we are acting on here.

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,155 @@
...
+	"k8s.io/client-go/third_party/forked/golang/template"
+	"k8s.io/client-go/util/jsonpath"
+)
+
+var (
+	tektonVar  = regexp.MustCompile(`^\$\((.*)\)$`)
+	jsonRegexp = regexp.MustCompile(`^\{\.?([^{}]+)\}$|^\.?([^{}]+)$`)
+)
+
+func ParseJSONPath(input interface{}, expr string) (string, error) {
```
wlynch: Add godoc comments to describe the inputs/outputs

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	fullResults, err := j.FindResults(input)
+	if err != nil { //
+		return "", err
+	}
+
+	// Copy of j.PrintResults
+	for ix := range fullResults {
+		if err := printResults(buf, fullResults[ix]); err != nil {
```
wlynch: Can you provide guidance on how to interpret fullResults (`[][]reflect.Value`) vs results (`[]reflect.Value`)? Is this a distinction of line vs field?

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,155 @@
...
+func printResults(wr io.Writer, results []reflect.Value) error {
+	for i, r := range results {
+		text, err := textValue(r)
+		if err != nil {
+			return err
+		}
+		if i != len(results)-1 {
+			text = append(text, ' ')
+		}
+		if _, err = wr.Write(text); err != nil {
```
wlynch: nit: `if _, err := ...`

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	switch t.Kind() {
+	// evalToText() returns <map> ....; return JSON string instead.
+	case reflect.Map, reflect.Slice:
+		text, err := json.Marshal(v.Interface())
+		if err != nil {
+			return nil, err
+		}
+		return text, nil
```
wlynch: `return json.Marshal(v.Interface())`

pkg/template/jsonpath.go
wlynch: Not sure if this is worth doing or not, but thoughts on doing things this way as opposed to passing the writer through?

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,155 @@
...
+//   * {metadata.name} (no leading '.')
+//   * .metadata.name (no curly braces '{...}')
+//   * {.metadata.name} (complete expression)
+// And transforms them all into a valid jsonpath expression:
+//   {.metadata.name}
+func relaxedJSONPathExpression(pathExpression string) (string, error) {
+	if len(pathExpression) == 0 {
+		return pathExpression, nil
+	}
+	submatches := jsonRegexp.FindStringSubmatch(pathExpression)
```
wlynch: Mostly because I don't fully grok the regex being used here, but what exactly are we trying to extract here? (might be helpful to provide an example).

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,155 @@
...
+	var fieldSpec string
+	if len(submatches[1]) != 0 {
+		fieldSpec = submatches[1]
+	} else {
+		fieldSpec = submatches[2]
+	}
+	return fmt.Sprintf("{.%s}", fieldSpec), nil
+}
+
+// unwrwapTektonExpr takes an experession of the form $(abc) and returns abc.
```
wlynch: unwrapTektonExpr, expression

pkg/template/jsonpath_test.go
```diff
@@ -0,0 +1,144 @@
...
+var objects = `{"a":"v","c":{"d":"e"},"empty": "","null": null, "number": 42}`
+var arrays = `[{"a": "b"}, {"c": "d"}, {"e": "f"}]`
+// Checks that we print JSON strings when the JSONPath selects
+// an array or map value and regular values otherwise
+func TestParseJSONPath(t *testing.T) {
+	var objectBody = fmt.Sprintf(`{"body":%s}`, objects)
+	tests := []struct {
+		name  string
+		expr string
+		in     string
```
wlynch: Instead of encoding the object body to a string then unmarshalling it out, would it be easier to make this interface{} and set the fields as maps?

pkg/template/jsonpath_test.go
```diff
@@ -0,0 +1,144 @@
...
+	}, {
+		name: "booleans",
+		in: `{"body": {"bool": true}}`,
+		expr: "$(body.bool)",
+		want: "true",
+	}, {
+		name: "null values",
+		in: objectBody,
+		expr: "$(body.null)",
+		want: "null",
```
wlynch: Is this WAI? Even though the payload says `null` we should probably make a distinction between `null` and `"null"`. I say either default to `""` or have ParseJSONPath return `*string` so we can represent this nil value (my guess is `""` is probably sufficient)

pkg/template/event.go
```diff
@@ -169,26 +169,65 @@ func getHeaderValue(header map[string][]string, headerName string) (string, erro
...
+
+
+// Event represents a HTTP event that Triggers processes
+type Event struct {
+	Header map[string]string `json:"header"`
+	Body   interface{}       `json:"body"`
+}
+
+// NewEvent returns a new Event from HTTP headers and body
+func NewEvent(body []byte, headers map[string][]string) (Event, error) {
```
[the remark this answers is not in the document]
dibyom: done

pkg/template/event.go
```diff
@@ -169,26 +169,65 @@ func getHeaderValue(header map[string][]string, headerName string) (string, erro
...
+	var data interface{}
+	if len(body) > 0 {
+		if err := json.Unmarshal(body, &data); err != nil {
+			return Event{}, xerrors.Errorf("failed to unmarshal request body: %q", err)
+		}
+	}
+	joinedHeaders := make(map[string]string, len(headers))
+	for k, v := range headers {
+		joinedHeaders[k] = strings.Join(v, ",")
```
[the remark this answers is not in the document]
dibyom: Ack. While we do use ArrayOrString in Param, we only ever use the string value, so I think its simpler to just use strings as much as possible expect when we have to assign the value to the Param.

pkg/template/event.go
```diff
@@ -169,26 +169,65 @@ func getHeaderValue(header map[string][]string, headerName string) (string, erro
...
 		resources[i] = ApplyParamsToResourceTemplate(params, template.Spec.ResourceTemplates[i].RawMessage)
-		resources[i] = ApplyUIDToResourceTemplate(resources[i], uid)
+		resources[i] = ApplyUIDToResourceTemplate(resources[i], UID())
 	}
 	return resources
 }
+
+
+// Event represents a HTTP event that Triggers processes
+type Event struct {
```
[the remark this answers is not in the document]
dibyom: Not really at the moment. Though it would be nice if Filters could reuse the same struct later...unexported for now

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,155 @@
...
+	if err := j.Parse(expr); err != nil {
+		return "", err
+	}
+
+	fullResults, err := j.FindResults(input)
+	if err != nil { //
+		return "", err
+	}
+
+	// Copy of j.PrintResults
```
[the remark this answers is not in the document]
dibyom: Old comment...don't think its really applicable anymore

pkg/template/event.go
dibyom: done!

pkg/template/jsonpath.go
dibyom: 🤕

pkg/template/jsonpath.go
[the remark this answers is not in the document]
dibyom: There isn't I think besides following the pattern from the original library at https://github.com/tektoncd/triggers/blob/master/vendor/k8s.io/client-go/util/jsonpath/jsonpath.go#L72

Changed!

pkg/template/event.go
```diff
@@ -169,26 +169,65 @@ func getHeaderValue(header map[string][]string, headerName string) (string, erro
 func ResolveParams(bindings []*triggersv1.TriggerBinding, body []byte, header map[string][]string, params []pipelinev1.ParamSpec) ([]pipelinev1.Param, error) {
 	out, err := MergeBindingParams(bindings)
 	if err != nil {
-		return nil, xerrors.Errorf("error merging trigger params: %v", err)
+		return nil, xerrors.Errorf("error merging trigger params: %q", err)
 	}
-	out, err = ApplyBodyToParams(body, out)
```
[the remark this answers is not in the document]
dibyom: Only used in tests. Wanted to remove it in a different PR to keep the size of this small but can also do it here.

pkg/template/jsonpath.go
dibyom: done

pkg/template/jsonpath.go
dibyom: done

pkg/template/jsonpath.go
dibyom: Great catch...why didn't the test fail :thinking:

pkg/template/event.go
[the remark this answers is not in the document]
dibyom: done

pkg/template/jsonpath.go
dibyom: What's the advantage? My thinking was to keep the copied code from the library (i.e. the printResults and the evalToText function) as close to the original as possible, try to upstream the fix in  the original library itself, and then at some point in the future remove the copy pasta!

pkg/template/jsonpath.go
dibyom: Added a check in the calling function. Need to add tests.

pkg/template/jsonpath_test.go
```diff
@@ -0,0 +1,145 @@
...
+	tests := []struct {
+		name string
+		expr string
+		in   string
+		want string
+	}{{
+		name: "objects",
+		in:   objectBody,
+		expr: "$(body)",
+		// TODO: Do we need to escape backslashes for backwards compat?
```
dibyom: @ncskier @vtereso In the current gjson bindings, we always escape the quotes `\"` in the returned value. Why do we do that?

ncskier: I think it's because we're passing we're passing that JSON object as a type String in the ResourceTemplate JSON: https://github.com/tektoncd/triggers/blob/master/pkg/sink/sink.go#L195

pkg/template/event.go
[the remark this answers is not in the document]
wlynch: I think it's fine to remove it here. Simple deletions are easy to grok, and I'd rather have a larger PR than someone accidentally reading / trying to modify a dead piece of code. Also helps in the event of a rollback (only 1 commit to revert instead of 2).

pkg/template/jsonpath.go
wlynch: SGTM

pkg/template/jsonpath.go
dibyom: so I think the only time the length of `[][]reflect.Value` is greater than 1 is when we use `range` keyword. AFAIK the `range` keyword is not standard JSONPath, its something that is only available in this library and used in kubectl

dibyom: https://gist.github.com/so0k/42313dbb3b547a0f51a547bb968696ba#jsonpath

pkg/template/jsonpath.go
dibyom: added tests in Test_NewResources

pkg/template/jsonpath_test.go
dibyom: ok, I'm adding escaping behavior for now for backwards compatibility. I'll open an issue to explore alternative approaches here -- maybe smarter concatenation somehow or using json patches

pkg/template/jsonpath_test.go
dibyom: Yeah, so I went back and forth on this a few times before settling on "null". I agree that if we were talking about return go values, the correct way should be `*string`. However, for our use case here (i.e. concatenating the json values in string in the trigger templates), that would result in the string `<nil>`.  If we use "" to represent the printable value of `null`, we have no way to differentiate between `null` and the empty string "".

The `"null"` and `null` collision is unfortunate though I think this is the least bad way at the moment. Also, this issue is only valid if the JSONPath selects a field whose value is `null`. If the selection is an object or an array which contains both the string null as well as a null value, we get the following: `{"stringNull":"null","nullValue":null}`

pkg/template/jsonpath_test.go
dibyom: yeah, don't have a strong opinion on this -> might be a bit more verbose for the array and objects that contain other embedded objects?

pkg/template/jsonpath.go
dibyom: The function docstring and the substring error has some examples of what the regex accepts but basically it tries to extract a substring - so:
`{blah}` or `{.blah}` or `.blah` or `blah`

The substring however cannot contain curly braces so `bla{h` won't match

pkg/template/jsonpath_test.go
ncskier: Sounds good, thanks!

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,158 @@
...
+	"k8s.io/client-go/util/jsonpath"
+)
+
+var (
+	// tektonVar is a regular expression for strings surrounded by $()
+	tektonVar = regexp.MustCompile(`^\$\((.*)\)$`)
+
+	// jsonRegexp is a regular expression for JSONPath expressions
+	// with or without the enclosing {} and the leading . inside the curly
+	// braces e.g.  'a.b' or '.a.b' or '{a.b}' or '{.a.b}
```
wlynch: nit: '{.a.b}'

dibyom: doh

dibyom:
> LGTM pending discussion of #253, since IIUC this will break bindings that mix static and interpolated values.

Yeah, let's keep backwards compatibility for now. If we decide to go with #253 later it should be simple enough to undo.

pkg/template/jsonpath_test.go
```diff
@@ -0,0 +1,145 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			var data interface{}
+			err := json.Unmarshal([]byte(tt.in), &data)
+			if err != nil {
+				t.Errorf("Could not unmarshall body : %q", err)
+			}
+			got, err := template.ParseJSONPath(data, tt.expr)
+			if err != nil {
+				t.Errorf("ParseJSONPath() error = %v", err)
+				return
```
bobcatfish: although using `Fatalf` in a goroutine is dangerous, it seems like it's fine to use `Fatalf` in a sub test, the golang blog even does it (https://blog.golang.org/subtests) so i think we you could use Fatalf here instead of Errorf + return

pkg/template/jsonpath_test.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+
+			if diff := cmp.Diff(got, strings.Replace(tt.want, " ", "", -1)); diff != "" {
```
bobcatfish: i think `cmp.Diff` output is way easier to read if you pass in "want, got" (e.g. like in this example https://godoc.org/github.com/google/go-cmp/cmp#Diff), then you intuitively get a - besides things you wanted but didnt get, and a + for things you didnt want that you got (the other way around is confusing imo)

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,158 @@
+package template
```
bobcatfish: hm do you think there's another name we could use besides "template"? in pipelines we are trying to avoid using the term "templating" for the variable substitution we are doing b/c "template" implies additional functionality we want to avoid adding

pkg/template/jsonpath_test.go
bobcatfish: (same for other occurrences of this)

pkg/template/jsonpath_test.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+		})
+	}
+}
+
+func TestTektonJSONPathExpression(t *testing.T) {
+	tests := []struct {
+		expr    string
+		want    string
+		wantErr bool
```
bobcatfish: i like how the above tests are separate for the error case and the success case, could we do that here as well?

pkg/template/resource.go
```diff
@@ -92,7 +93,10 @@ func ApplyParamsToResourceTemplate(params []pipelinev1.Param, rt json.RawMessage
 func applyParamToResourceTemplate(param pipelinev1.Param, rt json.RawMessage) json.RawMessage {
 	// Assume the param is valid
 	paramVariable := fmt.Sprintf("$(params.%s)", param.Name)
-	return bytes.Replace(rt, []byte(paramVariable), []byte(param.Value.StringVal), -1)
+	// Escape quotes for backwards compatibility. Needed since we are appending JSON
+	//
```
bobcatfish: extra comment line?

also is this temporary, if so can we have an issue to track removing it? and if it isn't temporary then maybe "backward compatibility" is a bit vague and we can explain a bit more about what we are being backward compatible with and why?

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,158 @@
...
+		fieldSpec = submatches[1]
+	} else {
+		fieldSpec = submatches[2]
+	}
+	return fmt.Sprintf("{.%s}", fieldSpec), nil
+}
+
+// unwrapTektonExpr takes an expression of the form $(abc) and returns abc.
+func unwrapTektonExpr(expr string) string {
+	return tektonVar.ReplaceAllString(expr, "$1")
```
bobcatfish: im not a huge fan of functions that only have one line in them, i feel like they just add a layer of indirection without adding too much (e.g. if i want to know what `unwrapTektonExpr` does i have to come over here and look at this vs looking direclty at `tektonVar.ReplaceAllString`), but maybe there is a benefit i'm missing?

pkg/template/jsonpath.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+	return nil
+}
+
+// textValue translates reflect value to corresponding text
+// If the value if an array or map, it returns a JSON representation
+// of the value (as opposed to the internal go representation of the value)
+// Otherwise, the text value is from the `evalToText` function, originally from
+// k8s.io/client-go/util/jsonpath/jsonpath.go
```
bobcatfish: it seems like there are a few functions in this file that are copied directly from other places - maybe your commit message could explain a bit more about why? (im thinking it's b/c they were unexported and you need to use them?)

pkg/template/jsonpath.go
bobcatfish: i just realized the type is literally called "trigger template" and this package is probably named after that so just ignore me :woman_facepalming:

pkg/template/event.go
```diff
@@ -18,177 +18,79 @@ package template
...
-}
-
 // ResolveParams takes a given trigger binding and produces the resulting
 // resource params.
-func ResolveParams(bindings []*triggersv1.TriggerBinding, body []byte, header map[string][]string, params []pipelinev1.ParamSpec) ([]pipelinev1.Param, error) {
+func ResolveParams(bindings []*triggersv1.TriggerBinding, body []byte, header http.Header, params []pipelinev1.ParamSpec) ([]pipelinev1.Param, error) {
 	out, err := MergeBindingParams(bindings)
 	if err != nil {
-		return nil, xerrors.Errorf("error merging trigger params: %v", err)
+		return nil, xerrors.Errorf("error merging trigger params: %q", err)
```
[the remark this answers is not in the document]
dibyom: 1. What about %w which calls `error.UnWrap` for errors and %v for everything else?
2. TIL!

pkg/template/event_test.go
```diff
@@ -28,628 +27,7 @@ import (
...
-		t.Run(tt.name, func(t *testing.T) {
-			got, err := applyHeaderToParam(tt.args.header, tt.args.param)
-			if err == nil {
-				t.Errorf("applyHeaderToParam() did not return error when expected; got: %v", got)
-			}
-		})
-	}
-}
-
+// TODO: 2. Split NewResourcesTests into separate tests for ResolveParams and ResolveResources
```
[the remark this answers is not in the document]
dibyom: Ahh, I opened the issue and forgot to update the comment :man_facepalming:

pkg/template/event.go
```diff
@@ -18,177 +18,79 @@ package template
...
+		Header: joinedHeaders,
+		Body:   data,
+	}, nil
+}
+
+// applyEventValuesToParams returns a slice of Params with the JSONPath variables replaced
+// with values from the event body and headers.
+func applyEventValuesToParams(params []pipelinev1.Param, ec *event) ([]pipelinev1.Param, error) {
+	for idx, p := range params {
+		// TODO: We need tests for binding values that are not jsonpath expressions
```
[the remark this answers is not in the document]
dibyom: Actually added the tests so its not a TODO anymore :sweat_smile:

pkg/template/jsonpath_test.go
dibyom: Neat!

docs/triggerbindings.md
```diff
@@ -33,46 +33,44 @@ Each parameter has a `name` and a `value`.
...
-In order to parse generic events as efficiently as possible,
-[GJSON](https://github.com/tidwall/gjson) is used internally. As a result, the
-binding [path syntax](https://github.com/tidwall/gjson#path-syntax) differs
-slightly from standard JSON. As of now, the following patterns are supported
-within `TriggerBinding` parameter value interpolation: -
-`\$\(body(\.[[:alnum:]/_\-\.\\]+|\.#\([[:alnum:]=<>%!"\*_-]+\)#??)*\)` -
-`\$\(header(\.[[:alnum:]_\-]+)?\)`
+TriggerBindings can access values from the HTTP JSON body and the headers using
+JSONPath expressions. The expressions have to be wrapped in `$()` but can omit
+the curly braces `{}` and the leading `.`.
```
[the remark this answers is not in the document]
dibyom: https://kubernetes.io/docs/reference/kubectl/jsonpath/ mentions the enclosing `{}` though it also says its a JSONPath "template"....maybe we keep things simple and never mention the `{}`?

pkg/template/event.go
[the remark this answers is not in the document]
dibyom: Done!

pkg/template/event.go
```diff
@@ -18,177 +18,79 @@ package template
...
 	for i := range template.Spec.ResourceTemplates {
 		resources[i] = ApplyParamsToResourceTemplate(params, template.Spec.ResourceTemplates[i].RawMessage)
-		resources[i] = ApplyUIDToResourceTemplate(resources[i], uid)
+		resources[i] = ApplyUIDToResourceTemplate(resources[i], UID())
 	}
 	return resources
 }
+
+// event represents a HTTP event that Triggers processes
+type event struct {
```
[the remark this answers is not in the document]
dibyom: might be confusing with `http.Request`. Also, I like event since we can later add the one-offs such as the `$(uid)` to this struct to unify how we do the "templating"

pkg/template/event.go
```diff
@@ -18,177 +18,79 @@ package template
...
+		}
+	}
+	joinedHeaders := make(map[string]string, len(headers))
+	for k, v := range headers {
+		joinedHeaders[k] = strings.Join(v, ",")
+	}
+
+	return &event{
+		Header: joinedHeaders,
+		Body:   data,
```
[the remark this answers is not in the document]
dibyom: Yeah I added that check after verifying the empty body case: https://github.com/tektoncd/triggers/pull/241/files/4fe79f8e2d17b38e31d2d871c983a545cb06db01#diff-8d7c043b29f6a04a92258e1aac64fc2bR44

pkg/template/jsonpath_test.go
dibyom: Flipped!

pkg/template/jsonpath_test.go
dibyom: done!

pkg/template/resource.go
dibyom: Opened #257 to discuss this.

pkg/template/jsonpath.go
dibyom: Given the discussion around JSONPath support in pipeline params, it might be worth having a common jsonpath pkg that both pipeline and triggers could use :thinking: 

(Though that's a yak that we should shave another day :smile:

dibyom:
> i think mostly you've done a great job, e.g. unit tests for ParseJSONPath but there are couple of functions which i think could also use tests which you probably skipped b/c they are unexported, and i think many of the functions (e.g. relaxedJSONPathExpression) are significant enough to actually be exported and have their own tests

The current `TektonJSONPathExpression` tests does test all the behaviors of RelaxedJSONPath expression. I do see the point about the behavior being easier to grok if some of the unexported functions have their own tests though I'm not sure if we should be exporting functions solely because they are complex or for testing.

pkg/template/jsonpath.go
dibyom: removed

docs/triggerbindings.md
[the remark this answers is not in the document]
dibyom: Removed the curly braces from the examples!

dibyom:
Ok, I think I addressed all the comments! 

Also, added a few more tests (header values, more relaxed jsonpath expressions etc.) 

There area a couple of if branches that are hard to test e.g. we only reach there if the unmarshalled value is a Function or a Channel --I don't think any HTTP JSON payload can be unmarshalled to that!
 
Quite a few followups from this:
* Escaping strings (#258, #257)
* Values for headers are sometimes inconsistent (#260)
* "Typed" Binding values? (#253)
* Keep existing behavior for 0.2, will look into changing for 0.3/
* More focussed tests (#252)

/hold cancel

bobcatfish:
Thanks for all the follow up issues @dibyom !! :heart:

pkg/template/event.go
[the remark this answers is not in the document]
bobcatfish: ive never used `%w` before! TIL :D

pkg/template/event.go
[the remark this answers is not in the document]
bobcatfish: > might be confusing with http.Request

it basically _is_ an http.Request tho since it has a header + a body :thinking: 

I see what you mean about the `uid` tho! you're probably right about `event` being better in the long run, sgtm :D

pkg/template/event.go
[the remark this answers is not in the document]
bobcatfish: i cant see that link anymore - good ol rebasing :sweat_smile:

pkg/template/jsonpath_test.go
```diff
@@ -0,0 +1,201 @@
...
+		"body)",
+		"body",
+		"$(body.missing)",
+		"$(body.key[0])",
+	}
+	var data interface{}
+	err := json.Unmarshal([]byte(testJSON), &data)
+	if err != nil {
+		t.Fatalf("Could not unmarshall body : %q", err)
+		return
```
bobcatfish: if you fatalf then you wont need to return b/c fatalf will halt execution

pkg/template/jsonpath_test.go
bobcatfish: i think this line could be fatalf and then no return needed

pkg/template/jsonpath.go
dibyom: added to commit message!

pkg/template/jsonpath_test.go
dibyom: done!

pkg/template/event.go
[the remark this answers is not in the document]
dibyom: :face_with_head_bandage:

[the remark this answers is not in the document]
dibyom: https://github.com/tektoncd/triggers/pull/241/files?file-filters%5B%5D=.go#diff-8d7c043b29f6a04a92258e1aac64fc2bR42

PR #242, opened 2019-11-27
Migrate EventListeners with binding to bindings

dibyom:
# Changes

After the 0.1 release, we replaced the `binding` field
with `bindings`. This commit migrates those to use the
`bindings` field. Without this, users who perform a controller
upgrade from the 0.1 release will not be able to update any
existing `EventListeners` i.e. they would have to delete and
recreate.

/cc @vdemeester @afrittoli @wlynch 
# Submitter Checklist

review them:

### DIFF
--- a/pkg/apis/triggers/v1alpha1/contexts.go
+++ b/pkg/apis/triggers/v1alpha1/contexts.go
@@ -1,0 +1,22 @@
+package v1alpha1
+
+import "context"
+
+// upgradeViaDefaultingKey is used as the key in a context.Context.
+// This variable doesn't really matter, so it can be a total random name.
+// Setting this key indicates that default values for a resource should be
+// updated to new values. This is used to ensure non breaking updates when
+// a default value of a resource changes or when a field is removed.
+type upgradeViaDefaultingKey struct{}
+
+// WithUpgradeViaDefaulting sets the upgradeViaDefaultingKey on the context
+// indicating that default values for a resource should be updated to new values.
+func WithUpgradeViaDefaulting(ctx context.Context) context.Context {
+	return context.WithValue(ctx, upgradeViaDefaultingKey{}, struct{}{})
+}
+
+// IsUpgradeViaDefaulting checks if the upgradeViaDefaultingKey is set on
+// the context.
+func IsUpgradeViaDefaulting(ctx context.Context) bool {
+	return ctx.Value(upgradeViaDefaultingKey{}) != nil
+}
--- a/pkg/apis/triggers/v1alpha1/event_listener_defaults.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_defaults.go
@@ -5,4 +5,22 @@ import (
 )
 
 // SetDefaults sets the defaults on the object.
-func (el *EventListener) SetDefaults(ctx context.Context) {}
+func (el *EventListener) SetDefaults(ctx context.Context) {
+	if IsUpgradeViaDefaulting(ctx) {
+		// Most likely the EventListener passed here is already running
+		for i := range el.Spec.Triggers {
+			t := &el.Spec.Triggers[i]
+			if t.DeprecatedBinding != nil {
+				if len(t.Bindings) > 0 {
+					// Do nothing since it will be a Validaiton Error.
+				} else {
+					// Set the binding to bindings
+					t.Bindings = append(t.Bindings, &EventListenerBinding{
+						Name: t.DeprecatedBinding.Name,
+					})
+					t.DeprecatedBinding = nil
+				}
+			}
+		}
+	}
+}
--- a/pkg/apis/triggers/v1alpha1/event_listener_defaults_test.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_defaults_test.go
@@ -1,0 +1,121 @@
+package v1alpha1_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
+)
+
+func TestEventListenerSetDefaults(t *testing.T) {
+	tests := []struct {
+		name string
+		in   *v1alpha1.EventListener
+		want *v1alpha1.EventListener
+		wc   func(context.Context) context.Context
+	}{{
+		name: "with upgrade context - binding present",
+		in: &v1alpha1.EventListener{
+			Spec: v1alpha1.EventListenerSpec{
+				Triggers: []v1alpha1.EventListenerTrigger{{
+					DeprecatedBinding: &v1alpha1.EventListenerBinding{
+						Name: "my-binding",
+					},
+				}},
+			},
+		},
+		want: &v1alpha1.EventListener{
+			Spec: v1alpha1.EventListenerSpec{
+				Triggers: []v1alpha1.EventListenerTrigger{{
+					Bindings: []*v1alpha1.EventListenerBinding{{
+						Name: "my-binding",
+					}},
+				}},
+			},
+		},
+		wc: v1alpha1.WithUpgradeViaDefaulting,
+	}, {
+		name: "with upgrade context - no binding present",
+		in: &v1alpha1.EventListener{
+			Spec: v1alpha1.EventListenerSpec{
+				Triggers: []v1alpha1.EventListenerTrigger{{
+					Bindings: []*v1alpha1.EventListenerBinding{{
+						Name: "my-binding",
+					}},
+				}},
+			},
+		},
+		want: &v1alpha1.EventListener{
+			Spec: v1alpha1.EventListenerSpec{
+				Triggers: []v1alpha1.EventListenerTrigger{{
+					Bindings: []*v1alpha1.EventListenerBinding{{
+						Name: "my-binding",
+					}},
+				}},
+			},
+		},
+		wc: v1alpha1.WithUpgradeViaDefaulting,
+	}, {
+		name: "with upgrade context - both binding and bindings present",
+		in: &v1alpha1.EventListener{
+			Spec: v1alpha1.EventListenerSpec{
+				Triggers: []v1alpha1.EventListenerTrigger{{
+					DeprecatedBinding: &v1alpha1.EventListenerBinding{
+						Name: "my-binding-1",
+					},
+					Bindings: []*v1alpha1.EventListenerBinding{{
+						Name: "my-binding",
+					}},
+				}},
+			},
+		},
+		want: &v1alpha1.EventListener{
+			Spec: v1alpha1.EventListenerSpec{
+				Triggers: []v1alpha1.EventListenerTrigger{{
+					DeprecatedBinding: &v1alpha1.EventListenerBinding{
+						Name: "my-binding-1",
+					},
+					Bindings: []*v1alpha1.EventListenerBinding{{
+						Name: "my-binding",
+					}},
+				}},
+			},
+		},
+		wc: v1alpha1.WithUpgradeViaDefaulting,
+	}, {
+		name: "no upgrade context",
+		in: &v1alpha1.EventListener{
+			Spec: v1alpha1.EventListenerSpec{
+				Triggers: []v1alpha1.EventListenerTrigger{{
+					DeprecatedBinding: &v1alpha1.EventListenerBinding{
+						Name: "my-binding",
+					},
+				}},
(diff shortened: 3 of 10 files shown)


pkg/apis/triggers/v1alpha1/event_listener_defaults.go
```diff
@@ -5,4 +5,24 @@ import (
...
-func (el *EventListener) SetDefaults(ctx context.Context) {}
+func (el *EventListener) SetDefaults(ctx context.Context) {
+	if IsUpgradeViaDefaulting(ctx) {
+		// Most likely those EventListener passed here is already running
+		for i := range el.Spec.Triggers {
+			t := &el.Spec.Triggers[i]
+			if t.DeprecatedBinding != nil {
+				if len(t.Bindings) > 0 {
+					// Both binding & bindings are set...we should validate for this?
+					// Maybe we should check to see if deprecatedBinding is present in Bindings?
```
afrittoli: Yeah. I think should make sure this cannot happen via validation.
For old resources, `bindings` was not defined, so will not be there. 
For new resources, if they try to define both `bindings` and `binding`, they should not be accepted.

dibyom: Added validation for this!

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -68,6 +68,9 @@ type EventListenerTrigger struct {
 	// +optional
 	// TODO(dibyom): Allow multiple interceptors
 	Interceptor *EventInterceptor `json:"interceptor,omitempty"`
+
+	// TODO(dibyom): Remove this before 0.3 release
```
bobcatfish: instead of TODO(name) i think its better in the long run to open an issue (or refer to an existing issue) e.g. TODO(#11111) - i think we'll have a better chance of remembering if there's an issue and less of a chance of stale TODOs floating around (we'll be able to tell when they're done - and we can even add automation if we want!)

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
```diff
@@ -23,6 +23,23 @@ func Test_EventListenerValidate_error(t *testing.T) {
...
+			Spec: v1alpha1.EventListenerSpec{
+				Triggers: []v1alpha1.EventListenerTrigger{{
+					Bindings:          []*v1alpha1.EventListenerBinding{{Name: "tb"}},
+					DeprecatedBinding: &v1alpha1.EventListenerBinding{Name: "bar"},
+					Template:          v1alpha1.EventListenerTemplate{Name: "tt"},
+				}},
+			},
+		},
+
+		raiseErr: true,
```
bobcatfish: in general i find that the tests are easier to read and maintain if we separate the error cases from the success cases into completely different functions

pkg/template/resource.go
```diff
@@ -42,7 +42,7 @@ type ResolvedTrigger struct {
 type getTriggerBinding func(name string, options metav1.GetOptions) (*triggersv1.TriggerBinding, error)
 type getTriggerTemplate func(name string, options metav1.GetOptions) (*triggersv1.TriggerTemplate, error)
 
-// ResolveBindings takes in a trigger containing object refs to bindings and
+// ResolveTrigger takes in a trigger containing object refs to bindings and
```
bobcatfish: :innocent:

pkg/apis/triggers/v1alpha1/contexts.go
```diff
@@ -0,0 +1,20 @@
...
+type upgradeViaDefaultingKey struct{}
+
+// WithUpgradeViaDefaulting notes on the context that we want defaulting to rewrite
+// from v1alpha1 pre-defaults to v1alpha1 post-defaults.
+func WithUpgradeViaDefaulting(ctx context.Context) context.Context {
+	return context.WithValue(ctx, upgradeViaDefaultingKey{}, struct{}{})
+}
+
+// IsUpgradeViaDefaulting checks whether we should be "defaulting" from v1alpha1 pre-defaults to
+// the v1alpha1 post-defaults subset.
```
bobcatfish: in these 2 comments im a bit confused by what 'defaulting" from v1alpha1 pre-defaults to the v1alpha1 post-defaults subset'  means - the quotes around "defaulting" make me also feel like im definitely missing something, maybe some more words in these docstrings could clear this up?

pkg/apis/triggers/v1alpha1/event_listener_defaults.go
```diff
@@ -5,4 +5,23 @@ import (
 )
 
 // SetDefaults sets the defaults on the object.
-func (el *EventListener) SetDefaults(ctx context.Context) {}
+func (el *EventListener) SetDefaults(ctx context.Context) {
+	if IsUpgradeViaDefaulting(ctx) {
+		// Most likely those EventListener passed here is already running
```
bobcatfish: "most likely the EventListeners passed here are already running"?

pkg/apis/triggers/v1alpha1/event_listener_validation.go
```diff
@@ -44,16 +44,19 @@ func (s *EventListenerSpec) validate(ctx context.Context, el *EventListener) *ap
...
 	clientset := ctx.Value("clientSet").(dynamic.Interface)
-
 	for i, t := range s.Triggers {
+		// Validate that only one of binding or bindings is set
+		if t.DeprecatedBinding != nil && len(t.Bindings) > 0 {
+			return apis.ErrMultipleOneOf(fmt.Sprintf("spec.triggers[%d].binding", i), fmt.Sprintf("spec.triggers[%d].binding", i))
+		}
 		// Validate optional TriggerBinding
-		for _, b := range t.Bindings {
+		for j, b := range t.Bindings {
```
bobcatfish: i _think_ this happens before `SetDefaults` is called, so we might want to validate the content of `DepreatedBinding` as well?

pkg/apis/triggers/v1alpha1/event_listener_defaults_test.go
```diff
@@ -0,0 +1,118 @@
...
+		name: "with upgrade context - both binding and bindings present",
+		in: &v1alpha1.EventListener{
+			Spec: v1alpha1.EventListenerSpec{
+				Triggers: []v1alpha1.EventListenerTrigger{{
+					DeprecatedBinding: &v1alpha1.EventListenerBinding{
+						Name: "my-binding",
+					},
+					Bindings: []*v1alpha1.EventListenerBinding{{
+						Name: "my-binding",
+					}},
```
bobcatfish: what if you have both but different names?

pkg/apis/triggers/v1alpha1/event_listener_defaults.go
dibyom: :face_with_head_bandage:

pkg/apis/triggers/v1alpha1/event_listener_types.go
dibyom: Excellent idea!

pkg/apis/triggers/v1alpha1/event_listener_validation.go
dibyom: So, that's what I assumed too but it looks like `setDefaults` happens before: https://github.com/tektoncd/triggers/blob/master/vendor/knative.dev/pkg/webhook/webhook.go#L606:L622

pkg/apis/triggers/v1alpha1/event_listener_defaults_test.go
dibyom: Shouldn't matter. We just check if both fields exist; not what their names are

dibyom: Doesn't matter; we check if the field is present not what the name is!

pkg/apis/triggers/v1alpha1/contexts.go
dibyom: Added some more comments! The two functions are to set and check whether or not the key was set. So, I added some more comments around what they key is used for i.e. upgrade default values that have changed between releases.

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
dibyom: Agree in general but I feel like these tests are simple enough that a single function should work too. Unless you feel strongly about this, I'd prefer keeping this in now and maybe splitting it later if the tests become more complex.

pkg/apis/triggers/v1alpha1/event_listener_defaults_test.go
bobcatfish: I thought that deprecated bindings were auto-magically internally re-written to `Bindings`, therefore if you specified something like:

```
						DeprecatedBinding: &v1alpha1.EventListenerBinding{
						Name: "my-same-binding",
					},
					Bindings: []*v1alpha1.EventListenerBinding{{
						Name: "my-same-binding",
					}},
```

Then the values in `Bindings` would be overwritten and you'd end up with only one binding? Which I'm thinking maybe should be an error? (I might be misunderstanding how this works tho!)

bobcatfish: Or, looking at the `want` field below, is the behavior that if you have both, the Deprecated field _isnt_ updated?

What about making this an error instead - i.e. if you use both, it's an error?

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
bobcatfish: I'm afraid I do feel strongly! Besides the additional complexity i think its much easier to see what cases are covered if you can look at a list of error cases completely separately from a list of success cases.

Looking at the body of the text i can see that a lot of setup happens that would theoretically need to be duplicated in both sets of tests, is that why you want to keep it in one test?

(also i think this is total scope creep since you are just updating this test in this PR tho so im happy to just let it go)

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -66,8 +67,11 @@ type EventListenerTrigger struct {
 	// +optional
 	Name string `json:"name,omitempty"`
 	// +optional
-	// TODO(dibyom): Allow multiple interceptors
+	// TODO(#249): Allow multiple interceptors
 	Interceptor *EventInterceptor `json:"interceptor,omitempty"`
+
+	// TODO(#248): Remove this before 0.3 release
```
bobcatfish: 🎉 thanks!!

pkg/apis/triggers/v1alpha1/contexts.go
```diff
@@ -0,0 +1,22 @@
+package v1alpha1
+
+import "context"
+
+// upgradeViaDefaultingKey is used as the key in a context.Context.
+// This variable doesn't really matter, so it can be a total random name.
+// Setting this key indicates that default values for a resource should be
+// updated to new values. This is used to ensure non breaking updates when
+// a default value of a resource changes or when a field is removed.
```
bobcatfish: 🎉

pkg/apis/triggers/v1alpha1/event_listener_defaults_test.go
dibyom: Yeah, the flow is a bit confusing -- in the webhook setDefaults is called first, followed by the Validate. So, if both Binding and Bindings are present, SetDefaults does nothing, but the Validate function (which is called after SetDefaults) returns an Error. The end result is that the user gets a Validation error.

In the case where the user only has binding set (i.e. upgrading from 0.1 to 0.2), we'll move the binding value to bindings,

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
dibyom: I don't feel too strongly about this but since I was just adding a test case, wanted to keep the scope limited. I'll make the change in a follow up

pkg/apis/triggers/v1alpha1/event_listener_defaults_test.go
dibyom: Also, I think we'd have both binding and bindings set is if a user manually creates (or updates) a TriggerBinding to have both `binding` and `bindings`. In such cases, rather than silently moving the binding to bindings, I think returning a validation error is the way to go!

PR #247, opened 2019-12-03
Add support for interceptor HTTP headers

ston1th:
# Changes

This adds support to access HTTP headers in TriggerBindings returned by a webhook interceptor.

If someone wants keep the initial webhook body intact, this method can be used, to return additional information back to the eventlistener.

# Submitter Checklist

review them:

# Release Notes

```
- The interface interceptors.Interceptor changed (ExecuteTrigger now returns an additional http.Header)
- HTTP headers returned from a webhook interceptor overwrite the initial webhook headers.
```


wlynch:
Thanks for the PR!

I think this raises an interesting question about how to handle headers. This change leans on additive behavior, but what if we want to modify headers?

Should we just just return an HTTP response, or take an approach similar to [http.HandlerFunc](https://godoc.org/net/http#HandlerFunc) instead? /cc @dibyom @dlorenc

dibyom:
Yeah, this is interesting. (I think we should discuss the use cases in a WG/issue before adding this)

Is the main benefit to doing this that it is easier to add to the headers instead of modifying the body?
Are there use cases for modifying/removing headers instead of adding them? And if we change the contract to return both headers and a body, returning a HTTP response makes sense to me!

ston1th:
@wlynch Yes returning a `http.Response` would be a better approach.

@dibyom 
Yes, I think it is easier to add additional information as headers and keep the (mostly) big webhook bodies intact.

A example usecase would be receiving a issue comment webhook with a body such as `/test`. As this webhook does not contain information like the git revision it is not easy to use such webhook to trigger test builds.

So the webhook interceptor needs to call out to the GitHub API fetching the latest PR commit.
This new information then needs to find its way back to the EventListener and I find it a bit overkill to patch this field `{"head":{"sha":"<value>"}}` into the response body using something like jsonpatch.

About modifying/removing headers:

I think we should not do this, since this could lead to somewhat unexpected behavior. I find the additive way the least invasive and more backwards compatible.

dibyom:
> About modifying/removing headers:
> 
> I think we should not do this, since this could lead to somewhat unexpected behavior. I find the additive way the least invasive and more backwards compatible.

Sounds good to me!

ston1th:
@dibyom great :)

I'll then start rewriting this patch next week, to return a `http.Response`.

wlynch:
Let's go ahead and review this without the http.Response change. I suspect that will be a fairly large change on it's own, since we'll need to modify the existing interceptors.

I'll be sure to take a closer look at this early tomorrow!

wlynch:
> > About modifying/removing headers:
> > I think we should not do this, since this could lead to somewhat unexpected behavior. I find the additive way the least invasive and more backwards compatible.
> 
> Sounds good to me!

My gut feeling is that some workflows are going to want to append headers and some will want to rewrite headers, and that this behavior should be decided on a per interceptor basis and not for all interceptors. The ideal end state in my mind is something similar to [`RoundTripper`](https://godoc.org/net/http#RoundTripper). This gives users the flexibility to either merge or override headers as desired.

For the time being for compatibility we can return the incoming headers (github/gitlab), or return the headers in the HTTP response (webhook). Thoughts?

dibyom:
> Let's go ahead and review this without the http.Response change. I suspect that will be a fairly large change on it's own, since we'll need to modify the existing interceptors.

I think adding it in a followup is fine since that's not a user facing change.

> My gut feeling is that some workflows are going to want to append headers and some will want to rewrite headers, and that this behavior should be decided on a per interceptor basis and not for all interceptors. The ideal end state in my mind is something similar to RoundTripper. This gives users the flexibility to either merge or override headers as desired.

Yeah, it should very similar to `RoundTripper` though there'd be a couple of differences e.g. around closing the request bodies etc. Also, RoundTripper does not care about statusCodes though we do

> For the time being for compatibility we can return the incoming headers (github/gitlab), or return the headers in the HTTP response (webhook). Thoughts?

Not sure I fully understand this -- are you saying that instead of returning a separate "header" return field, the webhook interceptors just modify the response?

wlynch:
> > For the time being for compatibility we can return the incoming headers (github/gitlab), or return the headers in the HTTP response (webhook). Thoughts?
> 
> Not sure I fully understand this -- are you saying that instead of returning a separate "header" return field, the webhook interceptors just modify the response?

My understanding (though I might be wrong :sweat_smile:) is that we wanted to append headers as opposed to full modify to avoid unexpected / backwards incompatible changes since interceptors currently not returning any headers. I think we'll want to pass through headers for existing core interceptors, and we'll need to make a decision with webhook interceptors - append or use whatever is returned (my preference is use whatever is returned).

ston1th:
> append or use whatever is returned (my preference is use whatever is returned).

So if I understand this correctly this can only be accomplished with a config option in the eventlister?

Something like:

```
...
      interceptor:
        webhook:
          headerMergeMethod: [override|merge]
          header:
          - name: X-Header
            value: value
          objectRef:
            kind: Service
            ...
```

With:
* override - just using all the headers returned by the webhook interceptor
* merge - doing the additional merge we already have

wlynch:
Discussed this in the triggers working group today. The consensus was we should follow the model of RoundTripper and default to replacing. Interceptors that want to merge received headers should do so in their own handlers.

ston1th:
@wlynch I updated the PR to overwrite the original headers. What do you think?

pkg/sink/sink.go
```diff
@@ -102,12 +102,15 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
 			finalPayload := event
 			if interceptor != nil {
-				payload, err := interceptor.ExecuteTrigger(event, request, &t, eventID)
+				payload, header, err := interceptor.ExecuteTrigger(event, request, &t, eventID)
 				if err != nil {
 					log.Error(err)
 					result <- http.StatusAccepted
 					return
 				}
+				if header != nil {
```
wlynch: Is a response with no headers treated as nil or an empty map? 

We might just want to explicitly return req.Headers everywhere explicitly.

ston1th: Done. Now I return an empty map instead of nil.

docs/eventlisteners.md
```diff
@@ -85,6 +85,9 @@ To be an Event Interceptor, a Kubernetes object should:
 * Accept JSON payloads over HTTP
 * Return a HTTP 200 OK Status if the EventListener should continue processing
   the event
+* Return HTTP headers back. These will be used by the EventListener as the event
```
vtereso: I think this makes senses, but overall maybe we could consolidate the bullets here?

pkg/interceptors/webhook/sink.go
```diff
@@ -50,27 +50,28 @@ func NewInterceptor(wh *triggersv1.WebhookInterceptor, c *http.Client, ns string
 	}
 }
 
-func (w *Interceptor) ExecuteTrigger(payload []byte, request *http.Request, trigger *triggersv1.EventListenerTrigger, eventID string) ([]byte, error) {
+func (w *Interceptor) ExecuteTrigger(payload []byte, request *http.Request, trigger *triggersv1.EventListenerTrigger, eventID string) ([]byte, http.Header, error) {
```
vtereso: The Interceptor interface seems very strange with regard for the method signature `ExecuteTrigger(payload []byte, request *http.Request, trigger *triggersv1.EventListenerTrigger, eventID string) ([]byte, error)`. The passed trigger is never used and since we only have validator interceptors (no decorators), the payload is always returned as is. I feel like ```type Interceptor interface {
	Intercept(req *http.Request) (*http.Response, error)
}``` might make more sense, where the payload and headers should be passed within the response. Any initialization/decoration of the original event request would be done before the interceptor chain (e.g. add eventID/triggers headers). Doesn't need to be done here, but just as a followup. @dibyom @ncskier @wlynch

dibyom: See #269 which is similar

vtereso: Right, I remember now Billy mentioned on the call.

docs/eventlisteners.md
ston1th: What exactly do you mean?

ston1th:
@wlynch @dibyom 
Since #326 this PR seems to be obsolete.

Also there are many changes in #272 which would need a rewrite.

dibyom:
Ok, then how about we close this and go ahead with #326?

PR #196, opened 2019-10-30
Listener Response in Json Format

khrm:
# Changes
EventListener Response is sent in json now
Fixes https://github.com/tektoncd/triggers/issues/194

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/triggers/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

```
Response of EventListener comes in Json format.
```

PR #261, opened 2019-12-10
Change sink response to return JSON

### LINKED ISSUE #194: EventListener response is pod log message not parseable JSON.
I've created a very simple test pipeline that just echos out the message it is sent and I'm experimenting with triggers and event listeners to invoke it from a script.
When the pipeline is triggered using curl the request it is sent contains a JSON body and the event listener picks the parameters out of the body (and could pick them from fields the header too) in a nice structured way.

However the message that is sent back seems to be just a copy of the log message written to the listeners pod's stdout (minus the timestamp) and not really something intended to be parsed. e.g.
```
$ curl -X POST http://el-echolistener:8080 --header "Content-Type: application/json" --data '{"message":"Hello World"}'  
EventListener: echolistener in Namespace: default handling event (EventID: lb7hq) with payload: {"message":"Hello World"} and header: map[Accept:[*/*] Content-Length:[25] Content-Type:[application/json] User-Agent:[curl/7.58.0]]
```
This isn't ideal as it's necessary to parse the event id out of that message in order to use a selector like `tekton.dev/triggers-eventid=lb7hq` to get, for example, the pod containing the logs for the PipelineRun. Given that the request is JSON formatted it would have been good to get a JSON response.

# Expected Behavior
A JSON response, e.g.
```json
{
  "eventid": "lb7hq"
}
```
(It could also include back the original body and headers or even a copy of the params from the TriggerBinding showing which names were given which values from the request.)

# Actual Behavior
A response like:
```
EventListener: echolistener in Namespace: default handling event (EventID: lb7hq) with payload: {"message":"Hello World"} and header: map[Accept:[*/*] Content-Length:[25] Content-Type:[application/json] User-Agent:[curl/7.58.0]]
```
This is the same as the log message written to the EventListeners pod, minus timestamps.

1. Create a pipeline with an EventListener.
2. Send a request to the event listener service, for example with curl:
`curl -X POST http://el-echolistener:8080 --header "Content-Type: application/json" --data '{"message":"Hello World"}' `
3. The response will be something like:
`EventListener: echolistener in Namespace: default handling event (EventID: lb7hq) with payload: {"message":"Hello World"} and header: map[Accept:[*/*] Content-Length:[25] Content-Type:[application/json] User-Agent:[curl/7.58.0]]`

# Additional Info
Discussion on #triggers here:
https://tektoncd.slack.com/archives/CKUSJ2A5D/p1572363277099400

dibyom:
# Changes

Return a valid JSON response instead of plain-text
Fixes #194 

(This is a port of another PR: #196)

# Submitter Checklist

review them:

# Release Notes

```
The EventListener sink now returns a valid JSON response instead of plain-text
```

### DIFF
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -18,7 +18,6 @@ package sink
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 
@@ -54,6 +53,16 @@ type Sink struct {
 	EventListenerName      string
 	EventListenerNamespace string
 	Logger                 *zap.SugaredLogger
+}
+
+// Response defines the HTTP body that the Sink responds to events with.
+type Response struct {
+	// EventListener is the name of the eventListener
+	EventListener string `json:"eventListener"`
+	// Namespace is the namespace that the eventListener is running in
+	Namespace string `json:"namespace,omitempty"`
+	// EventID is a uniqueID that gets assigned to each incoming request
+	EventID string `json:"eventID,omitempty"`
 }
 
 // HandleEvent processes an incoming HTTP event for the event listener.
@@ -139,10 +148,16 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 		}
 	}
 
-	// TODO: Do we really need to return the entire body back???
 	response.WriteHeader(code)
-	fmt.Fprintf(response, "EventListener: %s in Namespace: %s handling event (EventID: %s) with payload: %s and header: %v",
-		r.EventListenerName, r.EventListenerNamespace, string(eventID), string(event), request.Header)
+	response.Header().Set("Content-Type", "application/json")
+	body := Response{
+		EventListener: r.EventListenerName,
+		Namespace:     r.EventListenerNamespace,
+		EventID:       eventID,
+	}
+	if err := json.NewEncoder(response).Encode(body); err != nil {
+		eventLog.Errorf("failed to write back sink response: %w", err)
+	}
 }
 
 func (r Sink) createResources(res []json.RawMessage, triggerName, eventID string) error {
--- a/pkg/sink/sink_test.go
+++ b/pkg/sink/sink_test.go
@@ -22,7 +22,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
-	"strings"
 	"sync"
 	"testing"
 
@@ -144,20 +143,27 @@ func TestHandleEvent(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Error creating Post request: %s", err)
 	}
+
 	if resp.StatusCode != http.StatusCreated {
 		t.Fatalf("Response code doesn't match: %v", resp.Status)
 	}
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		t.Fatalf("Error reading response body: %s", err)
-		return
 	}
-	if !strings.Contains(string(body), "event (EventID: ") {
-		t.Errorf("Response body doesn't have eventid")
+
+	wantBody := Response{
+		EventListener: el.Name,
+		Namespace:     el.Namespace,
+		EventID:       eventID,
 	}
-	wantPayload := `EventListener: my-eventlistener in Namespace: foo handling event`
-	if !strings.Contains(string(body), wantPayload) {
-		t.Errorf("Diff response body: %s, should have: %s", string(body), wantPayload)
+
+	got := Response{}
+	if err := json.Unmarshal(body, &got); err != nil {
+		t.Fatalf("Error unmarshalling response body: %s", err)
+	}
+	if diff := cmp.Diff(wantBody, got); diff != "" {
+		t.Errorf("did not get expected response back -want,+got: %s", diff)
 	}
 
 	wg.Wait()
--- a/test/eventlistener_test.go
+++ b/test/eventlistener_test.go
@@ -24,16 +24,17 @@ import (
 	"fmt"
 	"net/http"
 	"os/exec"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
 
-	"strconv"
-
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	eventReconciler "github.com/tektoncd/triggers/pkg/reconciler/v1alpha1/eventlistener"
+	"github.com/tektoncd/triggers/pkg/sink"
 	bldr "github.com/tektoncd/triggers/test/builder"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -284,9 +285,27 @@ func TestEventListenerCreate(t *testing.T) {
 		t.Fatalf("Error creating POST request: %s", err)
 	}
 	req.Header.Add("Content-Type", "application/json")
-	_, err = http.DefaultClient.Do(req)
+	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		t.Fatalf("Error sending POST request: %s", err)
+	}
+
+	if resp.StatusCode > http.StatusAccepted {
+		t.Errorf("sink did not return 2xx response. Got status code: %d", resp.StatusCode)
+	}
+	wantBody := sink.Response{
+		EventListener: "my-eventlistener",
+		Namespace:     namespace,
+	}
+	var gotBody sink.Response
+	if err := json.NewDecoder(resp.Body).Decode(&gotBody); err != nil {
+		t.Fatalf("failed to read/decode sink response: %v", err)
+	}
+	if diff := cmp.Diff(wantBody, gotBody, cmpopts.IgnoreFields(sink.Response{}, "EventID")); diff != "" {
+		t.Errorf("unexpected sink response -want/+got: %s", diff)
+	}
+	if gotBody.EventID == "" {
+		t.Errorf("sink response no eventID")
 	}
 
 	for _, wantPr := range []v1alpha1.PipelineResource{wantPr1, wantPr2} {


pkg/sink/sink.go
```diff
@@ -145,10 +145,26 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
+		},
+	)
+	if err != nil {
+		http.Error(response, "Error Marshaling response"+
+			err.Error(), http.StatusInternalServerError)
+		return
+	}
+
 	response.WriteHeader(code)
-	fmt.Fprintf(response, "EventListener: %s in Namespace: %s handling event (EventID: %s) with payload: %s and header: %v",
```
wlynch: Make sure this (or something equivalent) is logged somewhere. This is useful for debugging whether requests were actually received or not.

pkg/sink/sink.go
```diff
@@ -145,10 +145,26 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
-		r.EventListenerName, r.EventListenerNamespace, string(eventID), string(event), request.Header)
+	response.Header().Set("Content-Type", "application/json")
+	_, _ = response.Write(js)
```
wlynch: You can simplify this by doing: 

```go
data := struct {
  EventListener string
  Namespace     string
  EventID       string
}{
  r.EventListenerName,
  r.EventListenerNamespace,
  eventID,
}
response.WriteHeader(code)
response.Header().Set("Content-Type", "application/json")
json.NewEncoder(response).Encode(data)
```

pkg/sink/sink_test.go
```diff
@@ -365,20 +364,35 @@ func TestHandleEvent(t *testing.T) {
...
 	}
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		t.Fatalf("Error reading response body: %s", err)
-		return
 	}
-	if !strings.Contains(string(body), "event (EventID: ") {
-		t.Errorf("Response body doesn't have eventid")
+
+	respData := struct {
```
wlynch: This is something that a lot of clients are also going to want to do. We should probably just make this an explicit type. This will also let us set JSON tags to set key values (e.g. "event_listener" vs "EventListener")

pkg/sink/sink_test.go
```diff
@@ -365,20 +364,35 @@ func TestHandleEvent(t *testing.T) {
...
+		t.Fatalf("EventID isn't available: %s", err)
+	}
+	if respData.EventListener != el.Name {
+		t.Errorf("Diff EventListener name: %s, should: %s", respData.EventListener, el.Name)
 	}
-	wantPayload := `EventListener: my-eventlistener in Namespace: foo handling event`
-	if !strings.Contains(string(body), wantPayload) {
-		t.Errorf("Diff response body: %s, should have: %s", string(body), wantPayload)
+
+	if respData.Namespace != el.Namespace {
```
wlynch: optional: Since UID can now be set deterministicly in tests, you could replace this with a diff call instead of doing ifs on each field.

pkg/sink/sink.go
dibyom: Yeah I think we already do this.

dibyom: Looks like its Debugf: https://github.com/tektoncd/triggers/blob/master/pkg/sink/sink.go#L83

Might be worth logging the eventID or something in Info and the body in Debug

pkg/sink/sink.go
```diff
@@ -145,10 +152,16 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
 	response.WriteHeader(code)
-	fmt.Fprintf(response, "EventListener: %s in Namespace: %s handling event (EventID: %s) with payload: %s and header: %v",
-		r.EventListenerName, r.EventListenerNamespace, string(eventID), string(event), request.Header)
+	response.Header().Set("Content-Type", "application/json")
+	err = json.NewEncoder(response).Encode(Response{
+		EventListener: r.EventListenerName,
+		Namespace:     r.EventListenerNamespace,
+		EventID:       eventID,
+	})
+	if err != nil {
```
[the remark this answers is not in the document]
dibyom: done

pkg/sink/sink_test.go
```diff
@@ -365,20 +364,28 @@ func TestHandleEvent(t *testing.T) {
...
+
+	wantBody := Response{
+		EventListener: el.Name,
+		Namespace:     el.Namespace,
+		EventID:       eventID,
+	}
+
+	got := Response{}
+	err = json.Unmarshal(body, &got)
+	if err != nil {
```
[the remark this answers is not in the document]
dibyom: done

pkg/sink/sink.go
```diff
@@ -62,6 +62,13 @@ type Sink struct {
 	Logger                 *zap.SugaredLogger
 }
 
+// Response defines the HTTP body that the Sink responds to events with.
+type Response struct {
```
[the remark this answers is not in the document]
dibyom: My only hesitation here is that we do not follow the lowercase and underscore pattern anywhere else in Tekton e.g. [ServiceAccountName](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1alpha1/pipelinerun_types.go#L56) or even [PipelineSpec](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1alpha1/pipelinerun_types.go#L46)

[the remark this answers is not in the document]
bobcatfish: I cant find hard evidence that the k8s standard is camelCase but looking at https://github.com/kubernetes/kubernetes/blob/release-1.1/docs/devel/api-conventions.md#typical-status-properties the example properties are camelCase and that seems to be the norm for other k8s types

pkg/sink/sink.go
```diff
@@ -62,6 +62,13 @@ type Sink struct {
...
+	EventListener string `json:"EventListener"`
+	Namespace     string `json:"namespace,omitempty"`
+	EventID       string `json:"eventID,omitempty"`
```
bobcatfish: could these fields have docstrings as well? e.g. i can't tell what kind of values ot expect in `EventListener` just looking at this

pkg/sink/sink.go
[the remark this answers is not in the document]
bobcatfish: i do think that `EventListener` should probably be `eventListener` for consistency?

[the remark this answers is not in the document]
wlynch: Ack. SGTM!

pkg/sink/sink.go
```diff
@@ -62,6 +62,13 @@ type Sink struct {
 	Logger                 *zap.SugaredLogger
 }
 
+// Response defines the HTTP body that the Sink responds to events with.
+type Response struct {
+	EventListener string `json:"EventListener"`
```
wlynch: eventListener

pkg/sink/sink.go
```diff
@@ -145,10 +152,16 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
-	fmt.Fprintf(response, "EventListener: %s in Namespace: %s handling event (EventID: %s) with payload: %s and header: %v",
-		r.EventListenerName, r.EventListenerNamespace, string(eventID), string(event), request.Header)
+	response.Header().Set("Content-Type", "application/json")
+	body := Response{
+		EventListener: r.EventListenerName,
+		Namespace:     r.EventListenerNamespace,
+		EventID:       eventID,
+	}
+	if err := json.NewEncoder(response).Encode(body); err != nil {
+		eventLog.Errorf("failed to write back sink response: %w", err)
```
bobcatfish: do we need to do anything else in this case, e.g. set the header or the result with an error (looks like above error handling does this?)

dibyom: Basically there are two cases where we'd get an error here:
1. The JSON encoding fails (this should basically never happen since we are setting just setting static values to a struct that we know of)\

2. Something goes wrong while sending the actual body out e.g. the underlying connection is closed

In both cases, the issue is with sending the response body back and not with the actual processing of the event. And in 2. we cannot guarantee that we can even successfully write the status code successfully.

The GitHub webhook best practices recommends sending a 500 only if it is a catastrophic error which I don't think this is since the actually trigger processing goes through and we only fail when writing out the response body.

dibyom:
> Could we add some docs about the sink response as well?

Add docstrings to the struct

> Do we have an end to end test that could be updated to verify the response?
Done

> Can we add a test for the error scenario in sink_test as well?
This is going to be hard to test since we are basically testing that we log the right error message:
https://github.com/tektoncd/triggers/pull/261#discussion_r357855907

test/eventlistener_test.go
```diff
@@ -284,11 +285,29 @@ func TestEventListenerCreate(t *testing.T) {
...
+	if err := json.NewDecoder(resp.Body).Decode(&gotBody); err != nil {
+		t.Fatalf("failed to read/decode sink response: %v", err)
+	}
+	if diff := cmp.Diff(wantBody, gotBody, cmpopts.IgnoreFields(sink.Response{}, "EventID")); diff != "" {
+		t.Errorf("unexpected sink response -want/+got: %s", diff)
+	}
+	if gotBody.EventID == "" {
+		t.Errorf("sink response no eventID")
+	}
+
```
bobcatfish: Thanks for adding this!! :tada: 

i think this test is getting super long - even more so with our changes queued up in #279 XD 

Maybe in a future PR we can start breaking this up into multiple functions .

test/eventlistener_test.go
```diff
@@ -284,11 +285,29 @@ func TestEventListenerCreate(t *testing.T) {
...
+	}
+	wantBody := sink.Response{
+		EventListener: "my-eventlistener",
+		Namespace:     namespace,
+	}
+	var gotBody sink.Response
+	if err := json.NewDecoder(resp.Body).Decode(&gotBody); err != nil {
+		t.Fatalf("failed to read/decode sink response: %v", err)
+	}
+	if diff := cmp.Diff(wantBody, gotBody, cmpopts.IgnoreFields(sink.Response{}, "EventID")); diff != "" {
```
bobcatfish: nice, i didnt know about `cmpopts.IgnoreFields`!

pkg/sink/sink.go
bobcatfish: ah okay makes sense! thanks for explaining :D

do you think this could be worth a comment in the code explaining it? (no worries if not!)

Issue #187, opened 2019-10-23
Fragile object creation order issues


kav:
# Expected Behavior
EventListeners, TriggerBindings, and TriggerTemplates can be installed by a helm chart, flux, jenkinsx gitops or other process that does not allow users to specify object creation order. In the event an object can not function, e.g. an event listener referencing a non-existing template, that object should report status rather than refusing to install.

# Actual Behavior
EventListeners, TriggerBindings, and TriggerTemplates must be created in a specific order to prevent validation errors.

# Steps to Reproduce the Problem
1. Create an EventListener, TriggerBinding, and TriggerTemplate json files
2. Install the EventListener first

## Alternatively
1. create a helm chart that contains an EventListener, TriggerBinding, and TriggerTemplate
2. attempt to install your chart

# Additional Info
Note in helm installs objects in a standardized order that can not be controlled via naming, sorting, position in file, etc. post install hook seemed fruitful but ultimately don't work. See the below issue for history and motivation:
https://github.com/helm/helm/issues/1228

dibyom:
Fixed in #263 
/close

PR #263, opened 2019-12-10
Stop validating binding/template in Triggers

dibyom:
# Changes

Currently, while validating triggers in an EventListener, we fetch
all the bindings and templates, and validate that they exist. This
means that the bindings and the templates have to be created before
creating EventListeners leading to the fragile object creation order
issues listed in #187

# Submitter Checklist

review them:

# Release Notes

```
We no longer validate the existence of `TriggerBindings` and `TriggerTemplates` when validating a EventListener.

```

### DIFF
--- a/pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	bldr "github.com/tektoncd/triggers/test/builder"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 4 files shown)


dibyom:
Looks like coverage has dropped off a bit :thinking:

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
```diff
@@ -4,26 +4,84 @@ import (
...
+		name: "Valid EventListener No Interceptor",
+		el: bldr.EventListener("name", "namespace",
+			bldr.EventListenerSpec(
+				bldr.EventListenerTrigger("tb", "tt", "v1alpha1"))),
+	}, {
+		name: "Valid EventListener Interceptor Name only",
+		el: bldr.EventListener("name", "namespace",
+			bldr.EventListenerSpec(
+				bldr.EventListenerTrigger("tb", "tt", "v1alpha1",
+					bldr.EventListenerTriggerInterceptor("svc", "", "", "")))),
```
bobcatfish: im confused why it is valid to only specify the name of an interceptor, is that enough to be able to address it?

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
```diff
@@ -4,26 +4,84 @@ import (
...
+				bldr.EventListenerTrigger("tb", "tt", "v1alpha1",
+					bldr.EventListenerTriggerInterceptor("svc", "v1", "Service", "namespace"),
+				),
+				bldr.EventListenerTrigger("tb", "tt", "v1alpha1"))),
+	},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			err := test.el.Validate(context.TODO())
```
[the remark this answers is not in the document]
dibyom: yeap, good call!

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
dibyom: Good catch. Not sure why the validation was done this way but we did return an error later when we actually try to resolve it to a URL.

dibyom: Made this an error test!

bobcatfish: wooooot :D

PR #268, opened 2019-12-11
Fetch resource metadata from unstructured

dibyom:
# Changes
In sink.createResource, we need access to resource metadata such as
name, namespace, apiVersion, and kind from the raw json template. We
were using gjson.GetBytes to do this. However, later in the function
we were also unmarshalling the raw template to
`unstructured.Unstructured` which already has methods for accessing
all of the required metadata.

This commit refactors the createResource function to first umarshall
the raw template to unstructured and then get all of the metadata we
need from that object.  This was also the only place where we were
using gjson so, as a side effect we can now remove our dependency on
the GJson library.

/cc @ImJasonH @wlynch 

# Submitter Checklist

review them:

### DIFF
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -28,7 +28,6 @@ import (
 
 	"github.com/tektoncd/triggers/pkg/interceptors"
 
-	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
 	"golang.org/x/xerrors"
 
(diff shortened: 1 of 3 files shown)


pkg/sink/sink.go
```diff
@@ -163,43 +162,46 @@ func (r Sink) createResources(resources []json.RawMessage, triggerName, eventID
...
-	if err != nil {
-		return err
+	name := data.GetName()
+	if name == "" {
+		name = data.GetGenerateName()
 	}
-
-	resourcename := gjson.GetBytes(rt, "metadata.name")
-	r.Logger.Infof("Generating resource: kind: %+v, name: %s", apiResource, resourcename)
+	r.Logger.Infof("Generating resource: kind: %+v, name: %s", apiResource, data.GetGenerateName())
```
wlynch: r.Logger.Infof("Generating resource: kind: %+v, name: %s", apiResource, name)

dibyom: ahh...great catch :sweat_smile:

dibyom: fixed!

Issue #269, opened 2019-12-11
Use RoundTripper-like interface for interceptors


wlynch:
Tracking bug for modify interceptors to model after [Golang HTTP RoundTrippers](https://godoc.org/net/http#RoundTripper)

Currently:

https://github.com/tektoncd/triggers/blob/3c645a9f50abb3a59cbd5613de8cdc9ee9d9a39f/pkg/interceptors/interceptors.go#L27-L30

Proposed:

```go
type Interceptor interface {
	ExecuteTrigger(req *http.Request, trigger *triggersv1.EventListenerTrigger, eventID string) (*http.Response, error)
}
```

PR #272, opened 2019-12-11
Add support for multiple interceptors.

### LINKED ISSUE #249: Support multiple interceptors
# Expected Behavior

A trigger can specify multiple interceptors. 

# Actual Behavior

A trigger can only have a single interceptor.

# Additional Info

Would be needed if we go with Option 3 from the [Filters design doc](https://docs.google.com/document/d/1c-cGLdRb4LuH2_rUn50cIhHPabAoj4iQvw49L4T2nSU/edit#) (#49)

wlynch:
# Changes

* Adds support for multiple interceptors to be chained together.
* Adds new `interceptors` field to EventListenerTrigger, converts
  now deprecated `interceptor` field to `interceptors` automatically.
* Separates interceptor processing into its own func for easier testing.
* Adds basic sink level unit testing for interceptors.

Fixes #249 

# Submitter Checklist

review them:

# Release Notes

```
interceptor field is now deprecated in favor of interceptors. Old field is backwards compatible for the time being, but will be removed in a future version.
```


pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
```diff
@@ -153,6 +153,31 @@ func Test_EventListenerValidate(t *testing.T) {
 				),
 				bldr.EventListenerTrigger("tb", "tt", "v1alpha1"))),
 		raiseErr: false,
+	}, {
+		name: "Valid EventListener Multiple Interceptor",
```
dibyom: Would it be valid to have multiple interceptor but of different types e.g. Github + Webhook. Going through the current validation code, it seems like that would result in an error? https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_validation.go#L102

It seems like multiple webhooks interceptors should be allowed. Not sure if there is a valid use case for having two Github/Gitlab interceptors though

pkg/apis/triggers/v1alpha1/event_listener_defaults.go
```diff
@@ -10,17 +10,34 @@ func (el *EventListener) SetDefaults(ctx context.Context) {
...
+			upgradeBinding(t)
+			upgradeInterceptor(t)
 		}
 	}
 }
+
+func upgradeBinding(t *EventListenerTrigger) {
+	if t.DeprecatedBinding != nil {
+		if len(t.Bindings) > 0 {
+			// Do nothing since it will be a Validaiton Error.
```
bobcatfish: s/Validaiton/Validation/

(below also!

pkg/sink/sink.go
```diff
@@ -150,6 +112,63 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
+	params, err := template.ResolveParams(rt.TriggerBindings, finalPayload, request.Header, rt.TriggerTemplate.Spec.Params)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	log.Info("params: %+v", params)
+	resources := template.ResolveResources(rt.TriggerTemplate, params)
+	if err := r.createResources(resources, t.Name, eventID); err != nil {
+		log.Error(err)
+		return err
```
bobcatfish: to make it a bit easier to trace through errors, could we wrap the error with a bit of context here? e.g. fmt.Errorf("something useful: %w", err)

pkg/sink/sink.go
```diff
@@ -150,6 +112,63 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
+	log := eventLog.With(zap.String(triggersv1.TriggerLabelKey, t.Name))
+
+	finalPayload, err := r.executeInterceptors(t, request, event, eventID, log)
+	if err != nil {
+		return err
+	}
+
+	rt, err := template.ResolveTrigger(*t,
+		r.TriggersClient.TektonV1alpha1().TriggerBindings(r.EventListenerNamespace).Get,
+		r.TriggersClient.TektonV1alpha1().TriggerTemplates(r.EventListenerNamespace).Get)
```
bobcatfish: Thanks for refactoring these functions out of the big one above!!

if we pass these clients into this funciton, then we can move the new functions into another package and have unit tests for them 🙏

pkg/sink/sink.go
```diff
@@ -150,6 +112,63 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
+		log.Error(err)
+		return err
+	}
+
+	params, err := template.ResolveParams(rt.TriggerBindings, finalPayload, request.Header, rt.TriggerTemplate.Spec.Params)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	log.Info("params: %+v", params)
```
bobcatfish: if we want to keep this log message can we add a bit more info? (also i think we need Infof?)

pkg/sink/sink.go
```diff
@@ -150,6 +112,63 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
+		switch {
+		case i.Webhook != nil:
+			interceptor = webhook.NewInterceptor(i.Webhook, r.HTTPClient, r.EventListenerNamespace, log)
+		case i.Github != nil:
+			interceptor = github.NewInterceptor(i.Github, r.KubeClientSet, r.EventListenerNamespace, log)
+		case i.Gitlab != nil:
+			interceptor = gitlab.NewInterceptor(i.Gitlab, r.KubeClientSet, r.EventListenerNamespace, log)
+		}
+
+		// TODO(https://github.com/tektoncd/triggers/issues/269): Chain interceptors
```
bobcatfish: nit: usually we do just TODO(#269)

pkg/sink/sink_test.go
```diff
@@ -478,3 +481,59 @@ func Test_addLabels(t *testing.T) {
...
+	var got map[string]int
+	if err := json.Unmarshal(resp, &got); err != nil {
+		t.Fatalf("json.Unmarshal: %v", err)
+	}
+	want := map[string]int{"i": 2}
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("-want +got: %s", diff)
+	}
+
+}
```
bobcatfish: nice test!

pkg/sink/sink_test.go
```diff
@@ -478,3 +481,59 @@ func Test_addLabels(t *testing.T) {
...
+
+func (f *fakeHTTPInterceptor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	data := map[string]interface{}{
+		"i": atomic.AddInt32(&f.i, 1),
+	}
+	_ = json.NewEncoder(w).Encode(data)
+}
+
+func TestExecuteInterceptor(t *testing.T) {
```
bobcatfish: this test is super cool: might be worth adding a docstring explaining how it works? e.g. uses 2 interceptors which increment an int so you can be sure they both are executed, and in order, etc.

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
```diff
@@ -153,6 +153,31 @@ func Test_EventListenerValidate(t *testing.T) {
...
+			Spec: v1alpha1.EventListenerSpec{
+				Triggers: []v1alpha1.EventListenerTrigger{{
+					Bindings:              []*v1alpha1.EventListenerBinding{{Name: "tb"}},
+					Template:              v1alpha1.EventListenerTemplate{Name: "tt"},
+					DeprecatedInterceptor: &v1alpha1.EventInterceptor{},
+					Interceptors:          []*v1alpha1.EventInterceptor{{}, {}},
+				}},
+			},
+		},
+		raiseErr: true,
```
bobcatfish: if you feel motivated, i find it much easier to maintain when there is one test for success cases and a separate test for error cases - much easier to see what is expected to be valid and what isnt

pkg/sink/sink_test.go
bobcatfish: are there any other test cases that we might want to include as well? e.g. if one interceptor fails, should the next be invoked?

bobcatfish: nice test!!

bobcatfish: are there other test cases we might want to add as well, e.g. what if one interceptor fails, should the next be run?

bobcatfish: that particular case could maybe also be handled in a unit test for `executeInterceptors` :D

pkg/sink/sink_test.go
bobcatfish: might be useful to include a brief docstring explaining how this test worked - after reading through it it seems super cool ✨ but had to read a bunch before understanding it (that the interceptor increments an int etc)

pkg/apis/triggers/v1alpha1/event_listener_defaults.go
wlynch: Done.

pkg/sink/sink_test.go
wlynch: Done.

wlynch: Done.

pkg/interceptors/webhook/interceptors.go
```diff
@@ -47,6 +47,7 @@ func createOutgoingRequest(ctx context.Context, original *http.Request, url *url
 	r := original.Clone(ctx)
 	r.RequestURI = "" // RequestURI cannot be set in outgoing requests
 	r.URL = url
+	r.Host = url.Host
```
dibyom: :+1: TIL 

>// For incoming requests, the Host header is promoted to the
    // Request.Host field and removed from the Header map.

dibyom:
Ohh, one more thing (really a question) that I just realized :sweat_smile: It seems like we are planning to change what body each interceptor in receives in #269 This seems like a breaking change for a user using multiple interceptors so my question is should we attempt #269 as part of 0.2? 
While this is definite scope creep, it might be worth reducing the number of breaking changes?

wlynch:
> Ohh, one more thing (really a question) that I just realized  It seems like we are planning to change what body each interceptor in receives in #269 This seems like a breaking change for a user using multiple interceptors so my question is should we attempt #269 as part of 0.2?
> While this is definite scope creep, it might be worth reducing the number of breaking changes?

Yeah I think this makes sense. This is likely the expected behavior anyway, I just didn't want to do everything in this one change.

Issue #258, opened 2019-12-09
Values injected into the template should be escaped


bobcatfish:
# Expected Behavior

It should be possible for me to pass along the body of a github pull request event to a TaskRun + Task.

# Actual Behavior

When I'm trying to trigger from a pull request and pass along the description of the pull request to a Task, I run into trouble b/c the description contains characters like `\r` and `\n` and I can't figure out how to escape this in the template properly :S

# Steps to Reproduce the Problem

Example trigger binding I was using:

```yaml
---
apiVersion: tekton.dev/v1alpha1
kind: TriggerBinding
metadata:
  name: bobcatfish-review-pipelinebinding
spec:
  params:
    - name: gitrevision
      value: $(body.pull_request.head.sha)
    - name: gitrepositoryurl
      value: "https://github.com/$(body.repository.full_name)"
    - name: pullrequesturl
      value: $(body.pull_request.html_url)
    - name: body
      value: $(body.pull_request.body) # <-- this is where im getting the description of the pull request
```

Here is an example of the body field of the pull request:

```json
[...]

PR #274, opened 2019-12-12
Refactor functions to not be members of Sink 🚰

bobcatfish:
# Changes

Just like with reconcilers / controllers, adding functions of as members
of the Sink (or controller) object itself has some downsides: it's hard
to tell what the responsibilities of the functions are and how they are
coupled to the Sink, and it's harder to write more focused unit tests.

I am working on #258 in another branch and in that branch I wanted to
make some updates to the unit tests for this logic; it's easier to do if
it's factored out a bit.

There is more factoring we can do here: now that the functions are moved
out of Sink we can make decisions around their interfaces and maybe make
them even more focused and unit testable.

(Though it was a happy surprise to see they were unit tested even though
they were members of Sink!)

# Submitter Checklist

review them:

- [n/a] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)


pkg/resources/create.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/tidwall/sjson"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	discoveryclient "k8s.io/client-go/discovery"
+)
+
+// FindAPIResource returns the APIResource definition using the discovery client c.
+func FindAPIResource(apiVersion, kind string, c discoveryclient.ServerResourcesInterface) (*metav1.APIResource, error) {
+	//resourceList, err := r.DiscoveryClient.ServerResourcesForGroupVersion(apiVersion)
```
dibyom: Remove?

pkg/resources/create.go
```diff
@@ -0,0 +1,121 @@
...
+	resourceList, err := c.ServerResourcesForGroupVersion(apiVersion)
+	if err != nil {
+		return nil, xerrors.Errorf("Error getting kubernetes server resources for apiVersion %s: %s", apiVersion, err)
```
dibyom: While we are refactoring this, could we switch the `xerrors.Errorf` to `fmt.Errorf`? :smile:

pkg/resources/create.go
bobcatfish: :woman_facepalming:

pkg/resources/create.go
bobcatfish: oh yeah good call!

bobcatfish:
Should be ready for another look @dibyom !

bobcatfish:
> Putting these functions into sink was done deliberately in order to reduce passing of parent resources to clients. (e.g. f(param, client, client, client...)). These are things that are never going to change between calls, and makes the API of these functions more complex.

If the function calls depend on attributes of the object they are method of, the attributes are still in the API of the function, it's just hidden in the parent, so I don't think this actually makes the functions more complex, this makes it clear what their API actually is. It makes the function call longer, but imo this is better for maintenance in the long run.

> I suspect what we really want are new types that would decouple these funcs from sink (perhaps also with an interface?)

That could be a good improvement - I this PR could be a step along the way to that? (i.e. i think would make that easier vs harder)

To tell you the truth tho, I'm not a huge fan of putting methods on objects vs passing values into the functions - as far as I can tell the effect is the same and it leads to hidden coupling that's hard to maintain - so if we do create the new types I'd probably rather pass them into the functions. (Maybe one day someone can show me why methods are superior!)

dibyom:
> If the function calls depend on attributes of the object they are method of, the attributes are still in the API of the function, it's just hidden in the parent, so I don't think this actually makes the functions more complex, this makes it clear what their API actually is. It makes the function call longer, but imo this is better for maintenance in the long run.

I really like the bit about complexity just hiding in the parent :100: 

I think an interface would be really cool but definitely something we can do in a later PR

dibyom:
Gonna LGTM this for now but let's discuss refactoring this and adding an interface @wlynch :angel: 

/lgtm

Issue #49, opened 2019-07-30
Add Conditionals to Resource Creation


vtereso:
# Expected Behavior
Users should be able to specify a `Conditional` against event payload values to ensure that resources are generated dependent on the payload data. For context, webhook POST data from GitHub specifies the source organization/repository/branch. Users will want to make `TriggerBindings` that execute against particular branches, etc.

# Actual Behavior
Currently, this is no conditional creation.

# Additional Info
Add `Conditionals` to the `TriggerBindingSpec`. Values should be interpolated from the event into the fields being asserted. Documentation should be included.

vincent-pli:
So, the `Condition` means general `Condition`, not the `Condition` in `tekton pipeline`?

vtereso:
> So, the `Condition` means general `Condition`, not the `Condition` in `tekton pipeline`?
I believe we should be using Tekton pipeline [`Conditions`](https://github.com/tektoncd/pipeline/commit/8f84f033413026c8594cf744ec27e4fe829b8f87) for this.

vincent-pli:
I feel it's a little heavy, `Conditions` in pipeline will launch `container` to do the comparing.
Image that, when received a event, we need create `pod` only for string comparison...

vtereso:
> I feel it's a little heavy, `Conditions` in pipeline will launch `container` to do the comparing.
> Image that, when received a event, we need create `pod` only for string comparison...

@vincent-pli  That's true. I made a point in counter argument to the `Conditionals` initially because I can't really see much besides string comparison to be done, especially here. Tekton is probably better served to used `Conditions` because the work has already been kicked off and the time it takes the `Condition` to complete is minuscule in comparison to the entire workload. Where as here, **many** events are likely to come in and we don't want to get bottle-necked this way. I see both sides because maybe the the payload has some special value that has to be computed upon/set somewhere and only providing some internal string match might not be sufficient.

vtereso:
This work is still blocked, but I think discussion in the issues like this is great. Opinions? @bobcatfish @ncskier @iancoffey

dlorenc:
So to clarify, today we have something like:

```yaml
http:
  headerMatch:
    foo=bar
```

We could extend this to support more advanced filtering with something like:

```yaml
http:
  headerMatch:
    foo=bar
  bodyMatch:
    a.b.c=foo
```

Where a.b.c is a nested json field.

Other questions:
- What type of matching?
- Do we need regex?
- Not equals? Equals?

[...]

PR #285, opened 2019-12-19
Cel interceptor

bigkevmcd:
# Changes

This implements a [CEL](https://github.com/google/cel-go) based interceptor that allows matching on request bodies, and request headers.

# Submitter Checklist

review them:

# Release Notes

```
This adds a new CEL interceptor, able to use expressions to determine whether or not to filter out incoming request bodies.
```


bigkevmcd:
Some thoughts I've had since the initial implementation...

Instead of one `expression` there could be several.

```yaml
expressions:
 - "headers.match(...)"
 - "body.pull_request.head.repo.fork == false"
```

Where each of the expressions would need to return true, this would improve readability (over very long expressions).

Also, there's no way currently to add a custom response header, this could be implemented similar to the `values:` key.

bigkevmcd:
Also, in order to avoid the invisible coupling between the interceptor modifying the request body, and bindings, CEL could be added to the list of expression languages for bindings.

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -105,6 +106,12 @@ type GitlabInterceptor struct {
 	EventTypes []string   `json:"eventTypes,omitempty"`
 }
 
+// CELInterceptor provides a webhook to intercept and pre-process events
+type CELInterceptor struct {
+	Expression string            `json:"expression,omitempty"`
```
dibyom: maybe we can call this `filterExpression` or `filters`? More explicit than the `Value` field which are also technically CEL expressions!

pkg/apis/triggers/v1alpha1/event_listener_validation.go
```diff
@@ -67,7 +67,7 @@ func (s *EventListenerSpec) validate(ctx context.Context, el *EventListener) *ap
 
 func (i *EventInterceptor) validate(ctx context.Context, namespace string) *apis.FieldError {
 	// Validate at least one
-	if i.Webhook == nil && i.Github == nil && i.Gitlab == nil {
+	if i.Webhook == nil && i.Github == nil && i.Gitlab == nil && i.CEL == nil {
```
dibyom: Should we also validate that if CEL has the `expression` field set?

pkg/interceptors/cel/cel.go
```diff
@@ -0,0 +1,183 @@
...
+		KubeClientSet:          k,
+		EventListenerNamespace: ns,
+	}
+}
+
+// ExecuteTrigger is an implementation of the Interceptor interface.
+func (w *Interceptor) ExecuteTrigger(payload []byte, request *http.Request, _ *triggersv1.EventListenerTrigger, _ string) ([]byte, error) {
+	env, err := makeCelEnv()
+	if err != nil {
+		return nil, err
```
dibyom: Could we add some extra context when returning errors in this file? 
e.g. `fmt.Errof("error creating cel environment: %w, err)`

pkg/interceptors/cel/cel_test.go
```diff
@@ -0,0 +1,247 @@
...
+			args: args{
+				payload: []byte(`{}`),
+			},
+			want:    []byte(`{}`),
+			wantErr: false,
+		},
+		{
+			name: "simple header check with non matching header",
+			CEL: &triggersv1.CELInterceptor{
+				Expression: "headers['X-Test'][0] == 'unknown'",
```
dibyom: The `headers` part here is slightly different from `headers` in `TriggerBindings`. 
In Bindings, the keyword is `header` not `headers` -- though `headers` does sound like the more correct term.
Also, you do not have to use array accessors e.g.  `$(headers.foo)` not `$(headers.foo[0])`
Its probably worth being consistent here.

dibyom:
> Instead of one expression there could be several.

Another more verbose option could be the multiple interceptor support that @wlynch is working on

> Also, in order to avoid the invisible coupling between the interceptor modifying the request body, and bindings, CEL could be added to the list of expression languages for bindings

This could be an interesting add on for the future...(JSONPath also supports filtering, maybe that could be a interceptor too :stuck_out_tongue: )

bigkevmcd:
> > Instead of one expression there could be several.
> 
> Another more verbose option could be the multiple interceptor support that @wlynch is working on

This would be slightly different, in that it would be more about readability to avoid having to write long expressions like `filter1 && filter2 && filter3 && filter4`

pkg/interceptors/cel/cel_test.go
bigkevmcd: Just to clarify, you would prefer it to emulate JSON path expressions in CEL?

dibyom: Yes!

pkg/interceptors/cel/cel.go
bigkevmcd: Done

pkg/apis/triggers/v1alpha1/event_listener_validation.go
bigkevmcd: Add validation for the renamed `filter` expression.

bigkevmcd:
The main driver for this is to get rid of an external interceptor that is doing matching, and extracting the SHA and trimming it (amongst other things).

The "values" behaviour emulates this behaviour, so, I'd like to see some discussion about what alternatives could look like.

pkg/interceptors/cel/cel_test.go
bigkevmcd: I'm wondering if CEL is the right expression language to use, maybe we could augment the JSON path expressions?

bigkevmcd: To be clear, the syntax `header['X-Test'][0]` is one possible option.

The header value is a https://golang.org/pkg/net/http/#Header value, but, it's translated in CEL into a map, so, there's a custom match overload, which uses `.Get()` on the value, this is doing more than a simple key-value lookup.

I'm not sure how easy it is to emulate JSON path expressions in CEL without a significant amount of work.

pkg/interceptors/cel/cel_test.go
```diff
@@ -0,0 +1,200 @@
...
+		{
+			name: "simple body check with matching body",
+			CEL: &triggersv1.CELInterceptor{
+				Filter: "body.value == 'testing'",
+			},
+			args: args{
+				payload: []byte(`{"value":"testing"}`),
+			},
+			want:    []byte(`{"value":"testing"}`),
+			wantErr: false,
```
wlynch: Mixing both error and non-error tests makes it a little harder to follow which cases should and should not work. Can we split these cases into 2 separate tests?

pkg/interceptors/cel/cel_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+// Allow configuration via a config map
+
+func TestInterceptor_ExecuteTrigger(t *testing.T) {
+	type args struct {
+		payload []byte
+	}
+	tests := []struct {
+		name    string
+		CEL     *triggersv1.CELInterceptor
```
wlynch: It looks like the only thing that changes here is the filter string. The test case struct would be simpler if we just use `filter string` here and create the CELInterceptor in the test.

pkg/interceptors/cel/cel_test.go
```diff
@@ -0,0 +1,200 @@
...
+		args    args
```
wlynch: Similarly, `args` is just a wrapper type for `[]byte`. Can we use the type directly here?

pkg/sink/sink.go
```diff
@@ -92,6 +93,8 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 				interceptor = github.NewInterceptor(t.Interceptor.Github, r.KubeClientSet, r.EventListenerNamespace, log)
 			case t.Interceptor.Gitlab != nil:
 				interceptor = gitlab.NewInterceptor(t.Interceptor.Gitlab, r.KubeClientSet, r.EventListenerNamespace, log)
+			case t.Interceptor.CEL != nil:
```
wlynch: Can we add a test at the interceptor level to make sure things work as expected end to end?

pkg/interceptors/cel/cel_test.go
dibyom: By "emulating", I was just referring to making it easy to access single valued headers i.e header['X-Test'][0] vs header['X-Test']. From reading @wlynch 's comment it seems like that might already be the case...
We definitely should not try to do anything more complicated than that 😅

pkg/interceptors/cel/cel.go
```diff
@@ -0,0 +1,150 @@
...
+
+	out, _, err := prg.Eval(data)
+	return out, err
+}
+
+func embeddedFunctions() cel.ProgramOption {
+	return cel.Functions(
+		&functions.Overload{
+			Operator: "match",
+			Function: matchHeader},
```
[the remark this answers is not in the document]
bigkevmcd: That's not quite what it does, the value of `header` is https://golang.org/pkg/net/http/#Header and the `match` delegates to the `Get` method, which does a bit more than accessing the header for single-value headers.

pkg/interceptors/cel/cel.go
```diff
@@ -0,0 +1,150 @@
...
+
+	}
+
+	return payload, nil
+}
+
+func evaluate(expr string, env cel.Env, data map[string]interface{}) (ref.Val, error) {
+	parsed, issues := env.Parse(expr)
+	if issues != nil && issues.Err() != nil {
+		return nil, issues.Err()
```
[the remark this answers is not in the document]
bigkevmcd: The error message from CEL is pretty good, the test I added only really tests a subset of the message.

```
failed to evaluate expression 'header['X-Test': ERROR: <input>:1:8: Syntax error: token recognition error at: ''X-Test'
 | header['X-Test
 | .......^
ERROR: <input>:1:15: Syntax error: mismatched input '<EOF>' expecting {'[', '{', '(', '.', '-', '!', 'true', 'false', 'null', NUM_FLOAT, NUM_INT, NUM_UINT, STRING, BYTES, IDENTIFIER}
 | header['X-Test
 | ..............^
```

pkg/apis/triggers/v1alpha1/event_listener_validation.go
```diff
@@ -67,7 +67,7 @@ func (s *EventListenerSpec) validate(ctx context.Context, el *EventListener) *ap
...
 		return apis.ErrMissingField("interceptor")
 	}
```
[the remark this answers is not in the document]
dibyom: I think this still needs to be added

dibyom:
> We should make sure to add a reference to the issue (#49) in the PR body / commit message.

Looks like we need to add this as well!

docs/eventlisteners.md
```diff
@@ -194,3 +194,32 @@ spec:
       template:
         name: pipeline-template
 ```
+
+### CEL Interceptors
+
+CEL interceptors parse expressions to filter requests based on JSON bodies and request headers, using the
+[CEL](https://github.com/google/cel-go) expression language.
+
+Supported features include case-insensitive matching on request headers.
```
wlynch: Let's document the custom methods we add similar to https://github.com/google/cel-spec/blob/master/doc/langdef.md#list-of-standard-definitions

bigkevmcd: Is this ok for a start?

https://github.com/tektoncd/triggers/blob/5d0379587187bef6ced05c6469b10d272d28009f/docs/cel_expressions.md

wlynch: Looks good! :smile: Thanks!

PR #329, opened 2020-01-09
Add support for adding additional fields to the body.

bigkevmcd:
# Changes

This adds an `overlays` field to the CELInterceptor, which can add extra fields to a response body.

In addition, it adds a CEL extension for truncating strings.

# Submitter Checklist

review them:

# Release Notes

```
This adds new functionality to the CEL interceptor, adding the ability to add new derived fields to the request body that can be picked up in template bindings.
```


pkg/interceptors/cel/cel_test.go
```diff
@@ -202,3 +231,61 @@ func TestFilterEvaluation(t *testing.T) {
 		})
 	}
 }
+
+func TestExpressionEvaluation(t *testing.T) {
```
wlynch: I like this focused method for testing expressions! Would we be able to also include another tests for errors - e.g. mismatched types, invalid expressions, etc?

examples/eventlisteners/cel-eventlistener-interceptor.yaml
```diff
@@ -8,7 +8,10 @@ spec:
     - name: cel-trig
       interceptor:
         cel:
-          filter: "headers.match('X-GitHub-Event', 'push')"
+          filter: "headers.match('X-GitHub-Event', 'pull_request')"
+          overlays:
+          - key: extensions.truncated_sha
+            expression: "truncate(body.pull_request.head.sha, 7)"
```
wlynch: Not a blocker, but one thing I'm worried about here is that there is now multiple ways to do value selection in Triggers: CEL, gjson, and jsonPath. I don't know if we can necessarily do better in this case, but we'll probably want to

bigkevmcd: This is a tricky one, CEL provides more than just extracting values, via the extension functions, lots of regularly used functionality can be made available (e.g. base64).

I have no particular attachment to CEL, but the functionality it provides isn't available in the alternatives?

pkg/interceptors/cel/cel_test.go
```diff
@@ -156,13 +186,14 @@ func TestInterceptor_ExecuteTrigger(t *testing.T) {
...
+	testSHA := "ec26c3e57ca3a959ca5aad62de7213c562f8c821"
 	jsonMap := map[string]interface{}{
 		"value": "testing",
-		"sha":   "ec26c3e57ca3a959ca5aad62de7213c562f8c821",
+		"sha":   testSHA,
 	}
 	header := http.Header{}
-	evalEnv := map[string]interface{}{"body": jsonMap, "header": header}
+	evalEnv := map[string]interface{}{"body": jsonMap, "headers": header}
```
dibyom: Is it `header` or `headers`? In TriggerBindings we use `header` I think we should be consistent!

bigkevmcd:
Anything else to do for this?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -110,7 +110,14 @@ type GitLabInterceptor struct {
 
 // CELInterceptor provides a webhook to intercept and pre-process events
 type CELInterceptor struct {
-	Filter string `json:"filter,omitempty"`
+	Filter   string       `json:"filter,omitempty"`
+	Overlays []CELOverlay `json:"overlays,omitempty"`
```
dibyom: I think we need to fix our validation below so that it accepts either filters or overalys. Right now I need to specify a `filter` even if only want to use `overlays`
https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_validation.go#L134

bigkevmcd: I hadn't thought about that use case before, and yes, it definitely sounds useful, I'll sort that out.

docs/eventlisteners.md
```diff
@@ -257,12 +256,16 @@ spec:
     - name: cel-trig
       interceptors:
         - cel:
-            filter: "header.match('X-GitHub-Event', 'push')"
+            filter: "headers.match('X-GitHub-Event', 'pull_request')"
+            overlays:
+            - key: extensions.truncated_sha
```
dibyom: Would be nice to have an example where the expression returns something other than a string e.g. a JSON object itself

pkg/interceptors/cel/cel.go
```diff
@@ -83,10 +84,22 @@ func (w *Interceptor) ExecuteTrigger(request *http.Request) (*http.Response, err
 
 	}
 
+	for _, u := range w.CEL.Overlays {
+		val, err := evaluate(u.Expression, env, evalContext)
+		if err != nil {
+			return nil, fmt.Errorf("failed to evaluate overlay expression '%s': %w", w.CEL.Filter, err)
```
dibyom: Should be `w.CEL.Overlays` or something like that?

bigkevmcd: Good catch, thanks!

docs/eventlisteners.md
bigkevmcd: From a subtree of the evaluated body?

pkg/apis/triggers/v1alpha1/event_listener_types.go
bigkevmcd: Should be doable now, and I've updated the docs with an example and explanation.

docs/eventlisteners.md
dibyom: well, Billy and I were playing around with this yesterday and we thought we needed to append a JSON body but we got around with just a string in the end. This is more of a feature so maybe we just do strings for now and hold of on anything more complex until we have a good use case for it!

bobcatfish:
is it possible to add something to https://github.com/tektoncd/triggers/blob/master/docs/eventlisteners.md#CEL-Interceptors to explain what `overlays` are and how they work?

bigkevmcd:
@bobcatfish on it

bigkevmcd:
@bobcatfish how does https://github.com/tektoncd/triggers/compare/master...bigkevmcd:improve-documentation-overlays?expand=1 look?

Want me to split it out into a separate file?

PR #326, opened 2020-01-08
Use RoundTripper-like interface for interceptors.

wlynch:
# Changes

This simplifies the interceptor interface to look like
http.RoundTripper, having interceptors take in an http.Request and
returning an http.Response.

The only substantial behavior change this makes is gives webhook interceptors
the ability to respond back with HTTP headers.

Fixes #269.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/triggers/blob/master/CONTRIBUTING.md) for more details._

PR #334, opened 2020-01-10
Daisy-chain interceptor requests.

### LINKED ISSUE #325: Daisy-chain Interceptors
# Expected Behavior

Interceptors should daisy chain into each other, so that they can modify payloads that other interceptors can consume.

e.g. for 2 interceptors A and B, the original request should go to A, the response of A should be the request of B, and the response of B should be used for resource resolution.

# Actual Behavior

[The original request is sent to each interceptor.](https://github.com/tektoncd/triggers/blob/256c7e4689fc6af4e70686f82c7b593d79db7c7d/pkg/sink/sink.go#L172).

# Additional Info

This is dependent on #269.

/assign wlynch
/kind feature

wlynch:
# Changes

This changes the behavior of interceptors to feed the body/header of an
interceptor response in as the request for the next interceptor. This
allows interceptors to be layered to modify the behavior of future
interceptors.

Fixes #325

# Submitter Checklist

review them:

# Release Notes

```
Interceptors will now feed body/header responses as the request for the next interceptor.
```

### DIFF
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -153,11 +153,12 @@ func (r Sink) processTrigger(t *triggersv1.EventListenerTrigger, request *http.R
 	return nil
 }
 
-func (r Sink) executeInterceptors(t *triggersv1.EventListenerTrigger, request *http.Request, event []byte, eventID string, log *zap.SugaredLogger) ([]byte, http.Header, error) {
+func (r Sink) executeInterceptors(t *triggersv1.EventListenerTrigger, in *http.Request, event []byte, eventID string, log *zap.SugaredLogger) ([]byte, http.Header, error) {
 	if len(t.Interceptors) == 0 {
-		return event, request.Header, nil
+		return event, in.Header, nil
 	}
 
+	request := in
 	var resp *http.Response
 	for _, i := range t.Interceptors {
 		var interceptor interceptors.Interceptor
@@ -174,12 +175,17 @@ func (r Sink) executeInterceptors(t *triggersv1.EventListenerTrigger, request *h
 			return nil, nil, fmt.Errorf("unknown interceptor type: %v", i)
 		}
 
-		// TODO(#325): Chain interceptors together using HTTP request/response.
 		var err error
 		resp, err = interceptor.ExecuteTrigger(request)
 		if err != nil {
 			log.Error(err)
 			return nil, nil, err
+		}
+		// Set the next request to be the output of the last response to enable
+		// request chaining.
+		request = &http.Request{
+			Header: resp.Header,
+			Body:   resp.Body,
 		}
 	}
 	payload, err := ioutil.ReadAll(resp.Body)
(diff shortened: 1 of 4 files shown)


pkg/sink/sink.go
```diff
@@ -153,11 +153,12 @@ func (r Sink) processTrigger(t *triggersv1.EventListenerTrigger, request *http.R
 	return nil
 }
 
-func (r Sink) executeInterceptors(t *triggersv1.EventListenerTrigger, request *http.Request, event []byte, eventID string, log *zap.SugaredLogger) ([]byte, http.Header, error) {
+func (r Sink) executeInterceptors(t *triggersv1.EventListenerTrigger, in *http.Request, event []byte, eventID string, log *zap.SugaredLogger) ([]byte, http.Header, error) {
```
ncskier: I have a few comments about this line (not about this PR specifically)
1. I noticed that we aren't using anything from the `t *triggersv1.EventListenerTrigger` except its Interceptors (`t.Interceptors`), so I think that the function parameter should just be `interceptors []*EventInterceptor` instead of the entire `t *triggersv1.EventListenerTrigger`.
2. I don't think that the `eventID` is used in the `executeInterceptors()` function, so we can remove it as a parameter.
3. The `event []byte` is already in the `in *http.Request` parameter, so it's duplicate information in the `executeInterceptors()` function. I think that we could clean up the function declaration a little bit by removing the `event []byte` parameter.

Cleaning up these parameters will make is cleaner and easier to test the function, because we won't have to create as many fake parameters to feed into our tests.

wlynch: Agreed! I tackled most of this in #326, but looks like I forgot to bubble this up to the event listener. I can tackle this in another change.

pkg/sink/sink_test.go
```diff
@@ -232,15 +233,24 @@ func TestHandleEvent(t *testing.T) {
...
 
 // sequentialInterceptor is a HTTP server that will return sequential responses.
+// The response body will always return with the next value set, whereas the
+// headers will append new values in the sequence.
 type sequentialInterceptor struct {
 	i int32
 }
 
 func (f *sequentialInterceptor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	i := atomic.AddInt32(&f.i, 1)
```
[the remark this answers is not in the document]
wlynch: Good idea! Done.

pkg/sink/sink.go
dibyom: Refactoring in a different PR sounds good to me. Any other concerns @ncskier ?

ncskier: Nope, as long as it's handled in another PR, should be good; thanks! 👍

dibyom:
The linter strikes again :stuck_out_tongue_closed_eyes: 

> I0113 18:17:29.721] pkg/sink/sink_test.go:251:2: assignOp: replace `data["i"] = data["i"] + 1` with `data["i"]++` (gocritic)
> I0113 18:17:29.722] 	data["i"] = data["i"] + 1

bobcatfish:
Do we have an example that demonstrates this? (assuming we have enough out of the box interceptors I guess :sweat_smile: )

PR #368, opened 2020-01-21
DNM Add unit tests to reproduce issues/365

afrittoli:
Unit test on jsonpath_test passes, the one on event_test fails.

This is not meant to merged, it's only a reproducer for the issue


afrittoli:
This is now integrated in the fix PR

PR #369, opened 2020-01-22
fix sink crashloop when upgrading from 0.1 to 0.2 

### LINKED ISSUE #366: Upgrade to v0.2.0 requires recreating event listeners
# Expected Behavior
Upgrading to triggers v0.2.0. Event listeners should be up and running.

# Actual Behavior
Event listeners are in crashloop, with the following error:
```
2020/01/20 14:18:55 Error loading logging configuration: lstat /etc/config-logging: no such file or directory

```

1. Run triggers v0.1.0
2. Create event listeners
3. Upgrade to triggers v0.2.0

# Additional Info
To fix this issue:
- [ ] Update Release notes to say that the Params field has been removed and explicitly mention that Params field will not work at all and users will have to create `TriggerBindings` with static values instead.
- [ ]  Add a `DeperecatedParams` field to fix the sink `CrashLoopBackoff`

dibyom:
# Changes

Fixes #366 

The crashloop is due to 2 reasons:
1. We deprecated and removed the `params` field from the EventListener. So, when the reconciler tries to  unmarshal an old EventListener, it fails since it cannot parse the deleted `params` field.

2. The new event listener sink image expects `/etc/config-logging` to be mounted. However the reconciler code does not add the `volumeMount` for any existing sink deployments. So, the new pods go into a crashloop

To fix this, this commit does the following:
1. Add back the `params` field. Also, update the mutating webhook to delete any set `params` field in existing eventlisteners.
2. Add volumeMount for existing deployments if not present.

# Submitter Checklist

review them:

# Release Notes

```
fixes a bug where upgrading an existing event listener from Triggers 0.1 to 0.2 causes a crashloop
```

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -17,7 +17,6 @@ limitations under the License.
 package main
 
 import (
-	"context"
 	"log"
 
 	"github.com/tektoncd/pipeline/pkg/system"
@@ -81,9 +80,7 @@ func main() {
 	}
 
 	// Decorate contexts with the current state of the config.
-	ctxFunc := func(ctx context.Context) context.Context {
-		return ctx
-	}
+	ctxFunc := v1alpha1.WithUpgradeViaDefaulting
 
 	controller, err := webhook.New(kubeClient, options, admissionControllers, logger, ctxFunc)
 	if err != nil {
--- a/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
+++ b/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
@@ -354,6 +354,14 @@ func (c *Reconciler) reconcileDeployment(el *v1alpha1.EventListener) error {
 				existingDeployment.Spec.Template.Spec.Containers[0].Command = nil
 				updated = true
 			}
+			if len(existingDeployment.Spec.Template.Spec.Containers[0].VolumeMounts) == 0 {
+				existingDeployment.Spec.Template.Spec.Containers[0].VolumeMounts = container.VolumeMounts
+				updated = true
+			}
+			if !reflect.DeepEqual(existingDeployment.Spec.Template.Spec.Volumes, deployment.Spec.Template.Spec.Volumes) {
+				existingDeployment.Spec.Template.Spec.Volumes = deployment.Spec.Template.Spec.Volumes
+				updated = true
+			}
 		}
 		if updated {
 			if _, err := c.KubeClientSet.AppsV1().Deployments(el.Namespace).Update(existingDeployment); err != nil {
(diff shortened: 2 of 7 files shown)


ncskier:
I just tested your branch. I can verify that I no longer see the EventListener running into a CrashLoop when I upgrade from v0.1.0 to this branch.

ncskier:
When I updated my Triggers version, my EventListener's fields did not update (`binding` -> `bindings` and `params` getting deleted). Also, in the [EventListener reconciler file](https://github.com/tektoncd/triggers/blob/master/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go), I don't see `Update()` called on the EventListener, so where do we update it with the changes from `event_listener_defaults.go`?

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -354,6 +354,10 @@ func (c *Reconciler) reconcileDeployment(el *v1alpha1.EventListener) error {
 				existingDeployment.Spec.Template.Spec.Containers[0].Command = nil
 				updated = true
 			}
+			if len(existingDeployment.Spec.Template.Spec.Containers[0].VolumeMounts) == 0 {
+				existingDeployment.Spec.Template.Spec.Containers[0].VolumeMounts = container.VolumeMounts
+				updated = true
+			}
```
ncskier: Do we also need to check if the `existingDeployment.Spec.Template.Spec.Volumes` has been updated?

ncskier: I think we should add this below line 360:
```go
if !reflect.DeepEqual(existingDeployment.Spec.Template.Spec.Volumes, deployment.Spec.Template.Spec.Volumes) {
	existingDeployment.Spec.Template.Spec.Volumes =  deployment.Spec.Template.Spec.Volumes
	updated = true
}
```

dibyom: Added VolumeMount but not the actual Volume so the deployment is invalid :face_with_head_bandage:

dibyom:
Ok, so the crashloop should be fixed though we still need to address @ncskier 's comment https://github.com/tektoncd/triggers/pull/369#issuecomment-577347870

dibyom:
quick and dirty script I used to test this: https://gist.github.com/dibyom/93b8455ee18e496d408ec4ef51d2b476

ncskier:
@googlebot I consent.

ncskier:
@googlebot I consent

cmd/webhook/main.go
```diff
@@ -82,7 +82,7 @@ func main() {
 
 	// Decorate contexts with the current state of the config.
 	ctxFunc := func(ctx context.Context) context.Context {
-		return ctx
+		return v1alpha1.WithUpgradeViaDefaulting(ctx)
 	}
```
ncskier: @dibyom I think `golangci` is complaining about this. It should be
```go
ctxFunc := v1alpha1.WithUpgradeViaDefaulting
```

PR #376, opened 2020-01-22
Allow brackets inside Tekton expressions

### LINKED ISSUE #365: JSONPath filter expression is invalid
# Expected Behavior
Given the expression: `$(body.taskRun.spec.outputs.resources[?(@.name == 'bucket')].resourceRef.name)`, the expression should be valid

# Actual Behavior
User gets a "unterminated expression" error

```
{"level":"error","logger":"eventlistener","caller":"sink/sink.go:144","msg":"failed to ApplyEventValuesToParams: &{%!w(string=failed to replace JSONPath value for param bucket: {string $(body.taskRun.spec.outputs.resources[?(@.name == 'bucket')].resourceRef.name) []}: &{%!w(string=unterminated filter)})}","knative.dev/controller":"eventlistener","/triggers-eventid":"958bw","/trigger":"log-collection","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.processTrigger\n\t/workspace/go/src/github.com/tektoncd/triggers/pkg/sink/sink.go:144\ngithub.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent.func1\n\t/workspace/go/src/github.com/tektoncd/triggers/pkg/sink/sink.go:93"}
```

dibyom:
We were previously using a regex to extract expressions within
TriggerBindings. The regex only extracted between the initial `$(`
and the first occurrence of the `)`. This meant that some valid
JSONPath expressions that contained brackets (e.g. filters) were
incorrectly extracted.

This commit fixes the issue by splitting the string on `$(` and then
extracting until the *first unbalanced* occurrence of `)`.

Fixes #365 

# Changes

# Submitter Checklist

review them:

# Release Notes

```
TriggerBinding values now correctly detect embedded brackets

```

### DIFF
--- a/pkg/template/jsonpath.go
+++ b/pkg/template/jsonpath.go
@@ -146,3 +146,37 @@ func relaxedJSONPathExpression(pathExpression string) (string, error) {
 func isTektonExpr(expr string) bool {
 	return tektonVar.MatchString(expr)
 }
+
+// findTektonExpressions searches for and returns a slice of
+// all substrings that are wrapped in $()
+func findTektonExpressions(in string) []string {
+	results := []string{}
+
+	// No expressions to return
+	if !strings.Contains(in, "$(") {
+		return results
+	}
+	// Splits string on $( to find potential Tekton expressions
+	maybeExpressions := strings.Split(in, "$(")
+	for _, e := range maybeExpressions[1:] { // Split always returns at least one element
+		// Iterate until we find the first unbalanced )
+		numOpenBrackets := 0
+		for i, ch := range e {
+			switch ch {
+			case '(':
+				numOpenBrackets++
+			case ')':
+				numOpenBrackets--
+				if numOpenBrackets < 0 {
+					results = append(results, fmt.Sprintf("$(%s)", e[:i]))
+				}
+			default:
+				continue
+			}
+			if numOpenBrackets < 0 {
+				break
+			}
+		}
+	}
+	return results
+}
--- a/pkg/template/jsonpath_test.go
+++ b/pkg/template/jsonpath_test.go
@@ -83,6 +83,11 @@ func TestParseJSONPath(t *testing.T) {
 		in:   `{"body":["", null, "thing"]}`,
 		expr: "$(body[:2])",
 		want: `["", null]`,
+	}, {
+		name: "Array filters",
+		in:   `{"body":{"child":[{"a": "b", "w": "1"}, {"a": "c", "w": "2"}, {"a": "d", "w": "3"}]}}`,
+		expr: "$(body.child[?(@.a == 'd')].w)",
+		want: "3",
 	}}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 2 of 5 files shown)


dibyom:
> ❌ CLAs are signed, but unable to verify author consent


Manually verified with @afrittoli that he does indeed consent to Co-Authorship :stuck_out_tongue_closed_eyes:

dibyom:
Looks like I can't manually update the cla label anymore :disappointed: 

@afrittoli you might have to comment `@googlebot I consent`

/hold cancel

afrittoli:
@googlebot I consent

pkg/template/jsonpath_test.go
```diff
@@ -213,3 +218,48 @@ func TestRelaxedJSONPathExpression_Error(t *testing.T) {
 		})
 	}
 }
+
+func TestFindTektonExpressions(t *testing.T) {
+	tcs := []struct {
+		in   string
+		want []string
+	}{{
+		in:   "$(body.blah)",
```
wlynch: Let's add a `""` case for sanity checking.

pkg/template/jsonpath.go
```diff
@@ -146,3 +146,37 @@ func relaxedJSONPathExpression(pathExpression string) (string, error) {
...
+	for _, e := range maybeExpressions[1:] { // Split always returns at least one element
+		// Iterate until we find the first unbalanced )
+		numOpenBrackets := 0
+		for i, ch := range e {
+			switch ch {
+			case '(':
+				numOpenBrackets++
+			case ')':
+				numOpenBrackets--
+				if numOpenBrackets < 0 {
```
wlynch: Would something like `$()))` cause a problem here? Let's add a test case! :D

pkg/template/jsonpath.go
```diff
@@ -146,3 +146,37 @@ func relaxedJSONPathExpression(pathExpression string) (string, error) {
...
+// all substrings that are wrapped in $()
+func findTektonExpressions(in string) []string {
+	results := []string{}
+
+	// No expressions to return
+	if !strings.Contains(in, "$(") {
+		return results
+	}
+	// Splits string on $( to find potential Tekton expressions
+	maybeExpressions := strings.Split(in, "$(")
```
wlynch: How do we want to handle expressions like `$($())`? Let's also add a test to verify the behavior.

dibyom: yeah, so this is the interesting one :smile: 

currently, we'd return $() i.e. the inner expression

pkg/template/jsonpath_test.go
dibyom: done!

pkg/template/jsonpath.go
dibyom: done!

pkg/template/jsonpath.go
dibyom: `$` is usually the root element in JSONPath though in the kubectl jsonpath template syntax that we are using does not mention $ at all

https://goessner.net/articles/JsonPath/

https://kubernetes.io/docs/reference/kubectl/jsonpath/

dibyom:
```
kubectl get pod -o=jsonpath='{$(.items[0])}'

error: error parsing jsonpath {$(.items[0])}, unrecognized character in action: U+0028 '('
```

dibyom: so, IMO we can document that if you have embed `$()`...we'll just use the innermost one e.g. `"$($($(blahblah)))"` will return `$(blahblah)`

what do you think @wlynch @afrittoli ?

wlynch: What happens if you have `$($($(foo))-bar)`?

It might just be easier to just disallow nested `$()` and make it an error.

dibyom: the innermost rule applies and we get `$(foo)`

dibyom: Returning a error is definitely an option though documented/defined behavior for nested `$(` is also an option. Don't feel too strongly about either

wlynch: Ignoring the `-bar` bit feels unexpected to me. I'm leaning towards making it an error.

dibyom: The bar is only ignored as part of the jsonpath expression...when we do the repalcements -> it still shows up as part of the string i,e, if $(foo) evals to "abc" the value of the binding is `$($(abc)-bar)`

wlynch: Ah! Okay, documenting that as a first step sounds good to me then! We can always revisit this later.

PR #378, opened 2020-01-23
fix sink crashloop when upgrading from 0.1 to 0.2

### LINKED ISSUE #366: Upgrade to v0.2.0 requires recreating event listeners

dibyom:
The crashloop is due to 2 reasons:
1. We deprecated and removed the `params` field from the EventListener. So, when the reconciler tries to  unmarshal an old EventListener, it fails since it cannot parse the deleted `params` field.

2. The new event listener sink image expects `/etc/config-logging` to be mounted. However the reconciler code does not add the `volumeMount` for any existing sink deployments. So, the new pods go into a crashloop

To fix this, this commit does the following:
1. Add back the `params` field. Also, update the mutating webhook to delete any set `params` field in existing eventlisteners.
2. Add volumeMount for existing deployments if not present.

Fixes #366

Co-Authored-by: Brandon Walker <bcwalker@ibm.com>

# Changes

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/pkg/apis/triggers/v1alpha1/event_listener_defaults.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_defaults.go
@@ -12,6 +12,7 @@ func (el *EventListener) SetDefaults(ctx context.Context) {
 			t := &el.Spec.Triggers[i]
 			upgradeBinding(t)
 			upgradeInterceptor(t)
+			removeParams(t)
 		}
 	}
 }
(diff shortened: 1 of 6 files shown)


ncskier:
@googlebot I consent

PR #407, opened 2020-02-03
Update CEL docs

ncskier:
# Changes

- Replace `match` with `matches`
(https://github.com/google/cel-spec/blob/master/doc/langdef.md).
- Add CEL to the list of Interceptor types.

# Submitter Checklist

review them:

### DIFF
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -7,7 +7,6 @@ resources will be created (or at least attempted) with. The service account must
 have the following role bound.
 
 <!-- FILE: examples/role-resources/role.yaml -->
-
 ```YAML
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
@@ -71,9 +70,10 @@ modify the behavior or payload of triggers.
 
 Event Interceptors can take several different forms today:
 
-- Webhook Interceptors
-- GitHub Interceptors
-- GitLab Interceptors
+- [Webhook Interceptors](#Webhook-Interceptors)
+- [GitHub Interceptors](#GitHub-Interceptors)
+- [GitLab Interceptors](#GitLab-Interceptors)
+- [CEL Interceptors](#CEL-Interceptors)
 
 ### Webhook Interceptors
 
@@ -114,7 +114,6 @@ To be an Event Interceptor, a Kubernetes object should:
   the body, it can simply return the body that it received.
 
 <!-- FILE: examples/eventlisteners/eventlistener-interceptor.yaml -->
-
 ```YAML
 ---
 apiVersion: tekton.dev/v1alpha1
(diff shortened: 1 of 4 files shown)


docs/eventlisteners.md
```diff
@@ -260,7 +256,7 @@ spec:
     - name: cel-trig
       interceptors:
         - cel:
-            filter: "headers.match('X-GitHub-Event', 'pull_request')"
+            filter: "headers.matches('X-GitHub-Event', 'pull_request')"
```
dibyom: I realised this after I hit /approve...
I think we added a custom `match` function : https://github.com/tektoncd/triggers/blob/master/pkg/interceptors/cel/cel.go

ncskier: I'm not very CEL savvy. Why did we create a custom `match` function?

dibyom: Some discussion in the original PR here: https://github.com/tektoncd/triggers/pull/285#discussion_r360061744

dibyom: Perhaps worth clarifying this in the docs? The built in `matches` vs `match` can get confusing...
@bigkevmcd

ncskier: If I understand correctly, it's used specifically to match agains a header with one value? Doesn't `headers.X-GitHub-Event[0].matches('pull_request')` have the some functionality (I'm probably missing something 😅 ).

Assuming I'm missing something, could we rename the built-in CEL function from `match` to `headerMatch` or `singleHeaderMatch` or something to be more explicit? I assumed it was a typo for `matches`.

dibyom: Fair point...what do you think @bigkevmcd ?

bigkevmcd: A couple of possibilities come to mind, I could add an overlay function  for requests something like `canonicalHeader` which would lookup the header using the Go http.Header `.Get` and return it.

This _might_ fall foul of the change from String to []byte for functions, I'd need to check.

Either way, yes, happy to rename `match` to `headerMatch`.

bigkevmcd:
Oh, and this PR was incorrect, "matches" != "match", match is the "use the Go http.Header method to compare the header name in a canonical way", "matches" is a CEL originated function.

(There is an issue in the doc, it should be `header.match`, I keep forgetting that it's singular `header` for accessing the headers).

docs/eventlisteners.md
bigkevmcd: Actually, looking at this again, the reason I called it match is that the expression reads as `header.match`, duplicating `match` again with `header.matchHeader` feels wrong.

ncskier: But we don't want people doing `body.match`, right? Does the function match the first element in the array? It could be `header.matchFirst` or `header.matchFirstElement`?

Issue #367, opened 2020-01-21
CEL expressions in TriggerBindings


bigkevmcd:
Being able apply functions to the hook bodies in the trigger bindings might be useful.

Perhaps prefixing the expressions?

I think this can probably mostly be done with the CEL interceptor (with the the overlays functionality) so this might be an easy way around this.

```yaml
apiVersion: tekton.dev/v1alpha1
kind: TriggerBinding
metadata:
  name: my-trigger-binding
spec:
  params:
  - name: shortsha
    value: cel:(truncate(body.pull_request.head.ref, 8))
```

ncskier:
Might need some discussion about replacing JSONPath or supporting both

wlynch:
Previous discussion: https://github.com/tektoncd/triggers/issues/178#issuecomment-549576420

dibyom:
At the Feb 11 WG, we decided we can push this after the 0.3 milestone

bitsofinfo:
would love this!

Fantaztig:
@dibyom I was just trying to find out why this isn't supported yet, do you maybe have some discussion that you could link in this issue? 😄 
Since it's been frozen for almost 4 years now, it seems like it's not as easy as it seems!

bigkevmcd:
@Fantaztig Do you have a specific use-case in mind?

I suspect most of this can be done with a CEL ["overlays"](https://tekton.dev/docs/triggers/interceptors/#cel-interceptors)?

Fantaztig:
It's not really a big issue, more a cosmetic thing.
Currently we extract most of the webhook data we need inside the TriggerBinding e.g. via `$(body.changes[0].ref.displayId)`.
Some info needs some light processing though, e.g. `body.pullRequest.fromRef.displayId.split('/')[1]`, those we have inside the EventListener right now as overlays.

What I dislike a bit about that is that 
- a) some of the overlays in the EventListener are quite redundant as they are always the same for the same TriggerBinding.
- b) the TriggerBinding has a mix of both expressions accessing data from the payload and those simply remapping the info returned by the EventListener which means I need to check two different files to know what data is actually loaded into which param.


I'm open to alternative approaches though, is the best practice to have all the value extraction inside the EventListener despite the redundancy? Or is there another way? 😃

PR #411, opened 2020-02-04
Add support for splitting strings to CEL expressions.

bigkevmcd:
# Changes

This adds an additional CEL function for splitting strings on a separator.

e.g. body.ref.split('/') would split the ref on '/' and return a list of strings.

# Submitter Checklist

review them:

# Release Notes

```
This adds a CEL expression function for splitting strings.

e.g. to extract the tag or branch from a hook event in an overlay:

            overlays:
            - key: intercepted.branch_name
              expression: "body.ref.split('/')[2]"
```

### DIFF
--- a/docs/cel_expressions.md
+++ b/docs/cel_expressions.md
@@ -95,5 +95,18 @@ interceptor.
      <pre>truncate(body.commit.sha, 5)</pre>
     </td>
   </tr>
-
+  <tr>
+    <th>
+      split
+    </th>
+    <td>
+      (string, string) -> string(dyn)
+    </td>
+    <td>
+      Splits a string on the provided separator value.
+    </td>
+    <td>
+     <pre>split(body.ref, '/')</pre>
+    </td>
+  </tr>
 </table>
--- a/pkg/interceptors/cel/cel.go
+++ b/pkg/interceptors/cel/cel.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"reflect"
+	"strings"
 
 	structpb "github.com/golang/protobuf/ptypes/struct"
 	"github.com/google/cel-go/cel"
@@ -151,11 +152,15 @@ func embeddedFunctions() cel.ProgramOption {
 		&functions.Overload{
 			Operator: "truncate",
 			Binary:   truncateString},
+		&functions.Overload{
+			Operator: "split",
+			Binary:   splitString},
 	)
 
 }
 func makeCelEnv() (cel.Env, error) {
 	mapStrDyn := decls.NewMapType(decls.String, decls.Dyn)
+	listStr := decls.NewListType(decls.String)
 	return cel.NewEnv(
 		cel.Declarations(
 			decls.NewIdent("body", mapStrDyn, nil),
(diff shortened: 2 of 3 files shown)


bigkevmcd:
This could easily be flipped to become `split(body.ref, '/')` instead of `body.ref.split('/')`

dibyom:
> This could easily be flipped to become split(body.ref, '/') instead of body.ref.split('/')

Yeah, I think I like `split(body.ref, '/')` slightly better since it also matches the pattern from `truncate`

bitsofinfo:
really, really need this function to be able to mutate values in the body. Please merge at least into the master so I can start playing with it. thanks!

bobcatfish:
Thanks for adding the tests cases and docs @bigkevmcd !!

This seems fine but I think it would be worth a larger conversation about how many of these custom functions we want to support.

Shameless plug for Rego, looks like it comes with a number of string manipulation functions built in: https://www.openpolicyagent.org/docs/latest/policy-reference/#strings

Related: @wlynch feature request for string splitting https://github.com/google/cel-go/issues/306

/approve
/lgtm

image

bitsofinfo:
@bobcatfish I agree, would be great if some bridge could be developed to just expose functions available in other libraries that would not have to be directly maintained by tekton. i.e. like how helm exposes all the go template functions provided by sprig (http://masterminds.github.io/sprig/) greatly enhancing the capabilities.

if another lib like you suggested could be added to expose much more CEL string manipulation functions that would be awesome!

bigkevmcd:
@bobcatfish yes, I'd like to deprecate `match` in favour of `canonical` as there's less ambiguity (there's a `matches` function in CEL already).

Other than that, I can't see many more items being added, splitting strings, decoding base64 and trimming strings are the things that I wanna do in my pipelines, but clearly, there might be other useful things, we're dealing with broadly the same data across different environments.

I currently have no plans for further functions, but it's fairly easy to add them... :-)

dibyom:
Agreed. I think we need base64 (for things like pub/sub events that are base64 encoded); trimming strings (tags, commits, refs). The header one is interesting...mainly due to how go represents header values as a array of strings. Other than these 3, I don't see any new ones in the short term!

bitsofinfo:
is there a nightly release I can point to that has this? or i'll have to build it all and deploy locally?

TristonianJones:
Hey everyone, awesome to see CEL being embraced in the wild. I am totally open to reviewing PRs for google/cel-go#306 if that would help accelerate your project.

bitsofinfo:
is this not possible? 

```
kind: TriggerBinding
metadata:
  name: deploy
spec:
  params:

    - name: appName
      value: "$(split(body.action_values[0],'_')[1])"
```

bigkevmcd:
> is this not possible?

No, see https://github.com/tektoncd/triggers/issues/367

But, you can use the CEL interceptor, without actually intercepting, and use the overlay function there.

Something like this should work:

```yaml
kind: EventListener
metadata:
  name: listener-interceptor
spec:
  triggers:
    - name: demo-trigger
      interceptors:
        - cel:
            overlays:
            - key: intercepted.appName
              expression: "split(body.action_values[0],'_')[1]"
      bindings:
        - name: deploy
      template:
        name: deploy-template
```

Then you could use the value like this:

```yaml
kind: TriggerBinding
metadata:
  name: deploy
spec:
  params:
    - name: appName
      value: $(body.intercepted.appName)
```

bitsofinfo:
@bigkevmcd yep thats what i ended up doing, thanks!

bitsofinfo:
@TristonianJones would be great to have a `replace` function. or... can anyone point me to an example of how to essentially do a string replace w/ CEL? I've looked but can't think of anything.

i need to take a value and replace `.` with `-` so I can use that value in meta-data names in a pipelinerun

PR #421, opened 2020-02-06
Fixes for the nightly release process

afrittoli:
# Changes

Move run resources to a dedicated folder, so they are not applied
by mistake when doing `kubectl apply -f tekton`.
Make the bucket resource name unique so that it does not conflict.
Point the bucket directly into the triggers folder, to avoid
downloading the releases of all tekton projects - making the release
process more separated from other projects release process, so we
don't have an issue if the releases run in parallel.
Remove `triggers` from the path when copying to the bucket.

Add a cloudevent resource to trigger post release services (log copy).
# Submitter Checklist

review them:

- [n/a] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```


dibyom:
Looks like some indentation issues in the resources file; otherwise LGTM

tekton/resources.yaml
```diff
@@ -51,5 +51,15 @@ metadata:
 spec:
   type: image
   params:
-   - name: url
-     value: cmd/eventlistenersink # Registry is provided via parameter, this is a hack see #569
\ No newline at end of file
+  - name: url
+    value: cmd/eventlistenersink  # Registry is provided via parameter, this is a hack see #569
+# ---
```
dibyom: Was this meant to be commented out?

afrittoli: Yeah, the URL of the sink there is cluster specific so I didn't want to include a resource that would actually be applied with a broken url

Issue #409, opened 2020-02-03
Better examples


dibyom:
In this iteration of improving our examples, I'd like to make our examples more like the ones in pipelines i.e. self contained files that a user can `kubectl apply` and then get something useful to happen e.g. something like the cron example we currently have.

Couple of conventions:
- Each file should contain all of the components needed e.g. the EventListener, TriggerBinding, TriggerTemplates etc.
 
- It'd be tedious to have a new serviceAccount/role for each; so we can reuse one or two service accounts

- Would be nice to have a curl command to exercise the Event Listener
 
- (Nice to have) A  task that waits for the eventlisteners to boot up and then makes a call to them to verify the behavior...(perhaps we can wrap https://github.com/eddiezane/hook into a task :thinking: )

PR #434, opened 2020-02-10
Replace gitlab-eventlistener with a working example

dibyom:
# Changes

I moved the example to a separate folder so that it was easy to add the sample
payload as well as instructions for trying it out locally.

Part of #409 

# Submitter Checklist

review them:


examples/gitlab/gitlab-push-listener.yaml
```diff
@@ -0,0 +1,75 @@
...
+stringData:
+  gitlabToken: abcde
+---
+apiVersion: tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: gitlab-listener
+spec:
+  # from examples/role-resources/servicaccount.yaml
+  serviceAccountName: tekton-triggers-example-sa
```
ncskier: I think we need to have the ServiceAccount `tekton-triggers-example-sa` in this file.

I think we'll also need a Role & RoleBinding, because the idea is that anyone can just `kubectl apply -f` this file, and it'll work, right?

dibyom:
> Let me know if I'm misunderstanding #409, but it was my understanding that a user would be able to kubectl apply -f the example YAML file into an empty namespace, and it would work (as long as Triggers is installed on their cluster).

Yeah, we'd need to copy the same Role/SA for each so I was hoping for the role resources we could reuse one or two. The user flow being 1. install triggers 2. apply the SA/roles 3. apply any of the examples.

I can definitely update the readme instructions to make this clearer

examples/gitlab/README.md
```diff
@@ -0,0 +1,32 @@
...
+To expose the EventListener, add a `serviceType: LoadBalancer`.
+
+### Try it out locally:
+
+1. Port forward:
+```shell script
+kubectl port-forward \ 
+ "$(kubectl get pod --selector=eventlistener=gitlab-listener -oname)" \
+  8080
+```
```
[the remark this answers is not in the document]
dibyom: Had it in an earlier section. Moved it here.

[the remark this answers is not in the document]
dibyom: The port forwarding should still work though. If we use a LoadBalancer, the user will have to find the LoadBalancerIP and then use that instead.

examples/gitlab/gitlab-push-listener.yaml
dibyom: Added the bit about applying role resources a a pre-requisite

examples/gitlab/README.md
```diff
@@ -0,0 +1,32 @@
+## GitLab Push EventListener
+
+Creates a EventListener that listens for webhook events from GitLab.
+To expose the EventListener, add a `serviceType: LoadBalancer`.
+
+### Try it out locally:
+
```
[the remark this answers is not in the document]
dibyom: added

examples/gitlab/README.md
```diff
@@ -0,0 +1,46 @@
...
+  8080
+```
+
+2. Test by sending the sample payload.
+
+```shell script
+curl -v \
+-H 'X-GitLab-Token: abcde' \
+-H 'X-Gitlab-Event: Push Hook' \
+-H 'Content-Type: application/json' \
```
ncskier: If you want to, we can put the Headers in a separate plaintext file:
```
X-GitLab-Token: abcde
X-Gitlab-Event: Push Hook
Content-Type: application/json
```
Then the curl command can look something like:
```bash
curl -v \
 -H @"examples/gitlab/gitlab-push-event-header" \
 --data-binary @"examples/gitlab/gitlab-push-event.json" \
 http://localhost:8080
```
which might be a little cleaner & easier to reproduce.

examples/gitlab/README.md
```diff
@@ -0,0 +1,46 @@
+## GitLab Push EventListener
+
+Creates an EventListener that listens for Gitlab webhook events. To expose the
+EventListener, add a `serviceType: LoadBalancer`.
+
+### Try it out locally:
+
+1. Create the service account:
```
ncskier: I think the numbers are off in this list. Maybe turn them into subheaders instead of a list?
`#### 1. Create the ServiceAccount`

examples/gitlab/README.md
```diff
@@ -0,0 +1,46 @@
+## GitLab Push EventListener
+
+Creates an EventListener that listens for Gitlab webhook events. To expose the
+EventListener, add a `serviceType: LoadBalancer`.
```
ncskier: I'm a little confused by the `serviceType` line because there's a step below about port-forwarding. Both are options for exposing the EventListener pod, so I think we should probably mention them in the same section. 

Maybe mention the `serviceType: LoadBalancer` as a side-note if someone doesn't want to port-forward?

dibyom: added a note in the port forward section

examples/gitlab/README.md
dibyom: fixed!

examples/gitlab/gitlab-push-event.json
```diff
@@ -0,0 +1,68 @@
...
+        "name": "GitLab dev user",
+        "email": "gitlabdev@dv6700.(none)"
+      },
+      "added": ["CHANGELOG"],
+      "modified": ["app/controller/application.rb"],
+      "removed": []
+    }
+  ],
+  "total_commits_count": 4
+}
```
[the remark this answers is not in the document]
dibyom: done :smile:

examples/gitlab/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+2. Test by sending the sample payload.
+
+```shell script
+curl -v \
+-H 'X-GitLab-Token: abcde' \
+-H 'X-Gitlab-Event: Push Hook' \
+-H 'Content-Type: application/json' \
+--data-binary "@gitlab-push-event.json" \
```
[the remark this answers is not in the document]
dibyom: done

examples/gitlab/gitlab-push-listener.yaml
```diff
@@ -0,0 +1,75 @@
...
+            secretRef:
+              secretName: gitlab-secret
+              secretKey: gitlabToken
+            eventTypes:
+              - Push Hook # Only push events
+      bindings:
+        - name: gitlab-push-binding
+      template:
+        name: gitlab-echo-template
+---
```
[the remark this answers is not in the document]
dibyom: removed

examples/gitlab/README.md
dibyom: Interesting idea...maybe we can come up with a convention each example can have a header.txt and a body.json or something...can be easily scripted that way!

dibyom:
> Also, I think that there's still a discussion to have about splitting up examples into multiple files or not...
> I kind of like the idea of having all resources for an example defined in one file (including the Role, RoleBinding, ServiceAccount, etc.). Using one file makes it easy for someone new to Triggers to see everything that's involved in making a Triggers example work. I think that it can be easy to forget about the ServiceAccount, Role, and RoleBinding resources when they're in separate files. Neglecting the ServiceAccount, Role, and RoleBinding could lead newcomers to make avoidable mistakes when getting started with Triggers.

My motivation for keeping the roles in one central place:
1. It adds about ~30 extra lines to the top of each example which IMO makes it harder to figure out what exactly the example is doing
2. (more importantly) we will be repeating the exact same role/binding/serviceaccount for almost all examples (the exception being the cluster binding ones). 

This is just incremental fix but what I'd eventually like is : 
1. each example in its own folder/file...I think each file should contain the triggers pieces it needs
2. a common folder for some common things e.g. the roles or a common pipeline 
3. An updated examples/Readme that tells users the pre-reqs - e.g. install triggers and the resources in the common folder. And then apply any of the examples --> I'm hoping we can move most of the info in gitlab/readme to examples/Readme once we have a few more examples converted.

What do you think?

examples/gitlab/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+1. Port forward:
+
+   ```shell script
+   kubectl port-forward \
+    "$(kubectl get pod --selector=eventlistener=gitlab-listener -oname)" \
+     8080
+   ```
+
+   **Note**: Instead of port forwarding, you can set the `serviceType` to
```
ncskier: Someone who is unfamiliar with the Triggers project might not know what the `serviceType` is, so I think that we should add a link to the `serviceType` docs (https://github.com/tektoncd/triggers/blob/master/docs/eventlisteners.md#serviceType)

ex:
'...Instead of port forwarding, you can set the [`serviceType`](https://github.com/tektoncd/triggers/blob/master/docs/eventlisteners.md#serviceType) to...'

examples/gitlab/README.md
```diff
@@ -0,0 +1,48 @@
...
+   --data-binary "@examples/gitlab/gitlab-push-event.json" \
+   http://localhost:8080
+   ```
+
+   The response status code should be `201 Created`
+
+1. You should see a new TaskRun that got created:
+
+   ```shell script
+   kubectl get taskruns |grep gitlab-run-
```
ncskier: nit: add a space
`kubectl get taskruns |grep gitlab-run-` -> `kubectl get taskruns | grep gitlab-run-`

examples/gitlab/README.md
ncskier: Yep 👍

dibyom:
> One idea to consider is using Symbolic Links to put the shared resource files into an example directory. We currently do this for the Cron example, and it could be a helpful way to make examples exist within their own directories while using shared files for common resources.

Hmm...would Kustomize work as well?

ncskier:
Hm, maybe 👍

PR #438, opened 2020-02-12
Ensure Interceptor request uses HTTP POST.

### LINKED ISSUE #437: EventListener Webhook interceptor does not retain HTTP method when proxying request to target Service
# Expected Behavior

If I have a `webhook` interceptor configured for an `EventListener`, if I do a http POST to the generated `Ingress` for that `EventListener` I expect my custom k8s service to also get a POST

# Actual Behavior

My custom interceptor service that receives the proxied/forwarded request from the webhook receives an http GET

1. Create a webhook interceptor and just log the http request it receives.
2. POST to it directly via a manually created service, you get a POST
3. POST to it via the EventListener generated Ingress endpoint and let it flow through the webhook. You receive a GET

See: https://github.com/tektoncd/triggers/blob/master/pkg/sink/sink.go#L165-L168

It appears to build a new request here, just using the body/headers sourced from the original request. if you follow the code path to this method you will see the method is never extracted or retained.

# Additional Info

https://tektoncd.slack.com/archives/CKUSJ2A5D/p1581459082059900
https://tektoncd.slack.com/archives/CKUSJ2A5D/p1581517068075700

I'm trying to do a POC of tekton for an eval, and this is a blocker for me.

wlynch:
Fixes #437

# Changes

# Submitter Checklist

review them:

# Release Notes

```
BREAKING CHANGE: Interceptors now require HTTP POST requests.
```

### DIFF
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -122,7 +122,7 @@ contains the default values defined in
 To access logs for the EventListener sink, you can query for pods with the
 `eventlistener` label set to the name of your EventListener resource:
 
-```shell script
+```shell
 kubectl get pods --selector eventlistener=my-eventlistener
 ```
 
@@ -186,6 +186,7 @@ To be an Event Interceptor, a Kubernetes object should:
 
 - Be fronted by a regular Kubernetes v1 Service over port 80
 - Accept JSON payloads over HTTP
+- Accept HTTP POST requests with JSON payloads.
 - Return a HTTP 200 OK Status if the EventListener should continue processing
   the event
 - Return a JSON body back. This will be used by the EventListener as the event
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -163,6 +163,7 @@ func (r Sink) executeInterceptors(t *triggersv1.EventListenerTrigger, in *http.R
 
 	// The request body to the first interceptor in the chain should be the received event body.
 	request := &http.Request{
+		Method: http.MethodPost,
 		Header: in.Header,
 		Body:   ioutil.NopCloser(bytes.NewBuffer(event)),
 	}
@@ -190,6 +191,7 @@ func (r Sink) executeInterceptors(t *triggersv1.EventListenerTrigger, in *http.R
 		// Set the next request to be the output of the last response to enable
 		// request chaining.
 		request = &http.Request{
+			Method: http.MethodPost,
 			Header: resp.Header,
 			Body:   ioutil.NopCloser(resp.Body),
 		}
--- a/pkg/sink/sink_test.go
+++ b/pkg/sink/sink_test.go
@@ -643,6 +643,10 @@ type sequentialInterceptor struct {
 
 func (f *sequentialInterceptor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	f.called = true
+	if r.Method != http.MethodPost {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
 	var data map[string]int
 	if err := json.NewDecoder(r.Body).Decode(&data); err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
@@ -697,25 +701,39 @@ func TestExecuteInterceptor(t *testing.T) {
 		Interceptors: []*triggersv1.EventInterceptor{a, a},
 	}
 
-	req, err := http.NewRequest(http.MethodPost, "/", nil)
-	if err != nil {
-		t.Fatalf("http.NewRequest: %v", err)
-	}
-	resp, header, err := r.executeInterceptors(trigger, req, []byte(`{}`), "", logger)
-	if err != nil {
-		t.Fatalf("executeInterceptors: %v", err)
-	}
+	for _, method := range []string{
+		http.MethodGet,
+		http.MethodHead,
+		http.MethodPost,
+		http.MethodPut,
+		http.MethodPatch,
+		http.MethodDelete,
+		http.MethodConnect,
+		http.MethodOptions,
+		http.MethodTrace,
+	} {
+		t.Run(method, func(t *testing.T) {
+			req, err := http.NewRequest(method, "/", nil)
+			if err != nil {
+				t.Fatalf("http.NewRequest: %v", err)
+			}
+			resp, header, err := r.executeInterceptors(trigger, req, []byte(`{}`), "", logger)
+			if err != nil {
+				t.Fatalf("executeInterceptors: %v", err)
+			}
 
-	var got map[string]int
-	if err := json.Unmarshal(resp, &got); err != nil {
-		t.Fatalf("json.Unmarshal: %v", err)
-	}
-	want := map[string]int{"i": 2}
-	if diff := cmp.Diff(want, got); diff != "" {
-		t.Errorf("Body: -want +got: %s", diff)
-	}
-	if diff := cmp.Diff([]string{"1", "2"}, header["Foo"]); diff != "" {
-		t.Errorf("Header: -want +got: %s", diff)
+			var got map[string]int
+			if err := json.Unmarshal(resp, &got); err != nil {
+				t.Fatalf("json.Unmarshal: %v", err)
+			}
+			want := map[string]int{"i": 2}
+			if diff := cmp.Diff(want, got); diff != "" {
+				t.Errorf("Body: -want +got: %s", diff)
+			}
+			if diff := cmp.Diff([]string{"1", "2"}, header["Foo"]); diff != "" {
+				t.Errorf("Header: -want +got: %s", diff)
+			}
+		})
 	}
 }
 


bitsofinfo:
The entire thing should not manually set anything on the inbound request (other than the additional headers/body changes defined on the interceptor yaml). It should take the inbound request (headers/method/body if present) and just send it along to the interceptors.  

Forcing it to POST would now break people who inadvertently are benefiting from the hardwired/buggy GET its doing now.

my vote is just leave the method be.

pkg/sink/sink.go
```diff
@@ -190,6 +191,7 @@ func (r Sink) executeInterceptors(t *triggersv1.EventListenerTrigger, in *http.R
 		// Set the next request to be the output of the last response to enable
 		// request chaining.
 		request = &http.Request{
+			Method: http.MethodPost,
```
bitsofinfo: Seems to me this should be `Method: in.Method`

wlynch: I think this would be okay for the first request, but not for additional requests since we would generally expect the next request to include whatever body was returned from the last interceptor. wdyt?

bitsofinfo: yes, if the interceptors, the way they are defined as I see in yaml (i.e. an array executed in order), then yes that implies a chain. 

So yes the 1st interceptor would get everything "as-is" from the initial request (unless altered in yaml spec for interceptor) Apply custom headers/custom method (i.e. if #439) then send to interceptor-1. 

Get interceptor-1 result, capture body, send mutated body + original headers + yaml defined headers + method (if #439).... The original body never gets sent but 1x. The original headers always get sent + custom headers. Original method always sent unless overriden per interceptor yaml config.

khrm:
> Forcing it to POST would now break people who inadvertently are benefiting from the hardwired/buggy GET its doing now.
Ideally, it should be only PUT or POST but I can see that we need to do something for backward compatibility. 
> my vote is just leave the method be (i.e. method should = original inbound method)
Maybe give choice to end-user to define what method to accept? We can deprecate GET in 0.4?

dibyom:
I think most webhooks follow POST + Json body; so we should update our docs to mention that's what we support. But to not break current users let's keep the original inbound method for now and then set it to POST for 0.4?

dibyom:
I think we should resolve this by 0.3? My suggestion:

1. We change the Interceptor interface to only accept POST requests. We can do this in a phased manner like out other breaking changes i.e. announce deprecation for 0.3 and remove support in 0.4. This should allow users to upgrade their webhook interceptors.

2. Unless there is a compelling use case for doing so, we should not add configuration to allow the user to pick the method in the interceptor config. Happy to reconsider this if there is some case where relying on the HTTP method is the only good way.

What do you think? @wlynch @khrm @bitsofinfo

bitsofinfo:
agree some fix is needed for 0.3 as I'm sort of stuck :)

disagree about forcing it to be `POST` only.... just let the request flow through unaltered as-is except the existing header mutations that are supported. I don't think anyone can predict peoples individual use cases for this and just only supporting POST seems limiting. Leave it more open and let the end operator customize the behavior they want. What if somebody needs to have triggers that fire due to PUTs or GETs? I just don't see the reason to force POST on everyone. 

I don't think its needed for 0.3, but for future, I definitely am advocating to permit the operator to alter the method for chained scenarios as @wlynch and I were chatting about. (i.e. where interceptorA might output a body who the operator wants to forward next to interceptorB that only accepts requests via a different method etc (PUT etc). In any case, this method configurability is less of a need IMHO than just not forcing any method on the initial request.

ncskier:
I'm coming into this a little late, so I might not understand the problem, but can we set the Method to `in.Method`? So if the `in` request is type `PUT`, then we'll pass it on to the request as type `PUT`?

bitsofinfo:
in sink.go around here: https://github.com/tektoncd/triggers/pull/438#pullrequestreview-357743922

ncskier:
Ah, thanks @bitsofinfo 👍Sorry for commenting without reading 🤦‍♂ 

> What if somebody needs to have triggers that fire due to PUTs or GETs? I just don't see the reason to force POST on everyone.

I think that what Dibyo is proposing is only for Interceptors to receive POST requests. The EventListener would still be able to receive all types of HTTP requests (PUT, POST, GET, etc). However, they would all be converted to a POST request when they are sent to the Interceptor chain.

@bitsofinfo what is your use case for your Interceptor? Is there a particular reason that you want it to accept PUT events instead of POST events (other than the fact that it's strange to change a PUT event into a POST event)?

Also, @wlynch do we have any explicit use cases for Interceptor chaining? It would be helpful to know what we need/want to support.

bitsofinfo:
> I think that what Dibyo is proposing is only for Interceptors to receive POST requests. The EventListener would still be able to receive all types of HTTP requests (PUT, POST, GET, etc). However, they would all be converted to a POST request when they are sent to the Interceptor chain.

EventListener would accept anything (which it already does today) but then force convert the http method to a POST...... but why? Thats just the inverse of what it's doing today (unintentional I assume) which already is causing problems, i.e. today EventListener accepts anything, but changes the http method to a GET. I think by doing anything like this you are just trading one problem for another. Making one user's life easier but making another's a pain. Just leave the inbound request method as-is and send it along for now. Future? #439 and give the operator full control of this behavior for any interceptor.

>@bitsofinfo what is your use case for your Interceptor? Is there a particular reason that you want it to accept PUT events instead of POST events (other than the fact that it's strange to change a PUT event into a POST event)?`

I currently do not have a need for PUT, I'm just advocating for a more flexible and configurable contract overall. 

I'm trying to prototype triggers firing in response to interactive slack messages (i.e. user presses a slack message interactive button and it hits my tekton event listener). Slack sends POSTs.... so the PR as it stands would fix my problem just fine and I'll move on.  

But again I'm just advocating for more flexibility and adaptability in the tekton listener/interceptor contract. How can we predict what people will utilize for interceptors they want to hook into the flow? I think assuming that everyone is crafting a custom coded thing whereby they can comply with a forced POST contract might not be correct. 

Some folks who are not coders might just want to wire up an "off the shelf" pre-existing 3rd party thing to serve as the backend target for an interceptor that accepts only GETS, or PUTS or whatever. Who knows; I don't think Tekton should care.

dibyom:
@ncskier A use case for chained interceptors - validate using a Github interceptors, add a webhook interceptor to call out to GH for adding additional info, and then perhaps a CEL filter. I think these would be fairly common.

@bitsofinfo If you are writing your own interceptor and can modify it to accept requests regardless of the HTTP method, you should be able to make the Slack based approach work regardless of the status of this PR.

In general, while I do get the point about flexibility, we try to add features only if we understand the use cases.

bitsofinfo:
Can someone please provide some arguments as to why just leaving the http method "as-is" would be detrimental to this component and what it would break?

ncskier:
> Can someone please provide some arguments as to why just leaving the http method "as-is" would be detrimental to this component and what it would break?

Sorry if this wasn't clear. Interceptors have the ability to daisy-chain into each other. So an Interceptor can return a Body. This Body will be used in the request to a second Interceptor, or if there are no more Interceptors, then the Body will be used to fill the specified TriggerTemplate according to its TriggerBindings. 

Let's take the case where a user has two Interceptors, and receives a webhook with a GET request, and we leave the HTTP Method "as-is." The GET request is received by the EventListener, and sent to the first Interceptor. Let's say the first Interceptor wants to send back a Body with some additional information. Now, we have the Body with additional information to send to the second Interceptor; however, now I don't think we want to send a GET request with a Body? So the question becomes, do we change the Method type now?

Here's a diagram that I hope will help explain the scenario 🙂 

```
GET -> [EventListener] -> GET -> [Interceptor 1] -> GET + Body -> [EventListener] -> ??? -> [Interceptor 2]
```

bitsofinfo:
@ncskier yep, totally get this and why I also made #439. So then tekton would let the operator (via config per interceptor just like the headers) to fill in the blank in your question mark `???` above

wlynch:
My idea for #439 was primarily some spit-balling - I wouldn't read too much into it about whether it is something we should definitely do or not. :upside_down_face:

The more I think about it the more I agree with @ncskier to prefer just using only POST for interceptors, at least until we have a more concrete use case where a POST would be insufficient.

@dibyom if this is the route we go down, any concerns about just enforcing POST for interceptors, or do you think we need to take extra steps to ensure compatibility?

dibyom:
> @dibyom if this is the route we go down, any concerns about just enforcing POST for interceptors, or do you think we need to take extra steps to ensure compatibility?

Nope, I think we should mention it in the release notes though!

bitsofinfo:
Disagree with forced POST but I'll stop repeating myself, but at least it will be fixed for me! 

This change will definitely breaks folks knowingly or unknowingly relying on the current behavior, clearly bold it out as a breaking change.

ncskier:
> This change will definitely breaks folks knowingly or unknowingly relying on the current behavior, clearly bold it out as a breaking change.

I'll make sure to list this as a breaking change in the release notes. We have a policy of documenting all breaking changes in our release notes 👍 

We'll also need explicit documentation outlining the Interceptor contract in this PR.

docs/eventlisteners.md
```diff
@@ -107,12 +107,12 @@ binding/templating.
 To be an Event Interceptor, a Kubernetes object should:
 
 - Be fronted by a regular Kubernetes v1 Service over port 80
-- Accept JSON payloads over HTTP
+- Accept JSON payloads as a HTTP POST request
```
dibyom:
```suggestion
- Accept HTTP POST requests with JSON payloads
```

wlynch: Applied.

wlynch:
> > @dibyom if this is the route we go down, any concerns about just enforcing POST for interceptors, or do you think we need to take extra steps to ensure compatibility?
> 
> Nope, I think we should mention it in the release notes though!

Done. The constraint is also listed in the event listener docs. @dibyom @ncskier Let me know if you think we should document this anywhere else.

wlynch:
> @wlynch Can you please add a test for `executeInterceptors` for the case when a non-POST method request is sent as the `in` request. For example, when the `in` request is a GET request. Then verify that the Interceptor service receives a POST request.

Done

Tallisado:
Thanks for the fix. To help anyone still on a previous version scratching their head -- I moved to using gin-gonic from net/http and was seeing this in the el logs (do to uncaught GET when I expected a POST):
```
request rejected; status: 404 Not Found; message: 404 page not found
```

bitsofinfo:
thanks all, got around to getting this working finally in my POC. Works great!

Issue #356, opened 2020-01-17
Mutating webhook times out when EventListener has multiple triggers


kaczyns:
# Expected Behavior
I can use the tekton dashboard to create multiple webhooks (greater than 8 or so).

# Actual Behavior
When creating the 9th webhook, the mutating webhook `triggers-webhook.tekton.dev` times out while processing the `EventListener` that was updated by the tekton dashboard.

# Steps to Reproduce the Problem

1. Install tekton dashboard v0.3.0 https://github.com/tektoncd/dashboard/releases/tag/v0.3.0
2. Create multiple webhooks https://github.com/tektoncd/experimental/blob/master/webhooks-extension/docs/GettingStarted.md#creating-a-new-webhook
3. After the 8th (or so) webhook is created, subsequent creations fail because the mutating webhook times out.

# Additional Info
Multiple users have reported this problem (kabanero-io/kabanero-foundation#150).  All are using OpenShift 4.2, openshift-pipelines-operator 0.8 (includes triggers 0.1.0) and tekton dashboard 0.3.0.  It appears that in OpenShift the mutating webhook timeout is 13 seconds.  The `EventListener` has about 24 triggers in it (3 per github webhook) before the mutating webhook starts to time out.

Perhaps a single `EventListener` was not intended to contain so many triggers... in any case it would be good to know if this is a reasonable configuration.

I re-built the tekton-triggers-webhook container and put an elapsed time in the mutating admission webhook method (vendor/knative.dev/pkg/webhook/webhook.go, added `time.Since(ttStart)`):
```golang
	logger.Infof("AdmissionReview for %#v: %s/%s response=%#v elapsedTime=%v",
		review.Request.Kind, review.Request.Namespace, review.Request.Name, reviewResponse, time.Since(ttStart))
```
The elapsed time increases by about 1.2 seconds per github webhook that I create on my system (using the dashboard).  The container can be found here:
```
docker.io/kaczyns/webhook-dd1edc925ee1772a9f76e2c1bc291ef6@sha256:a1c5a6bce855e97135cbbc7bfea8d7a6862084b19d4ef8434df4b677f7503227
```

ncskier:
I was able to reproduce this issue with Tekton Triggers v0.1.0 installed on OpenShift after creating an EventListener with 37 triggers.

I wrote up this quick bash script to reproduce the error if anyone wants to try it out:
```bash
#!/bin/bash

# Cleanup
rm eventlistener_*

# Define variables
max=50
[...]

PR #449, opened 2020-02-19
Add Regression test for large number of Triggers

### LINKED ISSUE #405: Regression test for multiple triggers admission webhook timeout
Write a test that catches the bug described in issue #356. The bug was fixed in the v0.2.1 release; however, we need a test to verify that the bug will not be re-introduced into our codebase.

### Additional Info
Issue https://github.com/tektoncd/triggers/issues/356 found a bug in Triggers v0.1.0 where the admission webhook times out after adding around 35 triggers to an EventListener. This was because the EventListener validation was taking too long. For more details, read issue https://github.com/tektoncd/triggers/issues/356.

ncskier:
# Changes

Fixes #405

Issue #356 found a bug in Triggers v0.1.0 (fixed in v0.2.0) where the
admission webhook times out after adding around 35 Triggers to an
EventListener.

This PR adds a regression test where we create an EventListener with
1000 Triggers.

---

I also moved the clean up code for cluster-scoped resources (from https://github.com/tektoncd/triggers/commit/7ba80a7757fbcd537624f3e24daed5f213c4dd96) out of `init_test.go` and into `eventlistener_test.go`, because the code was cleaning up resources specific to the `eventlistener_test.go`. @dibyom do you think I should put this in a separate commit, or is it ok as it is?

# Submitter Checklist

review them:

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -117,6 +117,13 @@ be used to run only [the unit tests](#unit-tests), i.e.:
 // +build e2e
 ```
 
+#### Cleaning up cluster-scoped resources
+
+Each integration test runs in its own Namespace; each Namespace is torn down
+after its integration test completes. However, cluster-scoped resources will not
+be deleted when the Namespace is deleted. So, each test must delete all the
+cluster-scoped resources that it creates.
+
 #### Setup tests
 
 The `setup` function in [init_tests.go](./init_test.go) will initialize client
--- a/test/eventlistener_scale_test.go
+++ b/test/eventlistener_scale_test.go
@@ -1,0 +1,62 @@
+// +build e2e
+
+/*
+Copyright 2019 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"fmt"
+	"testing"
+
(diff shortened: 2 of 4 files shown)


test/README.md
```diff
@@ -117,6 +117,13 @@ be used to run only [the unit tests](#unit-tests), i.e.:
 // +build e2e
 ```
 
+#### Cleaning up cluster-scoped resources
+
+Each integration test runs in its own Namespace; each Namespace is torn down
```
dibyom: :+1:

test/eventlistener_scale_test.go
```diff
@@ -0,0 +1,62 @@
...
+		trigger.Name = fmt.Sprintf("%d", i)
+		el.Spec.Triggers = append(el.Spec.Triggers, trigger)
+	}
+	el, err = c.TriggersClient.TektonV1alpha1().EventListeners(namespace).Create(el)
+	if err != nil {
+		t.Fatalf("Error creating EventListener: %s", err)
+	}
+
+	// Verify that the EventListener was created properly
+	if err := WaitFor(eventListenerReady(t, c, namespace, el.Name)); err != nil {
```
dibyom: Do we need to verify that the EL actually contains the 1000 triggers?

dibyom: Or perhaps that the sink got created?

ncskier: This should verify that the sink got created, right? Because it waits for the sink Pod to be ready. As for verifying that the EventListener actually contains the 1000 Triggers... I guess we could trigger an event to fire all of the Triggers (maybe create a TaskRun or something?). I don't know if this would stress out our test cluster though 😅

dibyom: ahh you are right...I did not notice that it actually waits for the sink to be created

ncskier: What about verifying the 1000 Triggers? Do you think it's alright without that?

dibyom: Yes, we have other tests that verify that we support multiple triggers

Issue #77, opened 2019-08-21
Users should only be able to create TriggerTemplates for resources they are permissioned to use


pmorie:
In the current design, users would be able to use the `spec.resourceTemplates` field to escalate their degree of write permission within the cluster; users of TriggerTemplate are able to create any resource that the service account responsible for creating resources in a TriggerTemplate has permission to create. For example, if a user does not have access to create Pods, but the service account servicing TriggerTemplate does, a user could put a Pod into `spec.resourceTemplates` and have it created.

We should protect against this form of privilege escalation by having a validating webhook that uses information from the authorizer (this is part of the payload webhooks are passed) to make SAR checks that establish that the user creating or updating a TriggerTemplate has permission to create those resources in the cluster.

vtereso:
@pmorie I'm not too familiar with the distinction between users and service accounts. [This](https://kubernetes.io/docs/reference/access-authn-authz/authentication/) says:
> Kubernetes does not have objects which represent normal user accounts

Could you further explain how a user would have access to such a service account where it would be an escalation?

bobcatfish:
After investigating and thinking about this a bit more, I think the TL;DR is:

afaik @pmorie is pointing out that in our current design, if a user can create an EventListener (with a serviceaccount that can make PipelineRuns, read TriggerBindings, etc.), and you can create Pipelines + Tasks, they'll then have additionally the ability to:

1. Create Resources like PipelineRuns
2. Creating a PipelineRun means you can create a pod
3. Being able to create a pod means you can basically do anything you want within a cluster, and access any existing service accounts w/in the cluster

So basically for example you might intend for a user to only be able to run Pipeline Foo in a cluster in response to a webhook, but you may not realize you actually gave the user the ability to do nearly anything they want in the cluster.

I'm not sure triggers or pipelines should solve this directly, but we might want to provide some guidance (and maybe even best practice setup Tasks [like the one @akihikokuroda added!](https://github.com/tektoncd/triggers/blob/master/docs/createwebhook.md)) for ppl to setup secure clusters, e.g. something like:

1. Restrict who can create Pipelines and Tasks <-- maybe this solves like 90% of the problem?
2. Use `SubjectAccessReview` to express stuff like: in order for a user to be allowed to make an `EventListener`, they must also be allowed to create a `PipelineRun` (and if you want to create a `PipelineRun`, you need to be allowed to create a `pod`)
[...]

PR #454, opened 2020-02-24
introduce authentication/authorization at the EventListenerTrigger le…

gabemontero:
…vel (with default back to existing EventListener level)

# Changes

While this does not fully address https://github.com/tektoncd/triggers/issues/77, this change intends to provide more more granular authorization, besides just the `EventListener` service account,
and sets the stage for "multi-tenant themed" authorization around creating tekton resources 
from triggers and the `EventListener` sink.

As discussed in recent instances of the weekly project sync call, the actual employing of SubjectAccessReview checks (and the k8s admission related resources those would require) belongs in the https://github.com/tektoncd/pipelines project, where the identity and permissions
around the existing SA of the `EventListener` or the new SA for the `EventListenerTrigger` introduced here would be used for any SAR calls in the core pipeline controller.

As such, we could ultimately gate merging of this, if we reached consensus on this, on that tektoncd/pipeline work.  However, this change does not *require* that work.  And it provides an opt in
to finer grained control over which if the tekton CRD types can be created from a given trigger.

And certainly, while reviewing this, if finer grained authorization is deemed desirable at all, one possible outcome is that we want even finer grained control, and incorporate service account on objects with say the `EventListenerTrigger`

# Submitter Checklist

review them:

- [/] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [/] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [/] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

# Release Notes

The specification of a service account on the `EventListenerTrigger` is now available as an override of the service account of the `EventListener` to facilitate permissions changes wrt creating Tekton objects as a result of the sink receiving an event.

### DIFF
--- a/docs/clustertriggerbindings.md
+++ b/docs/clustertriggerbindings.md
@@ -12,7 +12,6 @@ designed to encourage reusability clusterwide. You can reference a
 ClusterTriggerBinding in any EventListener in any namespace.
 
 <!-- FILE: examples/clustertriggerbindings/clustertriggerbinding.yaml -->
-
 ```YAML
 apiVersion: triggers.tekton.dev/v1alpha1
 kind: ClusterTriggerBinding
@@ -28,6 +27,7 @@ spec:
       value: $(header.Content-Type)
 ```
 
+
 You can specify multiple ClusterTriggerBindings in a Trigger. You can use a
 ClusterTriggerBinding in multiple Triggers.
 
@@ -35,7 +35,6 @@ In case of using a ClusterTriggerBinding, the `Binding` kind should be added.
 The default kind is TriggerBinding which represents a namespaced TriggerBinding.
 
 <!-- FILE: examples/eventlisteners/eventlistener-clustertriggerbinding.yaml -->
-
 ```YAML
 ---
 apiVersion: triggers.tekton.dev/v1alpha1
@@ -54,3 +53,4 @@ spec:
       template:
         name: pipeline-template
 ```
+
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -15,12 +15,73 @@ Tekton resources. In addition, EventListeners allow lightweight event processing
 using [Event Interceptors](#Interceptors).
 
 - [Syntax](#syntax)
+  - [ServiceAccountName](#serviceAccountName)
   - [Triggers](#triggers)
     - [Interceptors](#Interceptors)
-  - [ServiceAccountName](#serviceAccountName)
 - [Logging](#logging)
 - [Labels](#labels)
 - [Examples](#examples)
+
+## Multi-Tenant Concerns
+
+The EventListener is effectively an additional form of client into Tekton, versus what 
+example usage via `kubectl` or `tkn` which you have seen elsewhere.  In particular, the HTTP based
+events bypass the normal Kubernetes authentication path you get via `kubeconfig` files 
+and the `kubectl config` family of commands.
+
+As such, there are set of items to consider when deciding how to 
+
+- best expose (each) EventListener in your cluster to the outside world.
+- best control how (each) EventListener and the underlying API Objects described below access, create,
+and update Tekton related API Objects in your cluster.
+
+Minimally, each EventListener has its [ServiceAccountName](#serviceAccountName) as noted below and all
+events coming over the "Sink" result in any Tekton resource interactions being done with the permissions 
+assigned to that ServiceAccount.
+
+However, if you need differing levels of permissions over a set of Tekton resources across the various
+[Triggers](#triggers) and [Interceptors](#Interceptors), where not all Triggers or Interceptors can 
+manipulate certain Tekton Resources in the same way, a simple, single EventListener will not suffice.
+
+Your options at that point are as follows:
+
+### Multiple EventListers (One EventListener Per Namespace)
+
+You can create multiple EventListener objects, where your set of Triggers and Interceptors are spread out across the 
+EventListeners.
+
+If you create each of those EventListeners in their own namespace, it becomes easy to assign 
+varying permissions to the ServiceAccount of each one to serve your needs.  And often times namespace
+creation is coupled with a default set of ServiceAccounts and Secrets that are also defined.
+So conceivably some administration steps are taken care of.  You just update the permissions
+of the automatically created ServiceAccounts.
+
+Possible drawbacks:
+- Namespaces with associated Secrets and ServiceAccounts in an aggregate sense prove to be the most expensive
+items in Kubernetes underlying `etcd` store.  In larger clusters `etcd` storage capacity can become a concern.
+- Multiple EventListeners means multiple HTTP ports that must be exposed to the external entities accessing 
+the "Sink".  If you happen to have a HTTP Firewall between your Cluster and external entities, that means more
+administrative cost, opening ports in the firewall for each Service, unless you can employ Kubernetes `Ingress` to
+serve as a routing abstraction layer for your set of EventListeners. 
+
+### Multiple EventListeners (Multiple EventListeners per Namespace)
+
+Multiple EventListeners per namespace will most likely mean more ServiceAccount/Secret/RBAC manipulation for
+the administrator, as some of the built in generation of those artifacts as part of namespace creation are not
+applicable.
+
+However you will save some on the `etcd` storage costs by reducing the number of namespaces.
+
+Multiple EventListeners and potential Firewall concerns still apply (again unless you employ `Ingress`).
+
+### ServiceAccount per EventListenerTrigger
+
+Being able to set a ServiceAccount on an EventListenerTrigger allows for finer grained permissions as well.
+
+You still have to create the additional ServiceAccounts.
+
+But staying within 1 namespace and minimizing the number of EventListeners with their associated "Sinks" minimizes 
+concerns around `etcd` storage and port considerations with Firewalls if `Ingress` is not utilized.
 
 ## Syntax
 
@@ -47,6 +108,39 @@ the following fields:
 [kubernetes-overview]:
   https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields
 
+### ServiceAccountName
+
+The `serviceAccountName` field is required. The ServiceAccount that the
+EventListener sink uses to create the Tekton resources. The ServiceAccount needs
+a role with the following rules:
+
+<!-- FILE: examples/role-resources/triggerbinding-roles/role.yaml -->
+
+```YAML
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: tekton-triggers-example-minimal
+rules:
+# Permissions for every EventListener deployment to function
+- apiGroups: ["triggers.tekton.dev"]
+  resources: ["eventlisteners", "triggerbindings", "triggertemplates"]
+  verbs: ["get"]
+- apiGroups: [""]
+  # secrets are only needed for Github/Gitlab interceptors, serviceaccounts only for per trigger authorization
+  resources: ["configmaps", "secrets", "serviceaccounts"]
+  verbs: ["get", "list", "watch"]
+# Permissions to create resources in associated TriggerTemplates
+- apiGroups: ["tekton.dev"]
+  resources: ["pipelineruns", "pipelineresources", "taskruns"]
+  verbs: ["create"]
+```
+
+If your EventListener is using
+[`ClusterTriggerBindings`](./clustertriggerbindings.md), you'll need a
+ServiceAccount with a
+[ClusterRole instead](../examples/role-resources/clustertriggerbinding-roles/clusterrole.yaml).
+
 ### Triggers
 
 The `triggers` field is required. Each EventListener can consist of one or more
@@ -71,37 +165,27 @@ triggers:
       name: pipeline-template
 ```
 
-### ServiceAccountName
+Also, to support multi-tenant styled scenarios, where an administrator may not want all triggers to have
+the same permissions as the `EventListener`, a service account can optionally be set at the trigger level
+and used if present in place of the `EventListener` service account when creating resources:
 
-The `serviceAccountName` field is required. The ServiceAccount that the
-EventListener sink uses to create the Tekton resources. The ServiceAccount needs
-a role with the following rules:
+```yaml
+triggers:
+  - name: trigger-1
+    serviceAccount: 
+      name: trigger-1-sa
+      namespace: event-listener-namespace
+    interceptors:
+      - github:
+          eventTypes: ["pull_request"]
+    bindings:
+      - name: pipeline-binding
+      - name: message-binding
+    template:
+      name: pipeline-template
+``` 
 
-<!-- FILE: examples/role-resources/triggerbinding-roles/role.yaml -->
-
-```YAML
-kind: Role
-apiVersion: rbac.authorization.k8s.io/v1
-metadata:
-  name: tekton-triggers-example-minimal
-rules:
-# Permissions for every EventListener deployment to function
-- apiGroups: ["triggers.tekton.dev"]
-  resources: ["eventlisteners", "triggerbindings", "triggertemplates"]
-  verbs: ["get"]
-- apiGroups: [""]
-  resources: ["configmaps", "secrets"]  # secrets are only needed for Github/Gitlab interceptors
-  verbs: ["get", "list", "watch"]
-# Permissions to create resources in associated TriggerTemplates
-- apiGroups: ["tekton.dev"]
-  resources: ["pipelineruns", "pipelineresources", "taskruns"]
-  verbs: ["create"]
-```
-
-If your EventListener is using
-[`ClusterTriggerBindings`](./clustertriggerbindings.md), you'll need a
-ServiceAccount with a
-[ClusterRole instead](../examples/role-resources/clustertriggerbinding-roles/clusterrole.yaml).
+The default ClusterRole for the EventLister allows for reading ServiceAccounts from any namespace.
 
 ### ServiceType
 
@@ -414,7 +498,6 @@ processed, and the `overlays` applied.
 
 Optionally, no `filter` expression can be provided, and the `overlays` will be
 applied to the incoming body.
-
 <!-- FILE: examples/eventlisteners/cel-eventlistener-no-filter.yaml -->
 
 ```YAML
@@ -541,6 +624,7 @@ spec:
   - name: branch
     value: $(body.pull_request.head.short_sha)
 ```
+
 
 ## Examples
 
--- a/docs/triggerbindings.md
+++ b/docs/triggerbindings.md
@@ -12,7 +12,6 @@ parameters. The separation of `TriggerBinding`s from `TriggerTemplate`s was
 deliberate to encourage reuse between them.
 
 <!-- FILE: examples/triggerbindings/triggerbinding.yaml -->
-
 ```YAML
 apiVersion: triggers.tekton.dev/v1alpha1
 kind: TriggerBinding
@@ -27,6 +26,7 @@ spec:
   - name: contenttype
     value: $(header.Content-Type)
 ```
+
 
 `TriggerBinding`s are connected to `TriggerTemplate`s within an
 [`EventListener`](eventlisteners.md), which is where the pod is actually
--- a/docs/triggertemplates.md
+++ b/docs/triggertemplates.md
@@ -11,7 +11,6 @@ A `TriggerTemplate` is a resource that can template resources.
 **anywhere** within the resource template.
 
 <!-- FILE: examples/triggertemplates/triggertemplate.yaml -->
-
 ```YAML
 apiVersion: triggers.tekton.dev/v1alpha1
 kind: TriggerTemplate
@@ -52,6 +51,7 @@ spec:
           - name: url
             value: $(params.gitrepositoryurl)
 ```
+
 
 Similar to
 [Pipelines](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md),`TriggerTemplate`s
--- a/pkg/apis/triggers/v1alpha1/event_listener_types.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_types.go
@@ -67,6 +67,15 @@ type EventListenerTrigger struct {
 	// +optional
 	Name         string              `json:"name,omitempty"`
 	Interceptors []*EventInterceptor `json:"interceptors,omitempty"`
+	// ServiceAccount optionally associates credentials with each trigger;
+	// more granular authorization for
+	// who is allowed to utilize the associated pipeline
+	// vs. defaulting to whatever permissions are associated
+	// with the entire EventListener and associated sink facilitates
+	// multi-tenant model based scenarios
+	// TODO do we want to restrict this to the event listener namespace and just ask for the service account name here?
+	// +optional
+	ServiceAccount *corev1.ObjectReference `json:"serviceAccount,omitempty"`
 }
 
 // EventInterceptor provides a hook to intercept and pre-process events
--- a/pkg/resources/create.go
+++ b/pkg/resources/create.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"strings"
 
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+
 	"k8s.io/client-go/dynamic"
 
 	"go.uber.org/zap"
@@ -99,6 +101,9 @@ func Create(logger *zap.SugaredLogger, rt json.RawMessage, triggerName, eventID,
 	logger.Infof("For event ID %q creating resource %v", eventID, gvr)
 
 	if _, err := dc.Resource(gvr).Namespace(namespace).Create(data, metav1.CreateOptions{}); err != nil {
+		if kerrors.IsUnauthorized(err) || kerrors.IsForbidden(err) {
+			return err
+		}
 		return fmt.Errorf("couldn't create resource with group version kind %q: %v", gvr, err)
 	}
 	return nil
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -36,6 +36,7 @@ import (
 	"github.com/tektoncd/triggers/pkg/resources"
 	"github.com/tektoncd/triggers/pkg/template"
 	"go.uber.org/zap"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	discoveryclient "k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
@@ -55,6 +56,7 @@ type Sink struct {
 	EventListenerName      string
 	EventListenerNamespace string
 	Logger                 *zap.SugaredLogger
+	Auth                   AuthOverride
 }
 
 // Response defines the HTTP body that the Sink responds to events with.
@@ -93,6 +95,13 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 		go func(t triggersv1.EventListenerTrigger) {
 			localRequest := request.Clone(request.Context())
 			if err := r.processTrigger(&t, localRequest, event, eventID, eventLog); err != nil {
+				if kerrors.IsUnauthorized(err) {
+					result <- http.StatusUnauthorized
+					return
+				}
+				if kerrors.IsForbidden(err) {
+					result <- http.StatusForbidden
+				}
 				result <- http.StatusAccepted
 				return
 			}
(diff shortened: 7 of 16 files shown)


gabemontero:
@vdemeester @pmorie @chmouel FYI

gabemontero:
I'll defer on assigning reviewers myself until either one of the project admins does it, or after we've discussed this PR on the sync call and decided who should get assigned.

gabemontero:
ok at least fixed the gofmt 's 

need to look into my new integration test

gabemontero:
`eventlistener_test.go:540: Error creating ClusterTriggerBinding: clustertriggerbindings.tekton.dev "my-clustertriggerbinding" already exists`

my new test's reuse needs some work

/work-in-progress

gabemontero:
quick update: I have my new auth test passing locally when I temporary disable the existing test, in that the trigger SA without perms is getting forbidden on the create resources attempt.

When I next get to this I'll either:
- figure out how to use the event listener's `el-port` option from the integration tests so concurrent tests listen on different ports 
- or simply merge the tests to run in succession off of the same EL instance

gabemontero:
OK integration tests are passing for me locally ... let's see how it does on CI

docs/clustertriggerbindings.md
```diff
@@ -6,7 +6,6 @@ designed to encourage reusability clusterwide. You can reference a
 ClusterTriggerBinding in any EventListener in any namespace.
 
 <!-- FILE: examples/clustertriggerbindings/clustertriggerbinding.yaml -->
-
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/clustertriggerbindings.md
```diff
@@ -22,14 +21,14 @@ spec:
...
 
+
 You can specify multiple ClusterTriggerBindings in a Trigger. You can use a
 ClusterTriggerBinding in multiple Triggers.
 
 In case of using a ClusterTriggerBinding, the `Binding` kind should be added.
 The default kind is TriggerBinding which represents a namespaced TriggerBinding.
 
 <!-- FILE: examples/eventlisteners/eventlistener-clustertriggerbinding.yaml -->
-
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/clustertriggerbindings.md
```diff
@@ -22,14 +21,14 @@ spec:
       value: $(header.Content-Type)
 ```
 
+
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/clustertriggerbindings.md
```diff
@@ -48,3 +47,4 @@ spec:
       template:
         name: pipeline-template
 ```
+
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/eventlisteners.md
```diff
@@ -41,6 +41,38 @@ the following fields:
 [kubernetes-overview]:
   https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields
 
+### ServiceAccountName
```
gabemontero: I moved this section up since we now refer to it when discussing the trigger sa override in the next section

docs/eventlisteners.md
```diff
@@ -199,7 +241,6 @@ if desired. The response body and headers of the last Interceptor is used for
 resource binding/templating.
 
 <!-- FILE: examples/eventlisteners/eventlistener-interceptor.yaml -->
-
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/eventlisteners.md
```diff
@@ -230,6 +271,7 @@ spec:
         name: pipeline-template
 ```
 
+
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/eventlisteners.md
```diff
@@ -275,6 +316,7 @@ spec:
...
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/eventlisteners.md
```diff
@@ -251,7 +293,6 @@ The body/header of the incoming request will be preserved in this Interceptor's
 response.
 
 <!-- FILE: examples/eventlisteners/github-eventlistener-interceptor.yaml -->
-
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/eventlisteners.md
```diff
@@ -364,12 +405,12 @@ spec:
         name: pipeline-template
 ```
 
+
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/eventlisteners.md
```diff
@@ -364,12 +405,12 @@ spec:
...
 If no filter is provided, then the overlays will be applied to the body. With a
 filter, the `expression` must return a `true` value, otherwise the request will
 be filtered out.
 
 <!-- FILE: examples/eventlisteners/cel-eventlistener-no-filter.yaml -->
-
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/eventlisteners.md
```diff
@@ -390,6 +431,7 @@ spec:
         name: pipeline-template
 ```
 
+
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/triggerbindings.md
```diff
@@ -6,7 +6,6 @@ parameters. The separation of `TriggerBinding`s from `TriggerTemplate`s was
 deliberate to encourage reuse between them.
 
 <!-- FILE: examples/triggerbindings/triggerbinding.yaml -->
-
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/triggerbindings.md
```diff
@@ -22,6 +21,7 @@ spec:
     value: $(header.Content-Type)
 ```
 
+
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/triggertemplates.md
```diff
@@ -5,7 +5,6 @@ A `TriggerTemplate` is a resource that can template resources.
 **anywhere** within the resource template.
 
 <!-- FILE: examples/triggertemplates/triggertemplate.yaml -->
-
```
gabemontero: this stemmed from running `hack/update-docs.sh`

docs/triggertemplates.md
```diff
@@ -47,6 +46,7 @@ spec:
             value: $(params.gitrepositoryurl)
 ```
 
+
```
gabemontero: this stemmed from running `hack/update-docs.sh`

gabemontero:
integrations passed once ... fixed goimports ... running again

gabemontero:
OK everything is green

bump @wlynch @khrm PTAL

@dibyom - hey, @wlynch mentioned he had imminent vacation in the last sync call;  I'm going to

/assign @dibyom 

in case it makes sense and you have cycles to cover for him while he is out

of course make assignment adjustments from my current set here as you all see fit 

thanks

gabemontero:
OK I've rebased the PR to address the merge conflict that arose over the weekend.

dibyom:
Thanks @gabemontero . Taking a look later today!

docs/clustertriggerbindings.md
dibyom: yeah, there is some weird new line behavior between `./hack/update-docs.sh` and the `prettier` bot

docs/eventlisteners.md
```diff
@@ -65,38 +97,48 @@ triggers:
...
+    bindings:
+      - name: pipeline-binding
+      - name: message-binding
+    template:
+      name: pipeline-template
+``` 
 
-<!-- FILE: examples/role-resources/triggerbinding-roles/role.yaml -->
+*NOTE*: When employing per trigger service accounts, if the specified service account resides in a namespace that differs
+from the namespace of the `EventListener`, you'll need apply cluster level access for service accounts and secrets
```
dibyom: Even if it is on the same namespace, the EL will still need access to the `ServiceAccount`s right?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -67,6 +67,15 @@ type EventListenerTrigger struct {
 	// +optional
 	Name         string              `json:"name,omitempty"`
 	Interceptors []*EventInterceptor `json:"interceptors,omitempty"`
+	// ServiceAccount optionally associates credentials with each trigger;
+	// more granular authorization for
+	// who is allowed to utilize the associated pipeline
+	// vs. defaulting to whatever permissions are associated
+	// with the entire EventListener and associated sink facilitates
+	// multi-tenant model based scenarios
+	// TODO do we want to restrict this to the event listener namespace and just ask for the service account name here?
```
dibyom: Yeah, having a cluster role that has access to get all serviceAccounts seems like a lot of permissions for a EventListener

pkg/resources/create.go
```diff
@@ -99,6 +101,9 @@ func Create(logger *zap.SugaredLogger, rt json.RawMessage, triggerName, eventID,
 	logger.Infof("For event ID %q creating resource %v", eventID, gvr)
 
 	if _, err := dc.Resource(gvr).Namespace(namespace).Create(data, metav1.CreateOptions{}); err != nil {
+		if kerrors.IsUnauthorized(err) || kerrors.IsForbidden(err) {
+			return err
```
dibyom: Nice!

pkg/sink/sink.go
```diff
@@ -105,6 +114,13 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 	code := http.StatusAccepted
 	for i := 0; i < len(el.Spec.Triggers); i++ {
 		thiscode := <-result
+		// current take - if someone is doing unauthorized stuff, we abort immediately;
+		// unauthorized should be the final status code vs. the less than comparison
```
dibyom: sounds good to me...

docs/eventlisteners.md
gabemontero: ok I thought I saw somewhere it already have that but I'm probably mistaken

I'll make this more generic.

pkg/apis/triggers/v1alpha1/event_listener_types.go
gabemontero: It could be role / role bindings for SAs in specific namespaces vs. a cluster role etc. but same basic notion.  It would be consistent with what is their for the service account and the event listener.

Looking back part of what gave me pause was the gitlab/github webhook interceptors had secret refs that allow the setting of namespaces.

```
// SecretRef contains the information required to reference a single secret string
// This is needed because the other secretRef types are not cross-namespace and do not
// actually contain the "SecretName" field, which allows us to access a single secret value.
type SecretRef struct {
	SecretKey  string `json:"secretKey,omitempty"`
	SecretName string `json:"secretName,omitempty"`
	Namespace  string `json:"namespace,omitempty"`
}

```

I don't quite follow the comment, and have not discovered through code exploration why cross namespace is needed with those.

Can you clarify @dibyom ?

gabemontero:
> Will finish going through the PR tomorrow. Overall, really interesting idea..
> -- one question: do we validate at the controller level if the user has access to the SA that they enter in the trigger SA override field? Even if we don't is that a privilege escalation risk?

I think I know where you are headed @dibyom but a quick clarification is need.

By user accessing, do you mean
a) a user creating the `EventListenerTrigger` and assigning an SA to it?, or
b) some notion of a user associated with an event coming through the event sink, and then getting associated with a trigger as a result
c) or both

I'm in danger of tl;dr here ... but a high level summary for now, with more dialogue in the call for sure.

At a high level, a) is plausible and kind in place by default *I believe* if/when we remove the namespace element to the SA ref in the `EventListenerTrigger` (minimally pending my question above on that point vs. the webhook secret refs).  Though if you are worried about 
*specific* SAs in a namespace, then we are talking admission controller/webhooks to validate (though this would necessitate a 1 to 1 between actual users and service accounts).  If the `EventListenerTrigger` SA crosses namespace boundaries, *I believe* we are back to admission controller/webhooks SARs in the trigger controllers that manage the trigger specific CRDs

I don't see b)/c) working, as I remember how github/gitlab webhooks work, much less the other webhooks, which I am less familiar with.  And most likely you are talking about converting the event sink to an aggregated k8s apiserver.  I'm assuming that is not a consumable change.  But again we can discuss.

> 
> The use case for finer grained auth is definitely worth discussing. Will add this to tomorrow's WG agenda.

docs/eventlisteners.md
gabemontero: Furthermore @dibyom ... aside from your point in this thread, based on our resolutions to the questions below this doc IMO will need some more beef to it one way or the other around multi-tenant, etc. ... I'll do that once we get consensus on the open questions.

pkg/apis/triggers/v1alpha1/event_listener_types.go
dibyom: Yeah, I think I've only used secrets in the same namespace as well...Maybe we can get rid of the `Namespace` from those refs as well

khrm: Isn't this use when Trigger is in different namespace? Then secret is generally also in the Trigger Namespace. Otherwise, EL ns is used.

khrm: https://github.com/tektoncd/triggers/blob/master/pkg/interceptors/interceptors.go#L33

gabemontero: Yep thanks for the pointer @khrm 

And the next step for that is to do the signature validation so that the secret contents in the payload from github/gitlab that have arrived through the event sink match up ...i.e. 
https://github.com/tektoncd/triggers/blob/master/pkg/interceptors/github/github.go#L71

So ... characterizing "current support"
- we allow that secret to reside in a separate namespace from the `EventListener` and `EventListenerTrigger` namespace.
- meaning the `EventListener` SA today needs to be able to read `Secret` objects from namespaces other than its own.

What is unclear to me ... and I see no explanation of this in https://github.com/tektoncd/triggers/blob/master/docs/eventlisteners.md , is what scenario is served by letting this secret reside in namespaces besides the `EventListener` / `EventListenerTrigger`.  

Is it that the pipelines/tasks/pipelineruns/taskruns referenced by the trigger may be in the namesapce this secret resides in, and we are assuming the `EventListener` SA has permission to create those object in those namespaces?  And because of that, it makes sense for the owner of those to create the github/gitlab secret in the same namespace as the pipelines/task/pipelineruns/taskruns ?

If so, the doc really needs to clarify that.

If it is something else, please explain.

Now, on to this PR, and what we do with it, along with your feature https://github.com/tektoncd/triggers/issues/456 

I think I've come to this understanding.  That feature, and this PR, are around creating/access pipeline/tasks/pipelineruns/taskruns in a namespace that differs from the `EventListener` namespace **WITHOUT** adding permissions to create/access pipeline/tasks/pipelineruns/taskruns in that namesapce to the `EventListener` SA (in addition to being able to create those in the event listener's namespace).

Is that correct?  If so, yes this feature facilitates that.  If that is not correct, please clarify.

But per our discussion, the current work around is to create different `EventListener` instances in each namespace.  Where the discussion today speculated that isn't "too expensive".

Can you elaborate beyond what I mentioned today why that work around would be untenable for you?

thanks

pkg/resources/create.go
gabemontero: by the way @dibyom IMO we need this change here regardless of whether we add finer grained auth at the trigger level

thoughts?

pkg/sink/sink.go
gabemontero: same situation here re: https://github.com/tektoncd/triggers/pull/454#discussion_r387717992 @dibyom

gabemontero:
I went back and dug up old notes and got some memory refreshes from colleagues around the pain of being liberal in creating namespaces to effectively quarantine users with different sets of permissions.

Aside from the "creating a namespace is a cluster admin level event for shops and is bottlenecked" notion I mentioned on our last call, there is an etcd cost wrt namespaces, in particular around to cost of the set of secrets that get created by default along with the namespace.  Based on feature sets enabled the number of secrets can reach double digits.  Coupled with larger clusters, we've seen friction with etcd storage in the past.

pkg/apis/triggers/v1alpha1/event_listener_types.go
khrm: We are thinking of having only a single deployment for EventListener to reduce number of pods related to EventListener in a cluster. So that's why we give cluster role permission to EL to act as multitenant EL.

gabemontero: OK so the motivation is the concern over the cost of EventListeners, which was touch upon during the last sync call.  Perhaps we can revisit it again on today's call.

gabemontero:
> I went back and dug up old notes and got some memory refreshes from colleagues around the pain of being liberal in creating namespaces to effectively quarantine users with different sets of permissions.
> 
> Aside from the "creating a namespace is a cluster admin level event for shops and is bottlenecked" notion I mentioned on our last call, there is an etcd cost wrt namespaces, in particular around to cost of the set of secrets that get created by default along with the namespace. Based on feature sets enabled the number of secrets can reach double digits. Coupled with larger clusters, we've seen friction with etcd storage in the past.

And in case it was not clear from my last comment, with concern over namespace crawl, using a single namespace for multi-tenancy, admins would modify their RBAC add specific `resourceNames` to given Roles, and hence have Roles not just on Pipeline/PipelineRun/Task/TaskRun etc. in general, but specific instances of those.  And then associate different service accounts with those fine grained Roles.  And then set those Roles on EventListenerTriggers with the changes from this PR.

pkg/resources/create.go
dibyom: I totally agree!

pkg/apis/triggers/v1alpha1/event_listener_types.go
dibyom: Yes, this seems worth a discussion in today's call! :+1:

pkg/sink/sink.go
dibyom: Again, agreed!

dibyom:
> And in case it was not clear from my last comment, with concern over namespace crawl, using a single namespace for multi-tenancy, admins would modify their RBAC add specific resourceNames to given Roles, and hence have Roles not just on Pipeline/PipelineRun/Task/TaskRun etc. in general, but specific instances of those. And then associate different service accounts with those fine grained Roles. And then set those Roles on EventListenerTriggers with the changes from this PR.

Thanks! That makes it much clearer to me! Associating a Role to a specific set of Pipeline/Tasks seems like an effective way of doing RBAC. Would this be an alternative to the SAR checks in Pipelines or do they work together?

Also, would having two EventListeners in the same namespace but each having different service accounts be an alternative to this (I understand this method would have other concerns e.g. more pods etc.)

gabemontero:
> > And in case it was not clear from my last comment, with concern over namespace crawl, using a single namespace for multi-tenancy, admins would modify their RBAC add specific resourceNames to given Roles, and hence have Roles not just on Pipeline/PipelineRun/Task/TaskRun etc. in general, but specific instances of those. And then associate different service accounts with those fine grained Roles. And then set those Roles on EventListenerTriggers with the changes from this PR.
> 
> Thanks! That makes it much clearer to me! Associating a Role to a specific set of Pipeline/Tasks seems like an effective way of doing RBAC. Would this be an alternative to the SAR checks in Pipelines or do they work together?

They would work together.  The SAR check also has granularity down to the instance name.  This type is part of the SAR spec:

```
// ResourceAttributes includes the authorization attributes available for resource requests to the Authorizer interface
type ResourceAttributes struct {
	// Namespace is the namespace of the action being requested.  Currently, there is no distinction between no namespace and all namespaces
	// "" (empty) is defaulted for LocalSubjectAccessReviews
	// "" (empty) is empty for cluster-scoped resources
	// "" (empty) means "all" for namespace scoped resources from a SubjectAccessReview or SelfSubjectAccessReview
	// +optional
	Namespace string `json:"namespace,omitempty" protobuf:"bytes,1,opt,name=namespace"`
	// Verb is a kubernetes resource API verb, like: get, list, watch, create, update, delete, proxy.  "*" means all.
	// +optional
	Verb string `json:"verb,omitempty" protobuf:"bytes,2,opt,name=verb"`
	// Group is the API Group of the Resource.  "*" means all.
	// +optional
	Group string `json:"group,omitempty" protobuf:"bytes,3,opt,name=group"`
	// Version is the API Version of the Resource.  "*" means all.
	// +optional
	Version string `json:"version,omitempty" protobuf:"bytes,4,opt,name=version"`
	// Resource is one of the existing resource types.  "*" means all.
	// +optional
	Resource string `json:"resource,omitempty" protobuf:"bytes,5,opt,name=resource"`
	// Subresource is one of the existing resource types.  "" means none.
	// +optional
	Subresource string `json:"subresource,omitempty" protobuf:"bytes,6,opt,name=subresource"`
	// Name is the name of the resource being requested for a "get" or deleted for a "delete". "" (empty) means all.
	// +optional
	Name string `json:"name,omitempty" protobuf:"bytes,7,opt,name=name"`
}
```

> 
> Also, would having two EventListeners in the same namespace but each having different service accounts be an alternative to this (I understand this method would have other concerns e.g. more pods etc.)

Yes that is right, both as an alternative as well as the main pros/cons of such an approach.

docs/eventlisteners.md
gabemontero: OK circled back to this and yep I was mistaken ... `ServiceAccounts` are not in the `ClusterRole` as defined at https://github.com/tektoncd/triggers/blob/master/config/200-clusterrole.yaml

What I am now thinking I'll do is update https://github.com/tektoncd/triggers/blob/master/config/200-clusterrole.yaml as part of keeping the current form of this PR with adding auth at the `EventListenerTrigger` level, and add read level access to service accounts, and nuke this portion of the doc update along with adding the overall multi-tenancy piece with pros/cons, etc.

gabemontero:
OK I've pushed the clusterrole update and and initial broader multi-tenant doc update, including pros/cons on the different way of solving the problem, which IMO justifies the addition of the SA on the EventListenerTrigger.

@dibyom and other folks subscribed here PTAL

gabemontero:
Hmm ... I *think* the build-tests just timed out

```
I0312 21:08:13.930] -----------------------------------------
I0312 21:08:13.931] ---- Checking for forbidden licenses ----
I0312 21:08:13.931] -----------------------------------------
I0312 21:08:34.348] 2020/03/12 21:08:34 No errors found.
I0312 21:10:58.690] level=error msg="Timeout exceeded: try increase it by passing --timeout option"
I0312 21:10:59.011] Diffing against the documentation yaml
I0312 21:10:59.407] ============================
I0312 21:10:59.408] ==== BUILD TESTS FAILED ====
I0312 21:10:59.409] ============================
W0312 21:10:59.511] Run: ('/workspace/./test-infra/jenkins/../scenarios/../hack/coalesce.py',)
E0312 21:11:01.255] Command failed
I0312 21:11:01.257] process 476 exited with code 1 after 10.5m
E0312 21:11:01.258] FAIL: pull-tekton-triggers-build-tests
```

trying again

/test pull-tekton-triggers-build-tests

gabemontero:
yep it passed this time ... all tests green

docs/eventlisteners.md
```diff
@@ -9,13 +9,73 @@ Tekton resources. In addition, EventListeners allow lightweight event processing
...
+If you create each of those EventListeners in their own namespace, it becomes easy to assign 
+varying permissions to the ServiceAccount of each one to serve your needs.  And often times namespace
+creation is coupled with a default set of ServiceAccounts and Secrets that are also defined.
+So conceivably some administration steps are taken care of.  You just update the permissions
+of the automatically created ServiceAccounts.
+
+Possible drawbacks:
+- Namespaces with associated Secrets and ServiceAccounts in an aggregate sense prove to be the most expensive
+items in Kubernetes underlying `etcd` store.  In larger clusters `etcd` storage capacity can become a concern.
+- Multiple EventListeners means multiple HTTP ports that must be exposed to the external entities accessing
```
dibyom: Could they not be exposed using the same Ingress?

gabemontero: there is still a segregation problem ... I think ... but let's discuss

gabemontero: Pending that discussion, yes, some further spelling out of the flow might be required here.

config/200-clusterrole.yaml
```diff
@@ -17,6 +17,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: tekton-triggers-admin
 rules:
+  - apiGroups: [""]
+    resources: ["serviceaccounts"]
+    verbs: ["get", "list", "watch"]
```
gabemontero: per the conference call this may be the wrong spot ... will need to look some more, unless a reviewer has a pointer before I find it.

docs/eventlisteners.md
gabemontero: per conf call, I'll note Ingress solution (with some sort of mild verbiage on the user having to decide if Ingress is a barrier for them)

config/200-clusterrole.yaml
dibyom: All of our examples event listeners use the example service account here: https://github.com/tektoncd/triggers/tree/master/examples/role-resources

gabemontero: ah thanks @dibyom !

docs/eventlisteners.md
gabemontero: update to clarify port/firewall/ingress pushed

config/200-clusterrole.yaml
gabemontero: the undoing of this change and the updates to https://github.com/tektoncd/triggers/tree/master/examples/role-resources to add read access to serviceaccounts pushed

gabemontero:
OK another rebase based on recent changes causing conflicts pushed

@dibyom @ncskier - with a scenario documented in #456 what are the final steps needed for possibly tagging this PR for merge?

thanks

gabemontero:
> OK another rebase based on recent changes causing conflicts pushed
> 
> @dibyom @ncskier - with a scenario documented in #456 what are the final steps needed for possibly tagging this PR for merge (assuming my merges for conflicts pass all the tests)?

OK tests all green again

> 
> thanks

gabemontero:
Need another rebase ...

gabemontero:
Perhaps another "scenario", I also have some usage experience from other CI systems which I believe have bearing here.

I certainly don't think it makes sense to reference it in the doc updates in any way.  And I thought I mentioned it on prior calls, but if need be, I can further detail / reiterate in either forum.  

Perhaps a discussion point for today's call.

gabemontero:
latest rebase up / tests clean

gabemontero:
`level=error msg="Timeout exceeded: try increasing it by passing --timeout option"`

/retest

dibyom:
Thanks for adding this @gabemontero Appreciate your patience with this PR!
One thing that would be really cool to add would be a full example of an EventListener using two different service accounts (e.g. like the [Gitlab one](https://github.com/tektoncd/triggers/tree/master/examples/gitlab))

gabemontero:
> Thanks for adding this @gabemontero Appreciate your patience with this PR!

My pleasure @dibyom and likewise appreciate the patience and openness wrt the PR.  Authorization related matters are certainly not to be taken lightly.

> One thing that would be really cool to add would be a full example of an EventListener using two different service accounts (e.g. like the [Gitlab one](https://github.com/tektoncd/triggers/tree/master/examples/gitlab))

I would agree.  Poking around the examples, it seems like we would add the new SA and a more precise/specific set of roles under https://github.com/tektoncd/triggers/tree/master/examples/role-resources , and then provide an alternative to the example you noted accordingly, agree ?

I'm thinking I'm going to circle back with @khrm @siamaksade and @vdemeester on my end and see if we want to do a short term example independent of @khrm 's upcoming work, or if that might land reasonably soon, and use that instead to drive the example.

One of us will report back once we've had a chance to confer.

thanks again

dibyom:
> I would agree. Poking around the examples, it seems like we would add the new SA and a more precise/specific set of roles under https://github.com/tektoncd/triggers/tree/master/examples/role-resources , and then provide an alternative to the example you noted accordingly, agree ?

Yeah, we'd need a couple of more specific roles/SAs. They could either live with the example itself or in the role-resources folder...the role resources folder is meant for more of the "common" resources that are shared between multiple examples.

> I'm thinking I'm going to circle back with @khrm @siamaksade and @vdemeester on my end and see if we want to do a short term example independent of @khrm 's upcoming work, or if that might land reasonably soon, and use that instead to drive the example.

Sounds good to me! Thank you!

Issue #458, opened 2020-02-26
Header key should be treated as case-insensitive


akihikokuroda:
# Expected Behavior

In the TriggerBinding, 

```
params:
-name: github_event
  value: $(header.X-GitHub-Event)
```
should put the GitHub event (like push, pull_request) to the github_event parameter.

# Actual Behavior

This params definition causes failure on the receiving of the GitHub event.  It can not find the header.

# Steps to Reproduce the Problem

1. create eventlistener with the TriggerBinding with the parameter definition
2. receive an event from the GetHub
3.

# Additional Info

The header key is correct in the GitHub document .  It is `X-GitHub-Event`.  But in the eventlistener, it matches to `X-GitHub-Event`.
The go net/http library canonicalizes the key before it passes the headers to the eventlistener.(https://godoc.org/net/http#Header)
According to [RFC](https://tools.ietf.org/html/rfc7230#section-3.2), the header key should be case-insensitive.

PR #460, opened 2020-02-27
header key, case-insensitive match

akihikokuroda:
# Changes

This resolve https://github.com/tektoncd/triggers/issues/458

Issue:
The keys in the request header are converted by `CanonicalMIMEHeaderKey` before they are passed to the event sink code.  So the cases of the key string are not same as sent from the source.  This causes failure of matching header key.  The `header.X-GitHub-Event` was not found because it was converted to `X-Github-Event`. 

Change:
The header keys specified in the TriggerBinding are converted with the same function before it is searched in the headers.  

# Submitter Checklist

review them:

# Release Notes

```
- Bug fixes
The header key in the TriggerBinding is case insensitive
```

### DIFF
--- a/docs/triggerbindings.md
+++ b/docs/triggerbindings.md
@@ -34,7 +34,7 @@ Each parameter has a `name` and a `value`.
 ## Event Variable Interpolation
 
 TriggerBindings can access values from the HTTP JSON body and the headers using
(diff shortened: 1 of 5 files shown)


akihikokuroda:
I added a statement and an example in the triggerbindings.md

Issue #459, opened 2020-02-27
Problem upgrading to v0.3.0


ncskier:
# Expected Behavior
I expect to be able to upgrade to v0.3.0 without changing my EventListener.

# Actual Behavior
When upgrading to v0.3.0, the EventListener no longer works properly. When it receives an event, there seem to be errors extracting event data.

Here's an example error:
```
Get error when upgrading: {"level":"error","logger":"eventlistener","caller":"sink/sink.go:153","msg":"couldn't create resource with group version kind \"tekton.dev/v1alpha1, Resource=pipelineresources\": namespaces \"$(params.webhooks-tekton-target-namespace)\" not found","knative.dev/controller":"eventlistener","/triggers-eventid":"g8pzc","/trigger":"ghe-test-triggers-3-push-event","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.processTrigger\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:153\ngithub.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent.func1\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:95"}
```

I think that I also saw a case when the $(params.example) were not filled in the TriggerTemplate.

I've found that these errors go away after deleting and recreating the EventListener resource. However, we have not announced this as a breaking change. Also, I think that we should fix this unless there's a good reason why it is breaking.

# Steps to Reproduce the Problem

1. Install Triggers v0.2.1
2. Create a working EventListener + TriggerBinding + TriggerTemplate with parameters being extracted from the event.
3. Install Triggers v0.3.0
4. Send an HTTP event to the EventListener and see if everything works properly

akihikokuroda:
Issue is missing `b.kind` in the TriggerBindings

dibyom:
Yeah, we set the default in the reconciler but never persist it. And the ResolveTrigger function in the sink expects a Kind to be set

PR #461, opened 2020-02-27
Resolve TriggerBindings with no Kind set in sink

dibyom:
# Changes

When upgrading from 0.2.1 to 0.3, the Sink might have
to process a Binding that has no `Kind` set. This is
because, we only persist the default Kind when the Binding
object is created/updated. The reconciler also sets the defaults
by calling `SetDefaults` but that change is not persisted to etcd.

In the future, we might want to consider updating the object from the
reconciler itself if any default values were set.

Fixes: #459

# Submitter Checklist

review them:

### DIFF
--- a/pkg/template/resource.go
+++ b/pkg/template/resource.go
@@ -56,9 +56,7 @@ func ResolveTrigger(trigger triggersv1.EventListenerTrigger, getTB getTriggerBin
 				return ResolvedTrigger{}, fmt.Errorf("error getting ClusterTriggerBinding %s: %w", b.Name, err)
 			}
 			ctb = append(ctb, ctb2)
-		}
-
-		if b.Kind == triggersv1.NamespacedTriggerBindingKind {
+		} else {
 			tb2, err := getTB(b.Name, metav1.GetOptions{})
 			if err != nil {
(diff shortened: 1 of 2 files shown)


ncskier:
This might be a problem if a user does not add the `kind` field for their bindings, and then tries to update the EventListener, right? https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_validation.go#L53
```go
if b.Kind != NamespacedTriggerBindingKind && b.Kind != ClusterTriggerBindingKind {
	return apis.ErrInvalidValue(fmt.Errorf("invalid kind"), fmt.Sprintf("bindings[%d].kind", i))
}
```

dibyom:
> This might be a problem if a user does not add the kind field for their bindings, and then tries to update the EventListener, right? https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_validation.go#L53

I don't think so. For any newly created/updated bindings, we'll set the right defaults so that validation branch should never be hit. The issue is only for old bindings that are not updated.

dibyom:
Also, the build failures are real, still testing this out :D

PR #490, opened 2020-03-19
Allow creating v1beta1 resources

dibyom:
# Changes

- Update examples to use v1beta1 by default.
- Add an explicit v1alpha1 example.
- Use v1beta1 Pipeline types for Params and tests.

Fixes #481

Followup cleanup: #494

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/triggers/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

```
Triggers can now create both v1alpha1 and v1beta1 Pipeline resources.

```

PR #497, opened 2020-03-23
Switch API group for triggers to triggers.tekton.dev 

### LINKED ISSUE #332: Proposal: Change API Group for triggers
While experimenting on tektoncd/pipeline#1773, I experimented with api groups (`tekton.dev`) and [commented out](https://github.com/tektoncd/pipeline/pull/1773#issuecomment-568018607) that, I wished, triggers would have a different API group than `tektoncd/pipeline`.

>  (make it easier to *detect* the precense of trigger without having to know which types it exposes — right now.. ` | grep tekton.dev` returns pipeline and triggers :stuck_out_tongue:).

It might be a bit late to change `tektoncd/pipeline` api group, but I feel we could do it for `tektoncd/triggers` given we are still in 0.1 :angel:.

The proposal would be to use `triggers.tekton.dev` as APIGroup instead of `tekton.dev` (making it easier to detect by looking at existing APIs).

@vtereso @dibyom @bobcatfish @iancoffey @wlynch wdyt ?

/area api
/kind question
/kind feature

dibyom:
# Changes

BREAKING CHANGE: Users will have to delete, and recreate
resources after updating its API group.

Fixes #332

TODO:
- [x]  Fix e2e test
- [x] #490 

# Submitter Checklist

review them:

# Release Notes

```
All Triggers resources now use the triggers.tekton.dev API Group.

BREAKING CHANGE: Users will have to delete, and recreate
resources after updating its API group.
```


dibyom:
/hold

Ready for review but need to validate and document upgrade paths

dibyom:
Tried this out locally. 
* Users will have to change the role resources for their ELs on their own.
* Will write up a script in a followup that can change any existing old types to the new ones and then delete the old CRDs.

/hold cancel

docs/getting-started/ingress-run.yaml
```diff
@@ -1,11 +1,12 @@
-apiVersion: tekton.dev/v1beta1
+apiVersion: triggers.tekton.dev/v1beta1
 kind: TaskRun
 metadata:
   name: create-ingress-run
   namespace: getting-started
 spec:
   taskRef:
     name: create-ingress
+  inputs:
```
ncskier: I think that since this is a v1beta1 resource, it doesn't have the `inputs:` line.

docs/getting-started/ingress-run.yaml
```diff
@@ -1,11 +1,12 @@
-apiVersion: tekton.dev/v1beta1
+apiVersion: triggers.tekton.dev/v1beta1
```
ncskier: TaskRuns are a still a `tekton.dev` resource 😉

docs/getting-started/rbac/admin-role.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   namespace: getting-started
 rules:
 - apiGroups:
-  - tekton.dev
+  - triggers.tekton.dev
```
ncskier: I think we'll need to move the `- pipelineresources` here into a separate `apiGroup` for `tekton.dev`.

docs/getting-started/webhook-run.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: tekton.dev/v1beta1
+apiVersion: triggers.tekton.dev/v1beta1
```
ncskier: I think this should still be `tekton.dev` for the TaskRun.

test/eventlistener_test.go
```diff
@@ -383,18 +385,6 @@ func TestEventListenerCreate(t *testing.T) {
...
-	t.Logf("Deleting cluster-scoped resources")
-	if err := c.KubeClient.RbacV1().ClusterRoles().Delete("my-role", &metav1.DeleteOptions{}); err != nil {
-		t.Errorf("Failed to delete clusterrole my-role: %s", err)
-	}
-	if err := c.KubeClient.RbacV1().ClusterRoleBindings().Delete("my-rolebinding", &metav1.DeleteOptions{}); err != nil {
-		t.Errorf("Failed to delete clusterrolebinding my-rolebinding: %s", err)
-	}
-	if err := c.TriggersClient.TektonV1alpha1().ClusterTriggerBindings().Delete("my-clustertriggerbinding", &metav1.DeleteOptions{}); err != nil {
-		t.Errorf("Failed to delete clustertriggerbinding my-clustertriggerbinding: %s", err)
-	}
```
ncskier: I think that these test cleanup changes belong in a separate PR. There're so many changed files to dig through in this PR, I don't want these changes to get lost in all of it.

docs/getting-started/ingress-run.yaml
dibyom: good catch...we should start running these tests again.

docs/getting-started/ingress-run.yaml
dibyom: dome

docs/getting-started/rbac/admin-role.yaml
dibyom: moved

test/eventlistener_test.go
dibyom: fair point though I needed it to test it out locally. Opened #503

test/eventlistener_test.go
```diff
@@ -180,17 +180,19 @@ func TestEventListenerCreate(t *testing.T) {
...
-				},
-				{
-					APIGroups: []string{""},
-					Resources: []string{"configmaps"},
-					Verbs:     []string{"get", "list", "watch"},
-				},
+			Rules: []rbacv1.PolicyRule{{
+				APIGroups: []string{triggersv1.GroupName},
+				Resources: []string{"clustertriggerbindings", "eventlisteners", "triggerbindings", "triggertemplates"},
+				Verbs:     []string{"create", "get"},
```
ncskier: I think this only needs `"get"`. I think that we can delete `"create"`, because the TriggerTemplate doesn't create any Triggers resources.

docs/getting-started/rbac/admin-role.yaml
```diff
@@ -20,6 +19,7 @@ rules:
   - pipelineruns
   - pipelineresources
   verbs:
+  - get
```
ncskier: I'm not super familiar with the Getting Started guide, but I don't think that we need `- get` here, because I think that we only need to be able to `create` PipelineRuns and PipelineResources.

PR #500, opened 2020-03-25
Adds "commented" website front matter to docs

skaegi:
This change adds the front matter needed for the website without altering the site markdown or adding titles. It does this by adding the front matter in comment tags.

The tutorial and install are not part of the website currently so links to them under docs/ use absolute urls to allow them to work in both the website and in the pipeline repo

# Changes
update /docs markdown files with front matter

# Submitter Checklist

review them:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,5 +28,5 @@ See the following links for more on each of the resources involved:
 
 ## Getting Started Tasks
 
-- [Create an Ingress on the EventListener Service](https://github.com/tektoncd/triggers/blob/master/docs/create-ingress.yaml)
-- [Create a GitHub webhook](https://github.com/tektoncd/triggers/blob/master/docs/create-webhook.yaml)
+- [Create an Ingress on the EventListener Service](create-ingress.yaml)
+- [Create a GitHub webhook](create-webhook.yaml)
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,3 +1,12 @@
+<!--
+---
+title: "Triggers and EventListeners"
+linkTitle: "Triggers"
+weight: 3
+description: >
+  Event Triggers
+---
+-->
 # Tekton Triggers
 
 Triggers enables users to map fields from an event payload into resource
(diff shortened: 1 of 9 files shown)


docs/README.md
```diff
@@ -19,5 +28,5 @@ See the following links for more on each of the resources involved:
 
 ## Getting Started Tasks
 
-- [Create an Ingress on the EventListener Service](create-ingress.yaml)
-- [Create a GitHub webhook](create-webhook.yaml)
+- [Create an Ingress on the EventListener Service](https://github.com/tektoncd/triggers/blob/master/docs/create-ingress.yaml)
```
dibyom: Having the branch hard coded here would break links to any new versioned docs that we have here: https://github.com/tektoncd/triggers#read-the-docs

Do we need to keep the absolute urls forever or can that be changed  back when its part of the website?

skaegi: Definitely short term.
Let me see if I can workaround that as everything under docs should come as a unit. The "sync" is currently very simple and does not pick-up files that are not markdown.

skaegi: One option is to leave it as a broken link in the website until we can figure out the right thing to do. I have no problem err-ing on the side of keeping our current github doc pristine.

dibyom: Cool, let's keep the relative links for now then.

dibyom:
> I0325 15:52:21.743] pkg/apis/triggers/v1alpha1/event_listener_validation.go:65:31: S1039: unnecessary use of fmt.Sprintf (gosimple)
I0325 15:52:21.747] 		return apis.ErrMissingField(fmt.Sprintf("template.name"))


Not sure why the linter is failing for this in this PR 🤔

skaegi:
well... why not try another...
/retest

dibyom:
@skaegi I'll look into the linter issue....to unblock this PR would you mind making the lint change..it should be  simple...
pkg/apis/triggers/v1alpha1/event_listener_validation.go:65:31
`return apis.ErrMissingField(fmt.Sprintf("template.name"))` to
`return apis.ErrMissingField("template.name")`

PR #537, opened 2020-04-15
Implement parseJSON for strings.

bigkevmcd:
# Changes

This adds a CEL function "parseJSON" which can take a key and parse it into a dynamic map.

This is useful if you have embedded JSON inside a string in a request body.

# Submitter Checklist

review them:

# Release Notes

```
New CEL function `parseJSON` that can parse JSON embedded in a string in a request body.
```

### DIFF
--- a/docs/cel_expressions.md
+++ b/docs/cel_expressions.md
@@ -241,5 +241,18 @@ interceptor.
      <pre>header.canonical('X-Secret-Token').compareSecret('key', 'secret-name')</pre>
     </td>
   </tr>
-
+  <tr>
+    <th>
+     parseJSON
+    </th>
+    <td>
+      parseJSON(string) -> map
+    </td>
+    <td>
+     This parses a string that contains a JSON body into a map which which can be subsequently used in other expressions.
+    </td>
+    <td>
+     <pre>parseJSON('{"testing":"value"}').testing == "value"</pre>
+    </td>
+  </tr>
 </table>
--- a/pkg/interceptors/cel/cel.go
+++ b/pkg/interceptors/cel/cel.go
@@ -59,6 +59,8 @@ func NewInterceptor(cel *triggersv1.CELInterceptor, k kubernetes.Interface, ns s
 	}
 }
 
+var mapStrDyn = decls.NewMapType(decls.String, decls.Dyn)
+
 // ExecuteTrigger is an implementation of the Interceptor interface.
 func (w *Interceptor) ExecuteTrigger(request *http.Request) (*http.Response, error) {
 	env, err := makeCelEnv()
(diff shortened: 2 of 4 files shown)


tragiclifestories:
Thanks for this, solves some pain for us!

Just a drive-by comment - not all JSON values are objects. I wonder if perhaps the function should handle the various other cases predictably: 

```
"json_body": `"value"`
"json_body": `null`
"json_body": `5`
"json_body": `[{"foo": 5},{"foo":6}]`
```

Either that or have a name that clarifies that it only works for maps/objects. 

Most of these cases are degenerate, of course, but the array case is likely to have real use cases.

bigkevmcd:
> Either that or have a name that clarifies that it only works for maps/objects.
> 
> Most of these cases are degenerate, of course, but the array case is likely to have real use cases.

Yeah, I agree that maps are a possible use-case, not sure how we can identify these tho', other than as you're implying, growing a multitude of new functions to parse random strings as JSON.

What this does imply, is that perhaps a CEL function _isn't_ the right solution to this, and it might be better solved using a webhook interceptor?

jace-ys:
> Yeah, I agree that arrays are a possible use-case, not sure how we can identify these tho', other than as you're implying, growing a multitude of new functions to parse random strings as JSON.
> 
> What this does imply, is that perhaps a CEL function _isn't_ the right solution to this, and it might be better solved using a webhook interceptor?

I think that in the case of arrays, could we do a type assertion into either a `map[string]interface{}{}` or `[]map[string]interface{}{}` and return the appropriate CEL type?

But yeah it seems like a webhook interceptor would provide most control to the user and could be the ultimate solution nonetheless.

bigkevmcd:
> I think that in the case of arrays, could we do a type assertion into either a `map[string]interface{}{}` or `[]map[string]interface{}{}` and return the appropriate CEL type?

It's not as simple as that, we dictate the type of the value that the string is parsed into.

We could try with different types in the event of an error, but the code provides a map[string]interface{}{} to parse the "string" into.

bigkevmcd:
My main thinking behind a function to parse the `payload` body of the deployment hook event https://developer.github.com/v3/activity/events/types/#deploymentevent

This doesn't feel like an uncommon case,  and for the other cases cited, the deployment payload is user-controllable, and it feels like an object is the most useful.

tragiclifestories:
That's exactly our use case too - I'd certainly be happy to have the functionality in here. And Github will refuse to accept anything other than a string or an object for the deployment payload, so the other types are not relevant there. 

Also I suppose adding support for other JSON value types is a backwards compatible change so even if it was decided to add them at some point we need not have all of them in this PR?

dibyom:
@bigkevmcd I think you have to rebase and push for the EasyCLA Check to pass

vdemeester:
removed lgtm temporarly to unblock the merge queue

Issue #508, opened 2020-03-27
Confusing params syntax in trigger templates


afrittoli:
# Expected Behavior
`TriggerTemplates` allow defining a set of (Tekton) resources that should be provisioned.
`TriggerTemplates` expose parameters that can be injected into the resources using the `$(params.<param-name>)` syntax. The problem is that some of the resources that can be provisioned in `TriggerTemplates`  have their own parameters, and they can refer to them using the exact same syntax `$(params.<paramName>)`.

Resources that suffer from this issue are `TaskRuns` with a `taskSpec` embedded as `PipelineRuns` with a `pipelineSpec` embedded.

# Actual Behavior
It would be nice to have a way to distinguish parameters that come from the `TriggerTemplate`.
One way could be to prepend `tt` to params, e.g. `$(tt.params.<paramName>`.

# Steps to Reproduce the Problem

Example:
```
apiVersion: tekton.dev/v1alpha1
kind: TriggerTemplate
metadata:
  name: deploy-tekton-release
spec:
  params:
  - name: clusterResource
    description: Name of the cluster resource that points to the target cluster
  - name: targetCluster
    description: Name of the target cluster, used for the overlay
  - name: releaseBucket
    description: Base URL of the bucket where the release file is stored
    default: gs://tekton-releases
  - name: releaseFile
    description: Name of the release file
    default: release.yaml
  - name: releaseExtraPath
    description: Any extra path after bucket/project, starting with a "/"
    default: ""
  - name: tektonProject
    description: Name of the project to deploy (pipeline, triggers, dashboard)
  - name: tektonVersion
    description: Version of the project to deploy
  - name: namespace
    description: Target namespace. Not enforeced, for verification.
  - name: plumbingRepository
    description: URL of the repository that holds plumbing scripts
  - name: plumbingRevision
    description: Git revision of the repository that holds plumbing scripts
  resourcetemplates:
  - apiVersion: tekton.dev/v1beta1
    kind: TaskRun
    metadata:
      name: deploy-$(params.tektonProject)-release-$(params.targetCluster)-$(uid)
    spec:
      taskRef:
        name: install-tekton-release
      params:
      - name: projectName
        value: $(params.tektonProject)
      - name: namespace
        value: $(params.namespace)
      - name: version
        value: $(params.tektonVersion)
      - name: environment
        value: $(params.targetCluster)
      - name: release-file
        value: $(params.releaseFile)
      resources:
        inputs:
          - name: release-bucket
[...]

PR #548, opened 2020-04-21
Add Support for Embedding TriggerBindingSpec in EL

### LINKED ISSUE #371: Embed TriggerBinding in EventListener triggers
# Expected Behavior
I expect to be able to send static parameters to my TriggerTemplate without creating an extra TriggerBinding resource to define these parameters in.

# Actual Behavior
In v0.1.0, we had a `params` field in the EventListener:
```YAML
kind: EventListener
metadata:
  name: listener-interceptor
spec:
  triggers:
    - name: foo-trig
      binding:
        name: pipeline-binding
      template:
        name: pipeline-template
      params:
      - name: message
```
This field made it very easy for me to send static parameters to my TriggerTemplate. However, we removed it with the v0.2.0 release, because parameters were replaced with multiple TriggerBindings. So, now I must create a separate TriggerBinding resource with static parameters to use in my EventListener.
I expect to be able to define my static parameters in the EventListener. Having to create a separate TriggerBinding resource with my parameters is annoying.

I think that a nice solution to this problem is to implement the option to embed TriggerBindings in the EventListener. The implementation could be similar to how Tekton Pipelines handles embedding PipelineResources. They have a `resourceRef` or `resourceSpec` for each PipelineResource used in a TaskRun or PipelineRun ([more info here](https://github.com/tektoncd/pipeline/blob/master/docs/taskruns.md#providing-resources)). Similarly, we could have a `bindingRef` or `bindingSpec` for each TriggerBinding used in an EventListener.

khrm:
# Changes
Fixes #371

# Submitter Checklist

review them:

# Release Notes

```

-  Support Embedding in EventListenerSpec
   TriggerBindingSpec can be added to EventListener Trigger.

-  Add Ref for TriggerBinding  in EventListenerSpec Trigger
   In addition to name, Ref also need to be specified. 
   At present, if ref isn't given and spec is also not there, name is considered as Ref. This behaviour will be deprecated in future releases. 
```


dibyom:
Integration test failure is unrelated. Should be fixed by https://github.com/tektoncd/triggers/pull/559

test/builder/eventlistener.go
```diff
@@ -162,12 +162,13 @@ func EventListenerTriggerServiceAccount(saName, namespace string) EventListenerT
 }
 
 // EventListenerTriggerBinding adds a Binding to the Trigger in EventListenerSpec Triggers.
-func EventListenerTriggerBinding(name, kind, apiVersion string) EventListenerTriggerOp {
+func EventListenerTriggerBinding(name, kind, ref, apiVersion string, ops ...TriggerBindingSpecOp) EventListenerTriggerOp {
```
dibyom, line 165: can we just use the ref parameter instead of repeating both name and re?

third_party/cloud.google.com/go/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
dibyom, line 1: I think the license file changes should go away after a rebase... 🤔

docs/eventlisteners.md
```diff
@@ -149,7 +149,7 @@ The `triggers` field is required. Each EventListener can consist of one or more
 - `name` - (Optional) a valid
   [Kubernetes name](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)
 - [`interceptors`](#interceptors) - (Optional) list of interceptors to use
-- `bindings` - A list of names of `TriggerBindings` to use
+- `bindings` - A list of names of `TriggerBindings` to use, TriggerBindingsSpec can also be embedded.
```
dibyom, line 152:
```suggestion
- `bindings` - A list of references to `TriggerBindings` or embedded TriggerBindingsSpecs to use.
```

dibyom: Can we also provide a quick example snippet here showing a `ref` and an embedded spec?

pkg/template/resource_test.go
```diff
@@ -469,7 +469,7 @@ func Test_ResolveTrigger_error(t *testing.T) {
 		{
 			name: "error triggerbinding and triggertemplate",
 			trigger: bldr.Trigger("invalid-tt-name", "v1alpha1",
```
dibyom, line 471: Can we add a test that uses an embedded binding as well?

pkg/template/resource.go
```diff
@@ -50,6 +50,15 @@ func ResolveTrigger(trigger triggersv1.EventListenerTrigger, getTB getTriggerBin
 	tb := make([]*triggersv1.TriggerBinding, 0, len(trigger.Bindings))
 	ctb := make([]*triggersv1.ClusterTriggerBinding, 0, len(trigger.Bindings))
 	for _, b := range trigger.Bindings {
+		if b.Spec != nil {
+			tb = append(tb, &triggersv1.TriggerBinding{
```
dibyom, line 54: Don't we need to return out of  this loop here else  we'll try to fetch the binding in getTB and fail?

docs/clustertriggerbindings.md
```diff
@@ -48,8 +48,10 @@ spec:
       bindings:
         - name: pipeline-clusterbinding
           kind: ClusterTriggerBinding
+          ref: pipeline-clusterbinding
```
dibyom, line 51: Probably should specify only ref and no use name?

examples/eventlisteners/eventlistener-embedded-bindings.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: listener-embed-binding
+spec:
+  serviceAccountName: tekton-triggers-example-sa
+  triggers:
+    - name: foo-trig
+      bindings:
+        - name: pipeline-binding
+          ref: pipeline-binding
+        - name: message-binding
```
dibyom, line 13: Just an idea to consider (we don't have to do it in this PR) -- given that the bindings spec is relatively simple (just a list of name and values), I wonder if something like this would work:

```
 bindings:
        - ref: pipeline-binding
        - name: message
          value: "Hello from an embedded binding"
```

What I like about this is that it is a lot less verbose and more direct (we don't need the extra spec and params field as well as the name field (what value does the name field have for an embedded binding). Though this is different from other places where we embed resources so we should discuss this first before implementing.

docs/eventlisteners.md
khrm: I have given here: https://github.com/tektoncd/triggers/pull/548/commits/267b026ee58b3149d7cceba5bdb98a7b73a426ba#diff-be0d7f59d8d71d22c7dcc42f4be3eb3fR165

pkg/template/resource_test.go
khrm: Done.

pkg/template/resource.go
khrm: Used continue done.

examples/eventlisteners/eventlistener-embedded-bindings.yaml
khrm: Good idea. Would do in next PR.

khrm:
@dibyom 
I have kept name as an identifier to be used within EL irrespective of whether we have embedded or ref bindings. Atm, there doesn't seems to be a need for it but I can that we need to differentiate on how a param is coming from like here: https://github.com/tektoncd/triggers/issues/508
So we might need to do: tt.bindingname.param

dibyom:
> I have kept name as an identifier to be used within EL irrespective of whether we have embedded or ref bindings. Atm, there doesn't seems to be a need for it but I can that we need to differentiate on how a param is coming from like here: #508
So we might need to do: tt.bindingname.param

That's definitely a valid scenario though I'm not sure if we'd want to couple bindings within the template. At the moment though, keeping both `name` and `ref` seems confusing. Can we remove them from the examples at least? (we can keep/support the field but let's document and use `ref` as the default?). Should be simple enough to add back `name` if we need to later.

PR #556, opened 2020-04-28
Add recommended k8s labels

eddycharly:
This PR adds recommended k8s labels on resources.
Implementation for proposal https://github.com/tektoncd/pipeline/issues/2497

Related PR in pipelines: https://github.com/tektoncd/pipeline/pull/2501

Base: added labels and updated selectors for all resources in `/config`

NOTE: this will introduce a breaking change when trying to update deployments as label selectors have changed.

### DIFF
--- a/config/101-podsecuritypolicy.yaml
+++ b/config/101-podsecuritypolicy.yaml
@@ -16,6 +16,9 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: tekton-triggers
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-triggers
 spec:
   privileged: false
   allowPrivilegeEscalation: false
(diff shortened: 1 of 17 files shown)


config/101-podsecuritypolicy.yaml
```diff
@@ -16,6 +16,11 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: tekton-triggers
+  labels:
+    app.kubernetes.io/name: psp
```
vdemeester, line 20: Any reason for using `psp` instead of a "long" name ?

eddycharly: Shall i add the `app.kubernetes.io/name` label at all ?

eddycharly: I mean, we want to use label for two things:
- identify the application/component the resource belongs to
- be able to look up the resource when necessary

I don’t think we need to look up PSP anywhere and other labels are enough to identify the application so...

vdemeester: yeah, I don't think `app.kubernetes.io/name` has too much value indeed.

PR #574, opened 2020-05-12
Update eventlistener.yaml

posinaga-demo:
Using these examples currently fails with the error message below.  The current documentation in https://tekton.dev/docs/triggers/eventlisteners/ shows "name" as the attribute to set here instead of "ref".  NOTE: I didn't look at the other eventlistener examples exhaustively but it seems they all have the same problem.

Error from server (BadRequest): error when creating "test-event-listener.yaml": admission webhook "webhook.triggers.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "ref"

# Changes

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/examples/eventlisteners/eventlistener.yaml
+++ b/examples/eventlisteners/eventlistener.yaml
@@ -8,7 +8,7 @@ spec:
   triggers:
     - name: foo-trig
       bindings:
-        - ref: pipeline-binding
-        - ref: message-binding
+        - name: pipeline-binding
+        - name: message-binding
       template:
         name: pipeline-template


dibyom:
Hi @posinaga-demo this problem is likely due to a mismatch between the version of installed Triggers and the versions of docs. The `ref` field was introduced recently and will be part of the upcoming v0.5.0 release. We do have docs versioned at each release that you can follow instead: https://github.com/tektoncd/triggers#read-the-docs

dibyom:
I'm going to close this PR for now. Please feel free to re-open if the above suggestion did not help

PR #576, opened 2020-05-14
Refactored examples and added github example with README

savitaashture:
# Changes
Refactored examples and added github example along with README

# Submitter Checklist

review them:

### DIFF
--- a/examples/github/README.md
+++ b/examples/github/README.md
@@ -1,0 +1,46 @@
+## GitHub EventListener
+
+Creates an EventListener that listens for GitHub webhook events.
+
+### Try it out locally:
+
+1. To create the GitHub trigger and all related resources, run:
+
+   ```bash
+   kubectl apply -f examples/github/
+   ```
+
+1. Port forward:
+
+   ```bash
+   kubectl port-forward \
+    "$(kubectl get pod --selector=eventlistener=github-listener-interceptor -oname)" \
+     8080
+   ```
+
+   **Note**: Instead of port forwarding, you can set the
+   [`serviceType`](https://github.com/tektoncd/triggers/blob/master/docs/eventlisteners.md#serviceType)
+   to `LoadBalancer` to expose the EventListener with a public IP.
+
+1. Test by sending the sample payload.
+
+   ```bash
+   curl -v \
+   -H 'X-GitHub-Event: pull_request' \
+   -H 'X-Hub-Signature: sha1=0835c8c5dc317870c4e48659df5f3c53213cd348' \
+   -H 'Content-Type: application/json' \
+   -d '{"head_commit":{"id":"master"},"repository":{"url": "https://github.com/tektoncd/triggers"}}' \
+   http://localhost:8080
+   ```
+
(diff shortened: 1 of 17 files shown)


examples/github/README.md
```diff
@@ -0,0 +1,44 @@
...
+   -H 'X-GitHub-Event: pull_request' \
+   -H 'X-Hub-Signature: sha1=0835c8c5dc317870c4e48659df5f3c53213cd348' \
+   -H 'Content-Type: application/json' \
+   -d '{"head_commit":{"id":"master"},"repository":{"url": "https://github.com/tektoncd/triggers"}}' \
+   http://localhost:8080
+   ```
+
+   The response status code should be `201 Created`
+   
+   [`HMAC`](https://www.freeformatter.com/hmac-generator.html) tool used to create X-Hub-Signature
```
dibyom, line 38: Thanks for adding this! Might be overkill but we could mention that the "string" is the body and the secretKey is from the example secret?

examples/github/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+1. You should see a new PipelineRun that got created:
+
+   ```shell script
+   kubectl get pipelinerun
```
dibyom, line 43: Again, if we used our own pipeline/taskrun here, it would be easier to track what got created

examples/github/binding.yaml
```diff
@@ -0,0 +1 @@
+../role-resources/triggerbinding-roles/binding.yaml
```
dibyom, line 1: As I mentioned in #578 , I'd prefer adding the binding/template/pipeline/secret in this folder in the same file as it makes it much clearer what is going on (just look at this one file vs looking at 6-7 different files). I know this leads to some duplication but for examples I'm ok with having some repetition in order to help with clarity

dibyom:
Thanks for addressing the comments. Just two things:
1. Could we put the role/binding/serviceaccount file in one file (say, role.yaml)?
2. Can we squash the commits into one?
Otherwise LGTM!

savitaashture:
> Thanks for addressing the comments. Just two things:
> 
> 1. Could we put the role/binding/serviceaccount file in one file (say, role.yaml)?
> 2. Can we squash the commits into one?
>    Otherwise LGTM!

Thank you
1. Modified based on the suggestion
2. squashed commits

I have a question here is if we squash commits we actually loose the history of review comments fix :thinking: 
I know this is small PR but in case of big feature its actually make sense to keep the history of commits so that it will be easy for the reviewers to see actual fix.

dibyom:
> I have a question here is if we squash commits we actually loose the history of review comments fix 🤔
I know this is small PR but in case of big feature its actually make sense to keep the history of commits so that it will be easy for the reviewers to see actual fix.

Yeah, unfortunately GitHub code review does not handle diffs across commits well. And it is indeed helpful during review to see what comments are addressed. However, I think squashing still makes sense. Otherwise, the master branch gets cluttered with a bunch of small commits which makes it hard to rollback/cherry-pick (as well as figure out exactly what happened).

Issue #362, opened 2020-01-21
How to expose the EventListener to Bitbucket webhook


amlozano:
# Expected Behavior
I can call an EventListener from Bitbucket, authorising the call with a token

# Actual Behavior
I can configure a generic Webhook EventListener, but I don't find an easy way of exposing the EventListener to be called from BitBucket with a token Authentication. I can only think on exposing an interceptor that will handle the authentication, which requires a component to be coded from my side.

# Steps to Reproduce the Problem
No steps needed

# Additional Info
I was already able to trigger the pipeline from a generic webhook, doing port forwarding directly to the pod and calling it locally, but checking on how the GitLab and GitHub examples work, I would need an interceptor that checks for the token to be able to connect Bitbucket server with the webhook. Is there any plan to release a Bitbucket interceptor, or do I need to implement my own?

dibyom:
Hey @amlozano You are right that at the moment we do not have a built in BitBucket interceptor. That would definitely be a nice feature to add if you'd like to add it. We can follow the same patterns as the GitHub/Gitlab ones: https://github.com/tektoncd/triggers/blob/master/pkg/interceptors/github/github.go

amlozano:
Hi dibyom, I may have time to invest into the implementation, I have to check first since I am not really experienced with Golang and it might take too much time from my project to do that. I created the ticket to keep track of it at least, and if anyone with more experience wants to help, it would be awesome.

savitaashture:
Hi @amlozano 
Are you working on this feature?

amlozano:
@savitaashture no, sorry. I don’t find time now to work on this, plus we discarded going on with tekton for the moment. Maybe we will give it a try in the future.

savitaashture:
> @savitaashture no, sorry. I don’t find time now to work on this, plus we discarded going on with tekton for the moment. Maybe we will give it a try in the future.

@amlozano thank you for the conformation :+1:

sravankumar777:
Looking for this feature for our use case using tekton pipelines. 
Would like to listen to events from bitbucket webhook and trigger the pipeline for building & deployment operations. 
Would be great to see about it :) 
@savitaashture, thanks for working on this feature.

PR #578, opened 2020-05-14
Add BitBucket Interceptor with examples and README

savitaashture:
# Changes

1. Added BitBucket Interceptor
2. Added examples with README
3. Updated existing doc

**Fixes:** #362

# Submitter Checklist

review them:

# Release Notes

```
- Added BitBucket Interceptors
```


docs/eventlisteners.md
```diff
@@ -417,6 +418,48 @@ spec:
         name: pipeline-template
 ```
 
+### BitBucket Interceptors
+
+BitBucket Interceptors contain logic to validate and filter webhooks that come from
+BitBucket. Supported features include validating webhooks actually came from BitBucket as well as
```
jlpettersson, line 424: "Bitbucket Server" or "Bitbucket Cloud"? It is two different products (and implementations) as what I know. Or both? This should be specified in the documentation.

docs/eventlisteners.md
```diff
@@ -417,6 +418,48 @@ spec:
         name: pipeline-template
 ```
 
+### BitBucket Interceptors
+
+BitBucket Interceptors contain logic to validate and filter webhooks that come from
```
jlpettersson, line 423: nit: It look like the write "Bitbucket" instead of "BitBucket" ?

docs/eventlisteners.md
savitaashture: Both Bitbucket server and cloud provide Git repository management abilities.
Updated Doc accordingly 
Thank you

pkg/apis/triggers/v1alpha1/event_listener_validation.go
```diff
@@ -146,6 +149,11 @@ func (i *EventInterceptor) validate(ctx context.Context) *apis.FieldError {
 	//
 	// }
 
+	// No bitbucket validation required yet.
+	// if i.Bitbucket != nil {
+	//
+	// }
```
dibyom, line 155: Don't think this is necessary?

pkg/interceptors/bitbucket/bitbucket_test.go
```diff
@@ -0,0 +1,264 @@
...
+					t.Errorf("Interceptor.ExecuteTrigger() error = %v, wantErr %v", err, tt.wantErr)
+				}
+				return
+			}
+
+			got, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				t.Fatalf("error reading response body %v", err)
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
dibyom, line 259: hmm...looks like we are using `reflect.DeepEqual` in this package and `cmp.Diff` in others. Would be nice to use `cmp.Diff` everywhere (can be a separate cleanup PR in the future!)

pkg/sink/sink_test.go
```diff
@@ -928,3 +897,127 @@ func TestHandleEventWithInterceptorsAndTriggerAuth(t *testing.T) {
...
+
+	if resp.StatusCode != http.StatusCreated {
+		t.Fatalf("Response code doesn't match: expected status code %d vs. actual %d, entire statutes %v",
+			http.StatusCreated,
+			resp.StatusCode,
+			resp.Status)
+	}
+}
+
+func getResources(t *testing.T, value string) (*v1alpha1.TriggerBinding, *v1alpha1.TriggerTemplate) {
```
dibyom, line 989: nit: can we use a more descriptive name than value for the param?
Also, we should have a `t.Helper()` call in this function to indicate that this is a helper function

pkg/apis/triggers/v1alpha1/event_listener_validation.go
dibyom: Also, we should add validation for these 😄  (maybe in a future PR)

examples/bitbucket/README.md
```diff
@@ -0,0 +1,43 @@
+## Bitbucket EventListener
+
+Creates an EventListener that listens for Bitbucket webhook events.
+
+### Try it out locally:
+
+1. To create the Bitbucket trigger and all related resources, run:
+
+   ```shell script
```
dibyom, line 9: Should be `shell-script`? or just `bash`?
https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L4989

examples/bitbucket/README.md
```diff
@@ -0,0 +1,43 @@
...
+   kubectl apply -f examples/bitbucket/
+   ```
+
+1. Port forward:
+
+   ```shell script
```
dibyom, line 15: same as above

examples/bitbucket/bitbucket-eventlistener-interceptor.yaml
```diff
@@ -0,0 +1,20 @@
...
+    - name: bitbucket-triggers
+      interceptors:
+        - bitbucket:
+            secretRef:
+              secretName: githubsecret
+              secretKey: secretToken
+            eventTypes:
+              - repo:refs_changed
+      bindings:
+        - name: bitbucket-binding
```
dibyom, line 18: Can we put the binding and the template in this file as opposed being symlinks? (Like how we do in the GitLab example)?

examples/bitbucket/pipeline.yaml
```diff
@@ -0,0 +1 @@
+../example-pipeline.yaml
```
dibyom, line 1: If we use a new pipeline just for this with its own name instead of sharing one common pipeline, its a lot easier to find out which EL created the pipelinerun/taskrun i.e in the readme, we can say    `kubectl get pipelinerun| grep bitbucker-pr` or something

examples/bitbucket/secret.yaml
```diff
@@ -0,0 +1 @@
+../role-resources/secret.yaml
```
dibyom, line 1: Just like the pipeline, I'd prefer if each example that used a secret created and used its own.

examples/bitbucket/role.yaml
```diff
@@ -0,0 +1 @@
+../role-resources/triggerbinding-roles/role.yaml
```
dibyom, line 1: If we are symlinking the role resources, I think it would be nicer if we put the role, the rolebinding, and the serviceAccount in one file and symlink'd that single file instead of all 3.

pkg/interceptors/bitbucket/bitbucket_test.go
savitaashture: `reflect.DeepEqual` was used in 3 more test cases so handled here only

pkg/apis/triggers/v1alpha1/event_listener_validation.go
dibyom: Can we remove the commented out code?

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
```diff
@@ -347,8 +347,9 @@ func TestEventListenerValidate_error(t *testing.T) {
 					Bindings: []*v1alpha1.EventListenerBinding{{Name: "tb", Kind: v1alpha1.NamespacedTriggerBindingKind, Ref: "tb"}},
 					Template: v1alpha1.EventListenerTemplate{Name: "tt"},
 					Interceptors: []*v1alpha1.EventInterceptor{{
-						GitHub: &v1alpha1.GitHubInterceptor{},
-						GitLab: &v1alpha1.GitLabInterceptor{},
+						GitHub:    &v1alpha1.GitHubInterceptor{},
+						GitLab:    &v1alpha1.GitLabInterceptor{},
+						Bitbucket: &v1alpha1.BitBucketInterceptor{},
```
dibyom, line 352: I don't think adding it here is doing anything since this is just testing the validation fails with multiple interceptors

pkg/interceptors/cel/cel_test.go
```diff
@@ -203,7 +204,7 @@ func TestInterceptor_ExecuteTrigger(t *testing.T) {
 				rt.Fatalf("error reading response body: %v", err)
 			}
 			defer resp.Body.Close()
-			if !reflect.DeepEqual(got, tt.want) {
+			if diff := cmp.Diff(got, tt.want); diff != "" {
```
dibyom, line 207: Small nit: Usually, we'd want to  do something like the following:
```
if diff := cmp.Diff(got, tt.want); diff != "" {
  t.Errorf("FuncBeingTestedName() returned diff (-want +got): %s
}
```

This way we are consistent - the `want` always comes before `got` and the `diff` prints out with missing lines starting with a - and extra lines with a +

pkg/interceptors/github/github_test.go
```diff
@@ -256,7 +257,7 @@ func TestInterceptor_ExecuteTrigger_Signature(t *testing.T) {
 			if err != nil {
 				t.Fatalf("error reading response body %v", err)
 			}
-			if !reflect.DeepEqual(got, tt.want) {
+			if diff := cmp.Diff(got, tt.want); diff != "" {
```
dibyom, line 260: see comment above - want before got

pkg/interceptors/gitlab/gitlab_test.go
```diff
@@ -241,7 +241,7 @@ func TestInterceptor_ExecuteTrigger(t *testing.T) {
 				t.Fatalf("error reading response: %v", err)
 			}
 			defer resp.Body.Close()
-			if !reflect.DeepEqual(got, tt.want) {
+			if diff := cmp.Diff(got, tt.want); diff != "" {
```
dibyom, line 244: same as above

pkg/interceptors/cel/cel_test.go
savitaashture: Modified
Thank you

dibyom:
Excellent. Thank you @savitaashture Could you squash up the 2 commits and we can merge!
/approve

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
savitaashture: @dibyom I forgot to mention that adding `BitBucket` here will cover this [scenario](https://github.com/tektoncd/triggers/pull/578/files#diff-559661e0e83bb92fa19d37cc32078347R104-R106) 
The point of adding here is to cover the unit test so shall we add it back 

Let me know on this

dibyom: hmm, seems ok for now but  wondering if we should have a separate test for that...

Issue #406, opened 2020-02-03
Performance metrics


ncskier:
@dibyom commented from https://github.com/tektoncd/triggers/issues/356#issuecomment-578846320:
> I think having some performance numbers on the number of Triggers we can support would be nice!

> Given that we relaxed the validation to not check for the presence of bindings/templates, I don't think creating the EL is where we'll run into issues.

> We do use the Kubernetes/Triggers clientset to fetch resources in the EL sink at runtime (e.g. fetch the EL, then iterate over all the triggers; fetch the secret for GH/GL interceptors etc.) That might be problematic if we have a large number of incoming requests

# Expected Behavior
I expect to know the limitations of the Triggers project when using it on a large scale.

# Actual Behavior
I do not know the limitations of the Triggers project when using it on a large scale.

# Additional Info
This is a relatively open-ended issue, so please leave comments about performance metrics that you would find helpful.

lawrencejones:
We have ran into an issue where the fetching of the kubernetes secret for the github interceptors  validation is causing us to timeout our github webhooks. We're at about 40 triggers, and each triggers takes about 250ms to process, causing the webhook request to exceed the 10s Github timeout.

I think it's clear we need to cache these secrets. Do you have a preference on how? We could change the interceptors to use a caching Kubernetes client like a lot of controllers use, or we can build a `GetCachedSecretToken` that caches the value for a default TTL.

dibyom:
/lifecycle frozen

I think @savitaashture did some initial work on some perf metrics in one of here porposals!

PR #585, opened 2020-05-26
Cache trigger secrets for the duration of request

lawrencejones:
# Changes

This commit adds a request-local cache for interceptors to leverage
during the processing of triggers. It allows interceptors to avoid doing
expensive work more than once for each request, such as fetching a
Kubernetes secret for validating webhooks.

The implementation uses the request context to provide the cache. This
was the least disruptive method of providing a cache for use with
interceptors, and is appropriate if you consider the cache should live
only for the duration of each request.

Alternative implementations might have used the client-go informers to
extend the Kubernetes client to watch for secrets in the cluster. This
would cause the work required to fetch secrets to scale with the number
of secrets in the cluster, as opposed to making a fresh request per
webhook we process. That said, building caching clients seems like more
work than is necessary for fixing this simple problem, which is why I
went with a simple cache object.

The background for this change was finding Github webhooks timing out
once we exceeded ~40 triggers on our EventListener. While the CEL
filtering was super fast, the validation of Github webhook signatures
was being computed for every trigger, even though each trigger used the
same Github secret. Pulling the secret from Kubernetes was taking about
250ms, which meant 40 triggers exceeded the 10s Github timeout.

# Submitter Checklist

review them:

# Release Notes

```
Cache Kubernetes secret refs for each EventListener webhook, using the cached value to process each trigger
```

### DIFF
--- a/pkg/interceptors/interceptors.go
+++ b/pkg/interceptors/interceptors.go
@@ -17,7 +17,9 @@ limitations under the License.
 package interceptors
 
 import (
+	"context"
 	"net/http"
+	"path"
 
 	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


lawrencejones:
Hey team! I'm totally new to the tekton codebase, so I've opened this PR as a draft and left tests out for now.

Ideally you'd let me know if this approach looks viable to you, and if so, I'll add the tests and polish it.

This addresses a comment I made the other day here: https://github.com/tektoncd/triggers/issues/406#issuecomment-632088835

Thanks!

dibyom:
Hi @lawrencejones 

Thank you for the PR! Sorry I missed your comment on the issue (we were on holiday for the last few days). I think I have a slight preference for using a caching client -- we should be able to use the knative Store to make thing easier (like https://github.com/tektoncd/pipeline/pull/2637). That being said, this PR is definitely a step in the right direction and if the other approach ends up being too complex, we can merge this and then iterate!

lawrencejones:
That sounds fine to me. Didn't want to introduce the cached client unless it was used elsewhere. I'll have a look at your example and give that a shot :)

tragiclifestories:
Hi @dibyom - I might pick this up from Lawrence as my team's made time for investigating perf issues like this. It doesn't look like the knative library provides an equivalent to the `configmap` package for watching secrets (as per  tektoncd/pipeline#2637), so my instinctive approach would be to fall back on the underlying k8s client informer stuff to build a just-good-enough version. Does that make sense from your POV?

dibyom:
Hey @tragiclifestories that sounds good to me! Thanks for working on it!

tragiclifestories:
Hi there @dibyom, thanks for the reply. Having dug into the k8s client-go docs a little more, I've come to realise that implementing it that way is going to be a much bigger time sink than I had hoped. 

I have a [branch on my fork](https://github.com/tektoncd/triggers/compare/master...tragiclifestories:cache-trigger-secrets?expand=1) that essentially finishes this PR off in its current implementation with tests and so on. Though I plan to take another stab at the informer version this afternoon, it would be great if the request cache version was acceptable as a stop-gap, since there's a good chance it makes a major difference to the performance of event listeners with large numbers of Git(hub|lab) triggers as it currently is. Sorry for blowing hot and cold on this, still getting my head around all the various moving parts in k8s client land!

pkg/interceptors/interceptors.go
```diff
@@ -39,5 +74,8 @@ func GetSecretToken(cs kubernetes.Interface, sr *triggersv1.SecretRef, eventList
 		return nil, err
 	}
 
-	return secret.Data[sr.SecretKey], nil
+	secretValue := secret.Data[sr.SecretKey]
+	cache[cacheKey] = secret.Data[sr.SecretKey]
```
jace-ys, line 78: Shouldn't the intention here be to set the secret value into our request context cache rather than a local cache variable? I'd imagine this would be a function like `func setCache(req *http.Request) {}`

dibyom:
> Though I plan to take another stab at the informer version this afternoon, it would be great if the request cache version was acceptable as a stop-gap, since there's a good chance it makes a major difference to the performance of event listeners with large numbers of Git(hub|lab) triggers as it currently is.

Sure, this is definitely an improvement over what we have at the moment and since its not an external facing API change, we should be able to switch to a another implementation later. Happy to review a PR with the request cache changes!

tragiclifestories:
OK! 

Request cache version: #595

k8s cache prototype: #594

dibyom:
I'm going to close this in favor of the other two PRs mentioned!
/close

Issue #525, opened 2020-04-07
Can't use Pipeline level params with TriggerTemplate params


bobcatfish:
# Expected Behavior

I should be able to use params from the TriggerTemplate inside of a resource template that contains a Pipeline spec which also has it's own params.

# Actual Behavior

Both TriggerTemplate params + Pipeline params are specified with `$(params.my-param)` and TriggerTemplates can't distinguish between them, meaning that when you try to use a pipeline level param, TriggerTemplates will complain that it does not recognize the param.

# Steps to Reproduce the Problem

For example if I try to make a TriggerTemplate similar to this, where the Pipeline has a param `$(params.bucket-name)` with a default value:

```yaml
apiVersion: tekton.dev/v1alpha1
kind: TriggerTemplate
metadata:
  name: gcs-triggertemplate
spec:
  resourceTemplates:
    - apiVersion: tekton.dev/v1beta1
      kind: PipelineRun
      metadata:
        name: gcs-pr-$(uid)
      spec:
        pipelineSpec:
          params: bucket-name
          default: tekton-test-bucket-$(uid)
          tasks:
          - name: some-task
            taskRef:
              name: some-task
            params:
            - name: location
              value: $(params.gcs-bucket)
```

Because there is no TriggerTemplate param `bucket-name` I'll get an error like this:
```
Error from server (BadRequest): error when creating "./gcs/samples/gcs-sample.yaml": admission webhook "validation.webhook.triggers.tekton.dev" denied the request
: validation failed: invalid value: undeclared param '$(params.bucket-name)': spec.resourcetemplates[0]
'$(params.bucket-name)' must be declared in spec.params
```

# Additional Info

Ideas for possible fixes:
* Don't try to catch unknown params (kind of sad for a user who mistypes a TriggerTemplate param but maybe worth it for the flexibility?)
* Change the syntax to be more explicit, either in TriggerTemplates, Pipelines or both (e.g. `triggertemplate.params.foo` and `pipeline.params.foo`)
* Add some kind of escaping syntax??

(discovered this with @sbwsg )

dibyom:
Another possible solution might be the one suggested in #508 i.e. TriggerTemplate params use $(tt.params) vs $(params)

bobcatfish:
Looks like I duplicated #508? feel free to close this! (I like the tt.params syntax btw)

dibyom:
I think the solution is similar but this is technically a "different" bug 
/kind bug

skaegi:
[...]

PR #589, opened 2020-05-29
Modify spec params to avoid confusion between resourcetemplates and triggertemplate params

savitaashture:
# Changes
PR helps to avoid confusion from `TriggerTemplate params` to `resourceTemplate params`
by keeping params of spec as it is and using $(tt.params.) for the substitution.

So, this way we no need to handle the backward compatibility as per the discussion over the [comments](https://github.com/tektoncd/triggers/pull/589#issuecomment-646129728)
 
Fixes https://github.com/tektoncd/triggers/issues/508

Tried to do tag configurable, given the way ObjectDefaulter and SetDefaults work, neither could accomplish the switch properly. ObjectDefaulter could have done it, but the way the tests are configured do not cause the Defaults to be executed since the objects aren't actually decoded.

# Submitter Checklist

review them:

# Release Notes

The substitution for `params` modified from `$(params.gitrevision)` to `$(tt.params.gitrevision)` . 
And the the examples

```
kind: TriggerTemplate
metadata:
  name: github-template
spec:
  params:
    - name: gitrevision
    - name: gitrepositoryurl
  resourcetemplates:
    - apiVersion: tekton.dev/v1alpha1
      kind: TaskRun
      metadata:
        generateName: github-run-
      spec:
        taskSpec:
          inputs:
            resources:
              - name: source
                type: git
          steps:
            - image: ubuntu
              script: |
                #! /bin/bash
                ls -al $(inputs.resources.source.path)
        inputs:
          resources:
            - name: source
              resourceSpec:
                type: git
                params:
                  - name: revision
                    value: $(tt.params.gitrevision)
                  - name: url
                    value: $(tt.params.gitrepositoryurl)
```
For a release or two $(params) supported and later it will be deprecated.
Created [issue](https://github.com/tektoncd/triggers/issues/606) to track the same

cc @dibyom

### DIFF
--- a/docs/triggertemplates.md
+++ b/docs/triggertemplates.md
@@ -47,9 +47,9 @@ spec:
           type: git
           params:
           - name: revision
-            value: $(params.gitrevision)
+            value: $(tt.params.gitrevision)
           - name: url
-            value: $(params.gitrepositoryurl)
+            value: $(tt.params.gitrepositoryurl)
 ```
 
 `TriggerTemplates` currently support the following [Tekton Pipelines](https://github.com/tektoncd/pipelines) resources:
@@ -102,11 +102,11 @@ have an optional `description` and `default` value.
 substitution syntax, where `<name>` is the name of the parameter:
 
 ```YAML
-$(params.<name>)
+$(tt.params.<name>)
 ```
 
-`params` can be referenced in the `resourceTemplates` section of a
-`TriggerTemplate`. The purpose of `params` is to make `TriggerTemplates`
+`tt.params` can be referenced in the `resourceTemplates` section of a
+`TriggerTemplate`. The purpose of `tt.params` is to make `TriggerTemplates`
 reusable.
 
 ## Best Practices
--- a/pkg/apis/triggers/v1alpha1/trigger_template_validation.go
+++ b/pkg/apis/triggers/v1alpha1/trigger_template_validation.go
@@ -28,8 +28,8 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-// paramsRegexp captures TriggerTemplate parameter names $(params.NAME)
-var paramsRegexp = regexp.MustCompile(`\$\(params.(?P<var>[_a-zA-Z][_a-zA-Z0-9.-]*)\)`)
+// paramsRegexp captures TriggerTemplate parameter names $(tt.params.NAME) or $(params.NAME)
+var paramsRegexp = regexp.MustCompile(`\$\((params|tt.params).([_a-zA-Z][_a-zA-Z0-9.-]*)\)`)
 
 // Validate validates a TriggerTemplate.
 func (t *TriggerTemplate) Validate(ctx context.Context) *apis.FieldError {
@@ -91,16 +91,24 @@ func verifyParamDeclarations(params []ParamSpec, templates []TriggerResourceTemp
 		declaredParamNames[param.Name] = struct{}{}
 	}
 	for i, template := range templates {
-		// Get all params in the template $(params.NAME)
+		// Get all params in the template $(tt.params.NAME) or $(params.NAME)
 		templateParams := paramsRegexp.FindAllSubmatch(template.RawExtension.Raw, -1)
 		for _, templateParam := range templateParams {
-			templateParamName := string(templateParam[1])
+			templateParamName := string(templateParam[2])
 			if _, ok := declaredParamNames[templateParamName]; !ok {
+				// This logic is to get the tag and display error dynamically for both tt.params and params.
+				// TODO(#606)
+				var tag string
+				if string(templateParam[1]) == "params" {
+					tag = "params"
+				} else {
+					tag = "tt.params"
+				}
 				fieldErr := apis.ErrInvalidValue(
-					fmt.Sprintf("undeclared param '$(params.%s)'", templateParamName),
+					fmt.Sprintf("undeclared param '$(%s.%s)'", tag, templateParamName),
 					fmt.Sprintf("[%d]", i),
 				)
-				fieldErr.Details = fmt.Sprintf("'$(params.%s)' must be declared in spec.params", templateParamName)
+				fieldErr.Details = fmt.Sprintf("'$(%s.%s)' must be declared in spec.params", tag, templateParamName)
 				return fieldErr
 			}
 		}
--- a/pkg/sink/sink_test.go
+++ b/pkg/sink/sink_test.go
@@ -153,7 +153,7 @@ func TestHandleEvent(t *testing.T) {
 			Kind:       "PipelineResource",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "$(params.name)",
+			Name:      "$(tt.params.name)",
 			Namespace: namespace,
 			Labels: map[string]string{
 				"app":  "$(params.foo)",
@@ -273,7 +273,7 @@ func TestHandleEventWithInterceptors(t *testing.T) {
 			Type: pipelinev1alpha1.PipelineResourceTypeGit,
 			Params: []pipelinev1alpha1.ResourceParam{{
 				Name:  "url",
-				Value: "$(params.url)",
+				Value: "$(tt.params.url)",
 			}},
 		},
 	}
@@ -413,7 +413,7 @@ func TestHandleEventWithWebhookInterceptors(t *testing.T) {
 			Kind:       "PipelineResource",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "$(params.name)",
+			Name:      "$(tt.params.name)",
 			Namespace: namespace,
 		},
 		Spec: pipelinev1alpha1.PipelineResourceSpec{
@@ -1001,7 +1001,7 @@ func getResources(t *testing.T, triggerBindingParam string) (*v1alpha1.TriggerBi
 			Type: pipelinev1.PipelineResourceTypeGit,
 			Params: []pipelinev1.ResourceParam{{
 				Name:  "url",
-				Value: "$(params.url)",
+				Value: "$(tt.params.url)",
 			}},
 		},
 	}
--- a/pkg/template/resource.go
+++ b/pkg/template/resource.go
@@ -110,12 +110,17 @@ func ApplyParamsToResourceTemplate(params []triggersv1.Param, rt json.RawMessage
 // applyParamToResourceTemplate returns the TriggerResourceTemplate with the
 // param value substituted for all matching param variables in the template
 func applyParamToResourceTemplate(param triggersv1.Param, rt json.RawMessage) json.RawMessage {
-	// Assume the param is valid
-	paramVariable := fmt.Sprintf("$(params.%s)", param.Name)
-	// Escape quotes so that that JSON strings can be appended to regular strings.
-	// See #257 for discussion on this behavior.
-	paramValue := strings.Replace(param.Value, `"`, `\"`, -1)
-	return bytes.Replace(rt, []byte(paramVariable), []byte(paramValue), -1)
+	// The changes are for backward compatibility with both $(params) and $(tt.params)
+	// TODO(#606)
+	for _, tag := range []string{"params", "tt.params"} {
+		// Assume the param is valid
+		paramVariable := fmt.Sprintf("$(%s.%s)", tag, param.Name)
+		// Escape quotes so that that JSON strings can be appended to regular strings.
+		// See #257 for discussion on this behavior.
+		paramValue := strings.Replace(param.Value, `"`, `\"`, -1)
+		rt = bytes.Replace(rt, []byte(paramVariable), []byte(paramValue), -1)
+	}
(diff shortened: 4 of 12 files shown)


dibyom:
Thanks for opening this @savitaashture I won't get to this today but will put in on my list for tomorrow!

pkg/apis/triggers/v1alpha1/trigger_template_validation.go
```diff
@@ -85,22 +85,30 @@ func validateResourceTemplates(templates []TriggerResourceTemplate) *apis.FieldE
...
-		// Get all params in the template $(params.NAME)
+		// Get all params in the template $(ttparams.NAME) or $(params.NAME)
 		templateParams := paramsRegexp.FindAllSubmatch(template.RawExtension.Raw, -1)
 		for _, templateParam := range templateParams {
-			templateParamName := string(templateParam[1])
+			templateParamName := string(templateParam[2])
 			if _, ok := declaredParamNames[templateParamName]; !ok {
+				// This logic is to get the tag and display in error dynamically for both ttparams and params.
+				// TODO(savitaashture): remove params once after the complete deprecation of spec.DeprecatedParams
+				var tag string
```
dibyom, line 101: Question: Until we start using only $(ttparams) for replacement, we cannot support https://github.com/tektoncd/triggers/issues/525 right?

dibyom:
> params are now deprecated and will be removed in future but for now TriggerTemplate spec can support param as shown below

Might be worth rewording as - The `params` field in TriggerTemplates has been deperecated in favor of the `ttparams` field. And the the examples

pkg/apis/triggers/v1alpha1/trigger_template_defaults.go
```diff
@@ -21,4 +21,10 @@ import (
 )
 
 // SetDefaults initializes TriggerTemplate with default values.
-func (tt *TriggerTemplate) SetDefaults(ctx context.Context) {}
+func (tt *TriggerTemplate) SetDefaults(ctx context.Context) {
+	// This is a temporary logic to have a backward compatibility with the original params.
+	// TODO(savitaashture): remove below logic once after the complete deprecation of Spec.DeprecatedParams
```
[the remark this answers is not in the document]
savitaashture: Sure i will open and update in the code comments

pkg/apis/triggers/v1alpha1/trigger_template_validation.go
savitaashture: Yes because till that time we will have validation for `params`

pkg/apis/triggers/v1alpha1/trigger_template_defaults.go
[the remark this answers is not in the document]
savitaashture: Created [issue](https://github.com/tektoncd/triggers/issues/606) to track the removal of deprecation tag `params`

bobcatfish:
Drive by comment - instead of changing the name of the params field, you could change how the replacement works, e.g. use `$(tt.params.foo)` when using the field, but still call the section `params` in the trigger template spec

whaddaya think (ignore me if this has been discussed already)

savitaashture:
> Drive by comment - instead of changing the name of the params field, you could change how the replacement works, e.g. use `$(tt.params.foo)` when using the field, but still call the section `params` in the trigger template spec
> 
> whaddaya think (ignore me if this has been discussed already)

Hi @bobcatfish Its good point
Right now I don't remember that whether we discussed the point which you mentioned.
But i just want to know any reason behind this `$(tt.params.foo)` suggestion
Or is it because consistency with `params`

bobcatfish:
> But i just want to know any reason behind this $(tt.params.foo) suggestion
>Or is it because consistency with params

Just consistency! Across tekton we have multiple `spec.params` - it seems like the right name for the field. The problem is the ambiguity in the usage, feels weird to solve that by changing the name of the field.

Mostly a subjective opinion tho!

dibyom:
I like the `$(tt.params)` idea. I think we discussed this though there weren't any strong opinions one way or another. 
I agree with the consistency issue and  that renaming the field feels a bit weird. Calling the field `params` but using it as `tt.params` is also weird but less so in my opinion 👼 

How do you feel about keeping the `params` field and using `tt.params` for substitution? Also if we go ahead with this: I think we should support replacements with both `$(params)` and  `$(tt.params)` for a release before we drop support for `$(params)`?

savitaashture:
> I like the `$(tt.params)` idea. I think we discussed this though there weren't any strong opinions one way or another.
> I agree with the consistency issue and that renaming the field feels a bit weird. Calling the field `params` but using it as `tt.params` is also weird but less so in my opinion 
> 
> How do you feel about keeping the `params` field and using `tt.params` for substitution? Also if we go ahead with this: I think we should support replacements with both `$(params)` and `$(tt.params)` for a release before we drop support for `$(params)`?

@dibyom yes i think we need to keep `$(params)` and `$(tt.params)` for a release before we drop support for `$(params)`

Let me understand clearly one of your point here 
**How do you feel about keeping the `params` field and using `tt.params` for substitution?** 
Do you mean something like below 
```
spec:
  params:
    - name: gitrevision
    - name: gitrepositoryurl
  resourcetemplates:
    - apiVersion: tekton.dev/v1alpha1
      kind: TaskRun
      metadata:
        generateName: github-run-
      spec:
        taskSpec:
          inputs:
            resources:
              - name: source
                type: git
          steps:
            - image: ubuntu
              script: |
                #! /bin/bash
                ls -al $(inputs.resources.source.path)
        inputs:
          resources:
            - name: source
              resourceSpec:
                type: git
                params:
                  - name: revision
                    value: $(tt.params.gitrevision)
                  - name: url
                    value: $(tt.params.gitrepositoryurl)
```
If thats the thing i feel it creates bit confusion :thinking: 

OR

This way
```
spec:
  tt:
    params:
      - name: gitrevision
      - name: gitrepositoryurl
  resourcetemplates:
    - apiVersion: tekton.dev/v1alpha1
      kind: TaskRun
      metadata:
        generateName: github-run-
      spec:
        taskSpec:
          inputs:
            resources:
              - name: source
                type: git
          steps:
            - image: ubuntu
              script: |
                #! /bin/bash
                ls -al $(inputs.resources.source.path)
        inputs:
          resources:
            - name: source
              resourceSpec:
                type: git
                params:
                  - name: revision
                    value: $(tt.params.gitrevision)
                  - name: url
                    value: $(tt.params.gitrepositoryurl)
```

dibyom:
Yeah, the first one. I agree it can be a bit confusing 😄 On the other hand calling the field `ttparams` when everywhere else (e.g. TriggerBindings, Pipelines etc.) we call it `params` is also a bit inconsistent.

savitaashture:
> Yeah, the first one. I agree it can be a bit confusing  On the other hand calling the field `ttparams` when everywhere else (e.g. TriggerBindings, Pipelines etc.) we call it `params` is also a bit inconsistent.

Ya even i am thinking of confusing part.
Otherwise + point is we no need to do backward compatibility if we keep field `params` as it is and use  `$(tt.params)` everywhere

savitaashture:
@dibyom Modified as per the discussion

**Changes:**
Keep $(params) field and use $(tt.params) for the substitution

pkg/template/resource.go
```diff
@@ -111,7 +111,7 @@ func ApplyParamsToResourceTemplate(params []triggersv1.Param, rt json.RawMessage
 // param value substituted for all matching param variables in the template
 func applyParamToResourceTemplate(param triggersv1.Param, rt json.RawMessage) json.RawMessage {
 	// Assume the param is valid
-	paramVariable := fmt.Sprintf("$(params.%s)", param.Name)
+	paramVariable := fmt.Sprintf("$(tt.params.%s)", param.Name)
```
dibyom, line 114: Are we just supporting `$(tt.params)` for now.  Shouldn't we support both `$(params)` or `$(tt.params)` for a release or two?

savitaashture: Updated changes
Please do have a look 
Thank you

pkg/template/resource.go
```diff
@@ -110,8 +111,19 @@ func ApplyParamsToResourceTemplate(params []triggersv1.Param, rt json.RawMessage
...
+	// TODO(#606)
+	var (
+		paramsRegexp  = regexp.MustCompile(`\$\(params.(?P<var>[_a-zA-Z][_a-zA-Z0-9.-]*)\)`)
+		paramVariable string
+	)
+	templateParams := paramsRegexp.FindAllSubmatch(rt, -1)
 	// Assume the param is valid
-	paramVariable := fmt.Sprintf("$(params.%s)", param.Name)
+	if len(templateParams) == 0 {
+		paramVariable = fmt.Sprintf("$(tt.params.%s)", param.Name)
```
dibyom, line 123: Is there a reason we cannot attempt to replace with both `$(params)` as well as `$(tt.params)`?

savitaashture: The reason is we will not be getting the information on user provided ResourceTemplate params
whether its `$(params)` or `$(tt.params)` because based on that we need to replace.
So currently with the help of regular expression able to distinguish whether its $(params) or $(tt.params).

docs/triggertemplates.md
```diff
@@ -47,9 +47,9 @@ spec:
...
           params:
           - name: revision
-            value: $(params.gitrevision)
+            value: $(tt.params.gitrevision)
           - name: url
-            value: $(params.gitrepositoryurl)
+            value: $(tt.params.gitrepositoryurl)
 ```
 
 `TriggerTemplates` currently support the following [Tekton Pipelines](https://github.com/tektoncd/pipelines) resources:
```
khrm, line 55: We need to update https://github.com/tektoncd/triggers/blob/ff20c8592a1b419ac70b2f46987ded3658bfeada/docs/triggertemplates.md#parameters

savitaashture: Updated doc :+1:

pkg/apis/triggers/v1alpha1/trigger_template_validation.go
```diff
@@ -91,16 +91,24 @@ func verifyParamDeclarations(params []ParamSpec, templates []TriggerResourceTemp
 		declaredParamNames[param.Name] = struct{}{}
 	}
 	for i, template := range templates {
-		// Get all params in the template $(params.NAME)
+		// Get all params in the template $(tt.params.NAME) or $(params.NAME)
 		templateParams := paramsRegexp.FindAllSubmatch(template.RawExtension.Raw, -1)
 		for _, templateParam := range templateParams {
-			templateParamName := string(templateParam[1])
+			templateParamName := string(templateParam[2])
```
wlynch, line 97: I realize this is following what was set before, but the direct array access is at risk for nil panics. I think we should do 2 things:

1) Have a length check prior to accessing `templateParam[2]`.
2) Create a more exhaustive set of regexp checks instead of the more e2e tests we see here. (mainly to give us more confidence that the regexp is matching to what we expect, even in unexpected cases).

pkg/sink/sink_test.go
```diff
@@ -153,18 +153,18 @@ func TestHandleEvent(t *testing.T) {
 			Kind:       "PipelineResource",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "$(params.name)",
+			Name:      "$(tt.params.name)",
```
wlynch, line 156: We should probably leave some of these tests using the old params syntax to ensure backwards compatibility, since this is the first place where we are actually using the value outside of validation.

pkg/template/resource.go
wlynch: I don't think this works if a user mixes both params and tt.params in the same file. It should be safe to do a double pass - e.g.

```go
for _, tag := range []string{ "params", "tt.params"} {
  paramVariable := fmt.Sprintf("$(%s.%s)", tag, param.Name)
  rt = bytes.Replace(rt, []byte(paramVariable), []byte(paramValue), -1)
}
```

pkg/sink/sink_test.go
dibyom: +1 to having some tests with the old `$(params)`

pkg/apis/triggers/v1alpha1/trigger_template_validation.go
savitaashture: Actually 
```
templateParams := paramsRegexp.FindAllSubmatch(template.RawExtension.Raw, -1)
```
templateParams length will be 0 in any unexpected cases because of Regular Expression
```
regexp.MustCompile(`\$\((params|tt.params).([_a-zA-Z][_a-zA-Z0-9.-]*)\)`)
```
If user provides any such values like below
```
1. params1.foo
2. tt1.params.foo
3. tt.params1.foo
4. tt.foo
5. .foo
```
the `templateParams` itself is `0` and because of that it never enters into for loop 

But yes need to add more test cases to ensures the regex will not break.

savitaashture: Added test cases to cover few more scenario

PR #594, opened 2020-06-03
Cache secrets in interceptors with Reflector

tragiclifestories:
# Changes

This is an alternative implementation of #585 using client-go's cache package. We create a reflector for secrets across all namespaces and use it as a caching layer for secrets in both the Github/lab webhook parsers and CEL `compareSecret` calls. 

I'm putting this up as a draft to get immediate feedback. Still need to test. Also, now that it's also backing CEL functions, `WebhookSecretStore` is probably the wrong name ... 

# Submitter Checklist

review them:

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```


pkg/interceptors/interceptors.go
```diff
@@ -17,27 +17,83 @@ limitations under the License.
...
+	eventListenerNamespace string
+}
+
+// NewWebhookSecretStore provides cached lookups of k8s secrets, backed by a Reflector.
+func NewWebhookSecretStore(cs kubernetes.Interface, ns string, resyncInterval time.Duration, stopCh <-chan struct{}) WebhookSecretStore {
+	secretsClient := cs.CoreV1().Secrets(metav1.NamespaceAll)
+	store := cache.NewStore(func(obj interface{}) (string, error) {
+		secret, ok := obj.(corev1.Secret)
+		if !ok {
+			return "", errors.New("Object is not a secret")
```
jace-ys, line 48:
```suggestion
			return "", errors.New("object is not a secret")
```
I think the convention is to use lower-cased errors so you can wrap errors without uneven capitalisation.

pkg/interceptors/interceptors.go
```diff
@@ -17,27 +17,83 @@ limitations under the License.
...
 
-func GetSecretToken(cs kubernetes.Interface, sr *triggersv1.SecretRef, eventListenerNamespace string) ([]byte, error) {
-	ns := sr.Namespace
-	if ns == "" {
-		ns = eventListenerNamespace
+// Get returns the secret value for a given SecretRef.
+func (ws *WebhookSecretStore) Get(sr triggersv1.SecretRef) ([]byte, error) {
+	cachedObj, ok, _ := ws.store.GetByKey(getKey(sr))
+	if !ok {
+		return nil, fmt.Errorf("Secret not found: %s", sr.SecretName)
```
jace-ys, line 70:
```suggestion
		return nil, fmt.Errorf("secret not found: %s", sr.SecretName)
```

pkg/interceptors/interceptors.go
```diff
@@ -17,27 +17,83 @@ limitations under the License.
...
 
-// Interceptor is the interface that all interceptors implement.
-type Interceptor interface {
-	ExecuteTrigger(req *http.Request) (*http.Response, error)
+// WebhookSecretStore provides cached lookups of k8s secrets, backed by a Reflector.
+type WebhookSecretStore interface {
+	Get(sr triggersv1.SecretRef) ([]byte, error)
+}
+
+type webhookSecretStore struct {
```
jace-ys, line 37: To address your question, maybe just `secretStore` would work

pkg/interceptors/interceptors.go
```diff
@@ -17,27 +17,83 @@ limitations under the License.
...
+	return secretStore
 }
 
-func GetSecretToken(cs kubernetes.Interface, sr *triggersv1.SecretRef, eventListenerNamespace string) ([]byte, error) {
-	ns := sr.Namespace
-	if ns == "" {
-		ns = eventListenerNamespace
+// Get returns the secret value for a given SecretRef.
+func (ws *WebhookSecretStore) Get(sr triggersv1.SecretRef) ([]byte, error) {
+	cachedObj, ok, _ := ws.store.GetByKey(getKey(sr))
```
jace-ys, line 68: Not very familiar with the cache store, so curious why we're ignoring the error here?

pkg/interceptors/interceptors.go
```diff
@@ -17,27 +17,83 @@ limitations under the License.
...
+	if !ok {
+		return nil, fmt.Errorf("Secret not found: %s", sr.SecretName)
+	}
+
+	secret, ok := cachedObj.(corev1.Secret)
+	if !ok {
+		return nil, fmt.Errorf("Cached object is not a secret: %s", sr.SecretName)
```
jace-ys, line 75:
```suggestion
		return nil, fmt.Errorf("cached object is not a secret: %s", sr.SecretName)
```

pkg/interceptors/interceptors.go
tragiclifestories: Yes, which of course was what it was when I first wrote this struct def out 😅

pkg/interceptors/interceptors.go
tragiclifestories: It turns out that if you spelunk into the code, error is always nil from this particular store implementation. However, I will probably handle it anyway, since it may turn out that a different store makes more sense in future.

pkg/sink/sink.go
```diff
@@ -186,18 +185,23 @@ func (r Sink) executeInterceptors(t *triggersv1.EventListenerTrigger, in *http.R
...
 		var interceptor interceptors.Interceptor
 		switch {
 		case i.Webhook != nil:
 			interceptor = webhook.NewInterceptor(i.Webhook, r.HTTPClient, r.EventListenerNamespace, log)
 		case i.GitHub != nil:
-			interceptor = github.NewInterceptor(i.GitHub, r.KubeClientSet, r.EventListenerNamespace, log)
+			interceptor = github.NewInterceptor(i.GitHub, r.WebhookSecretStore, log)
 		case i.GitLab != nil:
-			interceptor = gitlab.NewInterceptor(i.GitLab, r.KubeClientSet, r.EventListenerNamespace, log)
+			interceptor = gitlab.NewInterceptor(i.GitHub, r.WebhookSecretStore, log)
```
bigkevmcd, line 202: This looks like a copy/paste mistake?

bigkevmcd:
Definitely would help with scaling both volume of requests, and triggers that use secrets.

This caching layer is something we do need to consider if we go for a process-based plugin mechanism.

tragiclifestories:
@bigkevmcd I'll be picking this up again. There are actually a whole host of uncached requests going on in `processTrigger` and it seems like a lot of caching will be necessary to keep this performant with a lot of triggers.

pkg/sink/sink.go
tragiclifestories: it does indeed ...

tragiclifestories:
So, on further investigation, the fix for this problem implemented in https://github.com/tektoncd/triggers/pull/595 did not work. The reason for this seems to be that the the triggers are executed asynchronously, and thus all immediately stampede to get the secret at the very top of the ExecuteTrigger methods, so in practice you get 100% cache misses, 100% of the time. Which was not what was intended, to put it mildly. 

This version speeds things up very significantly compared to master in my tests - more thorough than last time around ;-). It takes about 3s to process the hideous 500-github-trigger YAML I put in the examples in this PR - and most of that is now due to another, far less severe performance problem with compiling CEL expressions (will raise an issue when I get a second).

pkg/interceptors/interceptors.go
```diff
@@ -17,73 +17,84 @@ limitations under the License.
...
-func getCache(req *http.Request) map[string]interface{} {
-	if cache, ok := req.Context().Value(requestCacheKey).(map[string]interface{}); ok {
-		return cache
+	secretStore := webhookSecretStore{
+		store:                  store,
+		eventListenerNamespace: ns,
 	}
 
-	return make(map[string]interface{})
+	lw := cache.NewListWatchFromClient(cs.CoreV1().RESTClient(), "secrets", metav1.NamespaceAll, fields.Everything())
```
tragiclifestories, line 59: Using NamespaceAll here is probably  not a long term solution, since it requires cluster-wide list secrets in RBAC. I guess we could wrap a map of stores here - one per namespace referenced in calls to `Get` ...

cmd/eventlistenersink/main.go
```diff
@@ -80,12 +82,15 @@ func main() {
 		logger.Fatal(err)
 	}
 
+	webhookSecretStore := interceptors.NewWebhookSecretStore(kubeClient, sinkArgs.ElNamespace, 5*time.Second, stopCh)
```
tragiclifestories, line 85: The interval should probably be configurable here ...

tragiclifestories:
I hope I'll get some time to update this in the next week or so. Any ideas on how to write tests for this - and why they don't currently pass - very much welcome. FWIW, we've a version of this PR running in production for over a month with no incident.

tragiclifestories:
/remove-lifecycle stale Promise I'll get to this soon ...

bobcatfish:
hey @tragiclifestories ! do you have any strong feelings about us closing this PR for now? You can absolutely re-open it (or open a new one) when you're ready to get back to it

tragiclifestories:
At this point the rebase would probably take longer than redoing it from scratch 😅 . I think this PR can be closed. 

I will try to find a moment to see if the bug still exists - it may well not for all I know ...

PR #595, opened 2020-06-03
Cache trigger secrets for the duration of request (with tests)

tragiclifestories:
# Changes

This is essentially an additional commit on top of #585 by my colleague @lawrencejones adding tests. From that summary:

> This commit adds a request-local cache for interceptors to leverage
during the processing of triggers. It allows interceptors to avoid doing
expensive work more than once for each request, such as fetching a
Kubernetes secret for validating webhooks.
> 
> The implementation uses the request context to provide the cache. This
was the least disruptive method of providing a cache for use with
interceptors, and is appropriate if you consider the cache should live
only for the duration of each request.

# Submitter Checklist

review them:

# Release Notes

```
Cache Kubernetes secret refs for each EventListener webhook, using the cached value to process each trigger
```


pkg/interceptors/interceptors_test.go
```diff
@@ -0,0 +1,102 @@
...
+			if _, err := kubeClient.CoreV1().Secrets(testNS).Create(makeSecret("secret from API")); err != nil {
+				rt.Error(err)
+			}
+
+			secret, err := GetSecretToken(req, kubeClient, &secretRef, testNS)
+			if err != nil {
+				rt.Error(err)
+			}
+
+			if !bytes.Equal(secret, tt.wanted) {
```
dibyom, line 73: minor nit: usually, we use `cmp.Diff` which prints out the diff between want and got nicely: 
 https://github.com/tektoncd/triggers/blob/master/pkg/interceptors/webhook/webhook_test.go#L91

tragiclifestories: done!

dibyom:
Just one last thing before merge -- could you squash the commits into one?

tragiclifestories:
@dibyom Cheers - all done now.

PR #596, opened 2020-06-04
Modify BitBucketInterceptor to BitbucketInterceptor

savitaashture:
# Changes

1. Updated README to mention specifically that the interceptor supports for bitbucket server not for bitbucket cloud. 
2. Modified name for the Bitbucket interceptors in order to match with `atlassian` doc
# Submitter Checklist

review them:

Fixes: https://github.com/tektoncd/triggers/issues/362


docs/eventlisteners.md
```diff
@@ -421,7 +421,7 @@ spec:
 ### Bitbucket Interceptors
 
 Bitbucket Interceptors contain logic to validate and filter webhooks that come from
-Bitbucket server or cloud. Supported features include validating webhooks actually came from Bitbucket as well as
+Bitbucket server and not the Bitbucket cloud. Supported features include validating webhooks actually come from Bitbucket server as well as
```
bigkevmcd, line 424: How about something like...

> The Bitbucket interceptor provides support for hooks originating in [Bitbucket server](https://confluence.atlassian.com/bitbucketserver), providing server hook signature validation and event-filtering.
>
> [Bitbucket cloud](https://support.atlassian.com/bitbucket-cloud/) is _not_ currently supported by this interceptor, but it has no secret validation, so you could match on the incoming requests using the CEL interceptor.

dibyom: question: what would it take to support bitbucket cloud?

bigkevmcd: Bitbucket cloud interceptor doesn't support secret validation, but it could match on the event types.

Couple of options, maybe an extra parameter for the interceptor to indicate `cloud` or `server`?

Or try one and then the other, but the JSON bodies are actually quite different between them.

(This is "solved" by my hook standardisation proposal 🙂)

bigkevmcd: Oops, lemme clarify, Bitbucket cloud doesn't send secrets to validate.

bigkevmcd: https://confluence.atlassian.com/bitbucket/manage-webhooks-735643732.html

> Secure webhooks

> Webhooks work  Bitbucket sending a POST to the customer service at a particular URL.
> It's important that you protect your service from potential attacks. As a workaround, you
> could whitelist the Bitbucket IPs to restrict who can access your webhook resource. 

So, maybe instead of secrets, for the "cloud" version, you add the list of IPs?

bigkevmcd: (yeah, that's not a pretty option)

dibyom: oh ok, did not realize cloud did not support shared secret validation. IP whitelisting might get a bit complex for scenarios where there are other entities like reverse proxies or knative eventing processing the request before the EventListener. Maybe just the types are fine for now until the standardized hook type?

savitaashture: > How about something like...
> 
> > The Bitbucket interceptor provides support for hooks originating in [Bitbucket server](https://confluence.atlassian.com/bitbucketserver), providing server hook signature validation and event-filtering.
> > [Bitbucket cloud](https://support.atlassian.com/bitbucket-cloud/) is _not_ currently supported by this interceptor, but it has no secret validation, so you could match on the incoming requests using the CEL interceptor.

@bigkevmcd is that okay if we include work `hook` right now in the doc because it may lead to confusion on what can be `hook` so just wanted to clarify before updating PR

savitaashture: > oh ok, did not realize cloud did not support shared secret validation. IP whitelisting might get a bit complex for scenarios where there are other entities like reverse proxies or knative eventing processing the request before the EventListener. Maybe just the types are fine for now until the standardized hook type?

@dibyom `Maybe just the types are fine for now until the standardized hook type?` Could you please help me to understand what will be `type` here

dibyom: I was just referring to the `eventTypes` field. Going through this again -- I think adding @bigkevmcd 's [suggested change](https://github.com/tektoncd/triggers/pull/596#discussion_r435918398) seems like a good way to proceed here!

savitaashture: @dibyom Updated README as per the @bigkevmcd suggestion

Issue #505, opened 2020-03-27
Feature request: Eventlistener should support customized podTemplate


MartinForReal:
# Expected Behavior
Event listener should accept customizable default pod template as what tekton pipeline did. 
If we can add a field in CRD that would be better

# Actual Behavior
Event listener is created using the default deployment template.  But it doesn't comply with our schedule policy. 

# Additional Info
https://github.com/tektoncd/pipeline/pull/1901

MartinForReal:
@bobcatfish @dibyom @vdemeester @ckadner @Tomcli

ncskier:
This issue sounds similar to https://github.com/tektoncd/triggers/issues/440. I wonder if we can address both issues in the same way?

dibyom:
Yeah this seems like a very reasonable feature request. I'm for adding this

dibyom:
I wonder if a podTemplate would be enough though....is there a need for customizing things such as number of replicas etc.?

MartinForReal:
Sorry for the late reply.  
I browsed the sink's code and found a lot of useful information in the discussion in #370 .
At the service access level, we need to find a way to reduce the number of ingresses, svc resources so that the configurations are removed and attack risk is reduced.  
At the workload level, we need to reduce the usage of the resource. The best scenario is that the sink is created at-hot and destroyed if there are no requests. 

So I think a pub-sub structure can help in the service access level. I totally agree that we should have a global event listener which distributes the event to event sinks. We can leverage the message queue so that the sink doesn't need to listen to the HTTP port like what knative source did. And we have various options to implement the message queue. Such as depending on NATS, Kafka and etc, Or Adopting actor model(Akka HTTP, https://github.com/AsynkronIT/protoactor-go ). 
At the workload level, we also have a lot of options. we can define sink CR and refactor sink to subscribe to the event stream. And a new sink can be implemented as long as it implements the CR controller. 

In the short run, I think we can add a pod template into the event listener CR to address the need in workload level and don't bring in too many changes in the service configuration. 

And in the long run, I would prefer to refactor trigger code and adopt actor mode. 

How do you think? Any idea is appreciated. @dibyom

vdemeester:
This would be really nice to have this to be able to configure "resource request" for the event-listener pod :angel:

vdemeester:
> I wonder if a podTemplate would be enough though....is there a need for customizing things such as number of replicas etc.?
> Also, perhaps we might need to customize fields within the Service spec as well?

[...]

PR #597, opened 2020-06-05
add tolerations to eventlistener

GwonsooLee:
# Changes

Add `tolerations` to eventlistener. This is needed when I want to make event listener to the tainted node. 

# Submitter Checklist

review them:

# Release Notes

```
If all nodes are tainted for security and management issue, it is not possible to deploy event listener. (Of course.. it is possible to deploy by getting yaml file and modifying it, but then I have to keep two different files to manage)

If this feature is added, then user could add `tolerations` in EventListener crd, and deployments will be deployed with the same configuration.

```


dibyom:
Hi, @GwonsooLee Thanks for the PR! This looks related to https://github.com/tektoncd/triggers/issues/505

Could you add a `podTemplate` field to the EventListener that contains this `tolerations` field instead of directly adding the `tolerations` field in the EventListner spec?

GwonsooLee:
@dibyom Thanks for review. I added `podTemplate` to eventlistener and move the `tolerations` to it. In #505 , there are lots of things we could add to podTemplate, but this time I would like to add only tolerations feature because I still need time to deep dive into the code for better understanding :). I will try to add new features in podTemplate later. Please let me know if something's wrong wtih this PR.

examples/eventlisteners/eventlistener.yaml
```diff
@@ -5,6 +5,12 @@ metadata:
   name: listener
 spec:
   serviceAccountName: tekton-triggers-example-sa
+# podTemplate:
```
khrm, line 8: Add a separate example of eventlistener with tolerations

test/eventlistener_scale_test.go
```diff
@@ -47,8 +47,22 @@ func TestEventListenerScale(t *testing.T) {
 	// Create an EventListener with 1000 Triggers
 	var err error
 	el := bldr.EventListener("my-eventlistener", namespace, bldr.EventListenerSpec(
-		bldr.EventListenerServiceAccount(saName)),
-	)
+		bldr.EventListenerServiceAccount(saName),
+		bldr.EventListenerPodTemplate(
+			bldr.EventListenerPodTemplateSpec(
+				bldr.EventListenerPodTemplateTolerations(nil),
+				// If you want to test toleration feature, please use this commented part.
```
dibyom, line 54: Is the test commented out because it will fail in our CI runs?

test/eventlistener_test.go
```diff
@@ -227,6 +227,20 @@ func TestEventListenerCreate(t *testing.T) {
 			),
 			bldr.EventListenerSpec(
 				bldr.EventListenerServiceAccount(sa.Name),
+				bldr.EventListenerPodTemplate(
+					bldr.EventListenerPodTemplateSpec(
+						bldr.EventListenerPodTemplateTolerations(nil),
+						// If you want to test toleration feature, please use this commented part.
```
dibyom, line 233: Same question as above...is this failing in CI? If it is and it is really hard to test this on our current e2e setup, we can open an issue for it. But I think I'd prefer not having commented out bits of code

go.mod
```diff
@@ -13,7 +13,9 @@ require (
 	github.com/google/go-github/v31 v31.0.0
 	github.com/gorilla/mux v1.7.3
 	github.com/grpc-ecosystem/grpc-gateway v1.13.0 // indirect
+	github.com/onsi/ginkgo v1.10.2 // indirect
```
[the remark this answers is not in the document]
GwonsooLee: I think it was added indirectly when I ran dependency script. I ran e2e test without it and it worked so that I removed it.

test/eventlistener_scale_test.go
GwonsooLee: No.. it worked even though I uncommented it. I commented out those lines because I think test process cannot verify it really works because I guess there is no tainted node for e2e test. I uncommented every lines about tolerations in test files.

test/eventlistener_test.go
GwonsooLee: I left comment above! 👍

dibyom:
Just needs a rebase to resolve the go.mod conflicts

GwonsooLee:
I rebase master and I ran though e2e test.. but build test failed.. @dibyom, could you please let me know what I've been doing wrong..? I will try to figure out too! Thanks you in advance..

dibyom:
The problem is :
```
pkg/resources/create.go:47:8: G601: Implicit memory aliasing in for loop. (gosec)
		r := &apiResource
		     ^
```

Its strange that this popped up in this PR given that this line was not touched 🤔

dibyom:
Running gosec locally does not fail with this problem either 😕

GwonsooLee:
It seems a bug in gosec. Thank you so much for fixing this problem! 👍

PR #599, opened 2020-06-08
Use portforwarder instead of kubectl command line

### LINKED ISSUE #496: PortForward in e2e test is brittle
In /test/eventlistener_test.go TestEventListenerCreate, we port forward to the EventListener sink. To do this, we exec out to kubectl and sleep for a few seconds. A more robust way would be to use something like https://godoc.org/k8s.io/client-go/tools/portforward

/area testing
/kind cleanup

NavidZ:
# Changes

Make use of the portforward object instead of calling
into kubectl via command line exec.

This will remove the need for random sleep amount after
running the kubectl until port forwarding is ready and
make it more reliable.

Fixes #496

# Submitter Checklist

review them:


NavidZ:
I wasn't sure whether I should have included changes beyond the changes in eventlistener_test.go file. All those are the result of vendor update that this change needs to get newer version of k8s.io/client-go.

test/eventlistener_test.go
```diff
@@ -302,21 +305,39 @@ func TestEventListenerCreate(t *testing.T) {
 	// ElPort forward sink pod for http request
 	portString := strconv.Itoa(*eventReconciler.ElPort)
 	podName := sinkPods.Items[0].Name
-	cmd := exec.Command("kubectl", "port-forward", podName, "-n", namespace, fmt.Sprintf("%s:%s", portString, portString))
-	err = cmd.Start()
+
+	config, err := clientcmd.BuildConfigFromFlags("", knativetest.Flags.Kubeconfig)
+	if err != nil {
+		panic(err)
```
NavidZ, line 311: I just realized these "panic"s slipped through. I'll update them in the next patch.

test/eventlistener_test.go
```diff
@@ -302,21 +305,51 @@ func TestEventListenerCreate(t *testing.T) {
...
+			return
+		}
+		roundTripper, upgrader, err := spdy.RoundTripperFor(config)
+		if err != nil {
+			errChan <- err
+			return
+		}
+
+		path := fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/portforward", namespace, podName)
+		hostIP := strings.TrimLeft(config.Host, "htps:/")
```
dibyom, line 327: is the `htps:/` supposed to be `https:/`?

dibyom:
Would you mind squashing up the commits? Thanks!

test/eventlistener_test.go
NavidZ: It seem TrimLeft just gets a set of characters and remove all the ones from the beginning of the string. I can use the other function that only removes the exact keyword if you prefer.

NavidZ:
> Would you mind squashing up the commits? Thanks!

I had a question here. I also noticed this in other reviews. GitHub already has a squash button before merging when doing a manual merge but I guess since the bot merges these after an lgtm command it doesn't do it automatically.
So regarding doing this locally, is that alright to always do it? In other words do people in Tekton get upset if the history is lost between the comments in general (i.e. if I always just apply comments with a `commit --am` and force push it on this PR) or should I keep committing on top right until before merge and do one squash at the end?

test/eventlistener_test.go
wlynch: yeah, I would use `strings.TrimPrefix` instead, or call `url.Parse` and then set the path afterwards.

test/eventlistener_test.go
```diff
@@ -302,21 +305,51 @@ func TestEventListenerCreate(t *testing.T) {
...
 	defer func() {
-		if err = cmd.Process.Kill(); err != nil {
-			t.Fatalf("Error killing port-forward process: %s", err)
-		}
+		close(stopChan)
 	}()
-	// Wait for port forward to take effect
-	time.Sleep(5 * time.Second)
+	go func(stopChan chan struct{}, errChan chan error) {
+		readyChan := make(chan struct{}, 1)
```
wlynch, line 314: nit: move this below so that it's closer to where it's being used.

dibyom:
> So regarding doing this locally, is that alright to always do it? In other words do people in Tekton get upset if the history is lost between the comments in general (i.e. if I always just apply comments with a commit --am and force push it on this PR) or should I keep committing on top right until before merge and do one squash at the end?

We don't have a clear rule on this. For larger PRs, I do one squash at the end, for smaller ones I just keep ammending and pushing. I don't think anyone would be upset either way 😄

PR #609, opened 2020-06-15
Ensure that decodeB64 returns a String rather than Bytes.

### LINKED ISSUE #608: v0.5.0: CEL decodeb64: unexpected type 'bytes'
# Expected Behavior

I have an CEL overlay interceptor and this should work (docs say decodeb64 returns a string)

```
              - key: intercepted.id
                expression: "body.item1.decodeb64().parseJSON().appName"
```

# Actual Behavior

```
"caller":"sink/sink.go:145","msg":"failed to evaluate overlay expression 'body.item1.decodeb64().parseJSON().appName': unexpected type 'bytes' passed to parseJSON"
```
The docs say `decodeb64()` yields a string, but this error seems to indicate its passing along bytes

Workaround: `string(body.item1.decodeb64()).parseJSON().appName`

bigkevmcd:
# Changes

This changes the return value of the decodeB64 overload to ensure that it returns a string.

Fixes #608 

# Submitter Checklist

review them:

# Release Notes

```
Changes the output of decodeB64 to return a string, rather than bytes.
```

### DIFF
--- a/pkg/interceptors/cel/cel_test.go
+++ b/pkg/interceptors/cel/cel_test.go
@@ -120,14 +120,34 @@ func TestInterceptor_ExecuteTrigger(t *testing.T) {
 			want:    []byte(`{"replaced":"testing","new":"master","ref":"refs/head/master","name":"testing"}`),
 		},
 		{
-			name: "update with base64 decoding",
+			name: "decodeB64 with parseJSON",
+			CEL: &triggersv1.CELInterceptor{
+				Overlays: []triggersv1.CELOverlay{
+					{Key: "value", Expression: "body.value.decodeb64().parseJSON().test"},
+				},
+			},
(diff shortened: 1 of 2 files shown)


bigkevmcd:
I should probably have explained _why_ it was originally `types.Bytes` and not `types.String`, as it probably looks like a stupid bug :-)

When it's a string, it gets quoted when it's being inserted as an overlay value, what that means, is that previously, if you were decoding a base64 encoded JSON blob, the value that was inserted as, was a string, and not an object, so you couldn't use the fields.

Essentially, if you had something like `{"value": "test"}` encoded as base64, and you decoded it into an overlay as a `types.String` it would have looked like `"overlay": "{\"value\": \"test\"}"}`  and then, if you tried to get `body.overlay.value` in a binding, you couldn't.

But, with the addition of parseJSON, this isn't so much of a problem, as you can decode the base64 and then parse the JSON.

`body.value.decodeb64().parseJSON()`

PR #613, opened 2020-06-16
# Changes Link broken

jjasghar:
The getting-started link doesn't work, this links you to the README.md
on GitHub.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -83,7 +83,7 @@ To add the Tekton Triggers component to an existing cluster:
 
 You are now ready to create and run Tekton Triggers:
 
-- See [Tekton Triggers Getting Started Guide](./getting-started/) to
+- See [Tekton Triggers Getting Started Guide](https://github.com/tektoncd/triggers/blob/master/docs/getting-started/README.md) to
   get started.
 - Look at the
   [examples](https://github.com/tektoncd/triggers/tree/master/examples)


docs/install.md
```diff
@@ -83,7 +83,7 @@ To add the Tekton Triggers component to an existing cluster:
 
 You are now ready to create and run Tekton Triggers:
 
-- See [Tekton Triggers Getting Started Guide](./getting-started/) to
+- See [Tekton Triggers Getting Started Guide](https://github.com/tektoncd/triggers/blob/master/docs/getting-started/README.md) to
```
wlynch, line 86: Thanks for the PR! Unfortunately I don't think we want to do this.

This link is intentionally relative to enable pages to link to consistent versions. e.g. master will stay on master, v0.5.0 will stay on v0.5.0.

The problem here is that currently tekton.dev has an [explicit list of docs it syncs in](https://github.com/tektoncd/website/blob/master/sync/config/triggers.yaml) which doesn't include getting-started/README.md. Short term we should add the mapping there.

Long term I'm also working on bringing the tekton.dev source of truth to each individual repo (https://github.com/tektoncd/website/issues/140), which should help detect and catch issues like this earlier.

jjasghar: Ah ok, the link is busted busted, any advice on what it _should_ be?

wlynch: I think the link is correct, but the website is not importing the page. I think the fix is to modify https://github.com/tektoncd/website/blob/master/sync/config/triggers.yaml to make sure the page is being synced to the website, which should fix the 404.

jjasghar: Awesome, thanks, added PR to website.

jjasghar:
closing per https://github.com/tektoncd/website/pull/146

PR #591, opened 2020-06-02
Add URL parsing to CEL expressions.

bigkevmcd:
# Changes
This adds a new function `parseURL` to the CEL expression library.

This will act on a string, and parse it into a map which can be accessed from the remainder of the expression.

For example, `'https://example.com/testing#test'.parseURL().fragment == 'test'`

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [X] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [X] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [X] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

# Release Notes

```
Add URL parsing to CEL expressions.
    
This adds a new function `parseURL` to the CEL expression library.
    
This will act on a string, and parse it into a map which can be accessed from the remainder of the expression.
    
For example, 'https://example.com/testing#test'.parseURL().fragment == 'test'
```

This addresses #590

PR #625, opened 2020-06-19
add nodeSelector and replicas to eventlistener

GwonsooLee:
# Changes

Add nodeSelector and replicas feature to eventListener. With this, user could schedule eventListener pod to the node with specific label. Also, if needed, user could specify the number of replicas in yaml file. Previously I scaled pod by editing after deployment ( If other possible way to scale which I might not know..., please let me know!! 😄  ) 

# Submitter Checklist

review them:

# Release Notes

```
- add replicas field for event listener to be able to run more than one pod.
- add nodeSelector field for event listener to schedule event listener pod to the specific node.
```

### DIFF
--- a/pkg/apis/triggers/v1alpha1/event_listener_types.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_types.go
@@ -57,11 +57,20 @@ type EventListenerSpec struct {
 	ServiceAccountName string                 `json:"serviceAccountName"`
 	Triggers           []EventListenerTrigger `json:"triggers"`
 	ServiceType        corev1.ServiceType     `json:"serviceType,omitempty"`
+	Replicas           int32                  `json:"replicas"`
 	PodTemplate        PodTemplate            `json:"podTemplate,omitempty"`
 }
 
 type PodTemplate struct {
+	// If specified, the pod's tolerations.
+	// +optional
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// NodeSelector is a selector which must be true for the pod to fit on a node.
+	// Selector which must match a node's labels for the pod to be scheduled on that node.
+	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 }
 
 // EventListenerTrigger represents a connection between TriggerBinding, Params,
--- a/third_party/github.com/hashicorp/golang-lru/simplelru/lru_interface.go
+++ b/third_party/github.com/hashicorp/golang-lru/simplelru/lru_interface.go
@@ -34,6 +34,6 @@ type LRUCache interface {
 	// Clears all cache entries.
 	Purge()
 
-  // Resizes cache, returning number evicted
-  Resize(int) int
+	// Resizes cache, returning number evicted
+	Resize(int) int
 }
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -16,6 +16,7 @@ using [Event Interceptors](#Interceptors).
 
 - [Syntax](#syntax)
   - [ServiceAccountName](#serviceAccountName)
+  - [Replicas](#replicas)
   - [PodTemplate](#podTemplate)
   - [Triggers](#triggers)
     - [Interceptors](#interceptors)
@@ -45,6 +46,7 @@ the following fields:
 - Optional:
   - [`serviceType`](#serviceType) - Specifies what type of service the sink pod
     is exposed as
+  - [`replicas`](#replicas) - Specifies the number of EventListener pods
   - [`podTemplate`](#podTemplate) - Specifies the PodTemplate
     for your EventListener pod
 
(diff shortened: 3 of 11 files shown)


GwonsooLee:
@dibyom Is the reason of failure related to the nodeSelector..?  I added nodeSelector to test file with `beta.kubernetes.io/arch": "amd64`. If not then, I will figure out what's wrong more. Thanks in advance

dibyom:
> @dibyom Is the reason of failure related to the nodeSelector..? I added nodeSelector to test file with `beta.kubernetes.io/arch": "amd64`. If not then, I will figure out what's wrong more. Thanks in advance

Doesn't look like...looks like some test has a data race

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -57,11 +57,20 @@ type EventListenerSpec struct {
 	ServiceAccountName string                 `json:"serviceAccountName"`
 	Triggers           []EventListenerTrigger `json:"triggers"`
 	ServiceType        corev1.ServiceType     `json:"serviceType,omitempty"`
+	Replicas           int32                  `json:"replicas"`
```
khrm, line 60: Should Replica be part of PodTemplate? Maybe in future have a ducktype like PodSpecable mentioned here: 
 https://github.com/tektoncd/triggers/issues/505#issuecomment-620205840
Of course, we want to give only limited set of capability even if we do ducktype.
@dibyom

GwonsooLee:
> I think syncing with Master would resolve that test failure. We solve that Data Race here: #631

I rebase the master branch so that the error message was changed. I think there is another problem... I will try to figure it out.. Thank you so much @khrm :)

third_party/github.com/hashicorp/golang-lru/simplelru/lru_interface.go
```diff
@@ -34,6 +34,6 @@ type LRUCache interface {
 	// Clears all cache entries.
 	Purge()
 
-  // Resizes cache, returning number evicted
-  Resize(int) int
+	// Resizes cache, returning number evicted
+	Resize(int) int
```
dibyom, line 38: 👀

pkg/apis/triggers/v1alpha1/event_listener_types.go
dibyom: So, my understanding is that for PodTemplate or PodSpecable, the object can contain fields from a `PodSpec`: https://godoc.org/k8s.io/api/core/v1#PodSpec

`replicas` are part of the DeploymentSpec -> looking through the fields it doesn't seems like we'd want to support customizing any fields besides `replicas`. So, maybe we just leave `replicas` as part of the EL spec (maybe we can call it `deploymentReplicas` ?) What do you think @khrm ?

dibyom:
Thanks for adding this @GwonsooLee 
Looks like the integration test failure is:
```
The EventListener "listener-podTemplate" is invalid: metadata.name: Invalid value: "listener-podTemplate": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

examples/eventlisteners/eventlistener-podTemplate.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: EventListener
+metadata:
+  name: listener-podTemplate
+spec:
+  serviceAccountName: tekton-triggers-example-sa
```
dibyom, line 7: Thanks for adding examples, can you also update the docs to mention the new fields?

third_party/github.com/hashicorp/golang-lru/simplelru/lru_interface.go
GwonsooLee: I think this is because I ran one of the scripts in `./hack` directory...

examples/eventlisteners/eventlistener-podTemplate.yaml
GwonsooLee: I added `replicas` and `podTemplate.nodeSelector` to docs. :)

GwonsooLee:
> Thanks for adding this @GwonsooLee
> Looks like the integration test failure is:
> 
> ```
> The EventListener "listener-podTemplate" is invalid: metadata.name: Invalid value: "listener-podTemplate": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
> ```

Thanks for finding the error! I fixed it :) 
Is there  any other way to run integration test locally??

GwonsooLee:
> Could you add a `Release Notes` section to this PR description ? Very helpful when we are doing a release!
> for an example, see: #591

Oh.. I am sorry that I missed this one till now.. I will update this section from now on whenever I open PR. Thanks @dibyom

dibyom:
> Is there any other way to run integration test locally??

In general, for go e2e tests, you can run something like ` go test -count=1 -v -tags=e2e ./...`

In addition, we have some tests, that apply our YAML examples to make sure they are valid: `https://github.com/tektoncd/triggers/blob/master/test/e2e-tests-yaml.sh`

I think you should be able to run the `e2e-tests-yaml.sh` script locally as well too!

pkg/apis/triggers/v1alpha1/event_listener_types.go
dibyom: I'll add this to tomorrow's WG agenda. Unless there are objections, I feel inclined to keep the current field naming!

GwonsooLee: Thanks :) @dibyom

Issue #624, opened 2020-06-19
CLI tool for manual Trigger runs


dibyom:
#  Expected Behavior

I can easily test out their Trigger config e.g. does the CEL expression work, does my binding expression extract the right variable, does my trigger template replace the right variable etc.

# Actual Behavior

I have to setup an EventListener, TriggerBinding, and TriggerTemplate. I have to port-forward to the Listener, and make a curl call, and then look into the Listener logs to figure out what happened.

# Additional Info

Example CLI surface from discussing with @bigkevmcd :

```
triggers-run -H "X-GitHub-Event: testing" -d @testing.json -f triggers.yaml
```

1. triggers.yaml is a file containing the trigger definitions
2. The CLI supports curl like flags for sending headers and body
3. The tool will prints out the trigger processing logs to stdout
4. It will also print out the fully formed Tekton resource to be created once the trigger is processed instead of creating them

Out of scope for now:
1. Testing Trigger level service account permissions

Can be added later:
1. Actually create the resources instead of just printing them out
2. Dry run against a already running listener instead of definitions from a file
3. Make it part of the `tkn` surface e.g. `tkn trigger run ....`
4. Account for the Trigger service account permissions

Related :
1. TriggerTemplateRun #200 
2. Binding eval tool #314  

/kind feature

vdemeester:
Any reason for not pushing for having this in `tkn` early ? :thinking: 
/cc @danielhelfand @pradeepitm12 @piyush-garg

danielhelfand:
> Any reason for not pushing for having this in `tkn` early ? 🤔

Reading this, it sounds like more of a proof of concept/basic requirements at this point and could potentially be added to `tkn` later on (e.g. `tkn trigger run` under `Can be added later`) if understanding correctly? Either as part of `tkn` itself or maybe could be added as some type of plugin to `tkn` if that would be a more preferred route. 

But I would see no reason not to want something like this as part of `tkn`. Just would maybe need some thought on where it would best fit structurally.

dibyom:
yeah, one motivation for keeping this in Triggers would be faster/easier iteration -- we'd have to do some refactoring to how we execute a Trigger to ensure we can run it outside the context of an EventListener. Once that is done, it should be easy enough to make it part of the `tkn` surface! 
(The plugin model sounds cool too!!)

dibyom:
We have some experimental code for this that processes Triggers locally. We should do a release for it and then figure out if it makes sense to merge that into the CLI

PR #628, opened 2020-06-23
Add a new CRD type called Trigger

dorismeixing:
# Changes

/kind misc

 Adds the Trigger type. It has not been wired up and cannot be used yet. 

# Submitter Checklist

review them:


dorismeixing:
add reviewer @dibyom

pkg/apis/triggers/v1alpha1/trigger_types.go
```diff
@@ -0,0 +1,85 @@
...
+
+// TriggerSpec represents a connection between TriggerBinding, Params,
+// and TriggerTemplate; TriggerBinding provides extracted values for
+// TriggerTemplate to then create resources from.
+type TriggerSpec struct {
+	Bindings []*TriggerBinding `json:"bindings"`
+	Template TriggerTemplate   `json:"template"`
+	// +optional
+	Name         string                `json:"name,omitempty"`
+	Interceptors []*TriggerInterceptor `json:"interceptors,omitempty"`
```
khrm, line 32: I don't think Interceptor should be like this. Keep like it's there in EventListener Spec Triggers.

pkg/apis/triggers/v1alpha1/trigger_types.go
```diff
@@ -0,0 +1,85 @@
...
+// +k8s:openapi-gen=true
+type Trigger struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Spec holds the desired state of the Trigger
+	// +optional
+	Spec TriggerSpec `json:"spec"`
+	// +optional
+	Status TriggerStatus `json:"status,omitempty"`
```
khrm, line 64: I think for now we don't require TriggerStatus or TriggerCRD controller.

pkg/apis/triggers/v1alpha1/trigger_types.go
savitaashture, line 32: I think as per the design discussion we are on consensus that there may be more the one EventListener support  so it would be better if we move the structure definition from event_listener_type.go to this file itself and keeping things which are relevant there

dorismeixing: After discussing with the team, we will move the structure definition from event_listener_type.go to this file itself.

pkg/apis/triggers/v1alpha1/trigger_types.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// TriggerInterceptor provides a hook to intercept and pre-process events
+type TriggerInterceptor struct {
```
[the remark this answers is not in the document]
dibyom: Yup! Let's move all of the interceptor types below to this file as well. In addition, let's remove the EventInterceptor type from the other file and switch it to use the `TriggerInterceptor` type from this file.

pkg/apis/triggers/v1alpha1/trigger_types.go
dibyom: fair enough, let's remove this for now. We can always add it later!!

pkg/apis/triggers/v1alpha1/trigger_types.go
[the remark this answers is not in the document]
dibyom: Let's do the following:
1. move WebhookInterceptor and all the other types to this file.
2. Remove the EventInterceptor type and replace it with the TriggerInterceptor from here:
https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_types.go#L89

dibyom:
The integration test failure looks unrelated. 
Once you address @khrm 's comment and squash the two commits into one, we are good to go!

config/200-clusterrole.yaml
```diff
@@ -35,6 +35,9 @@ rules:
   - apiGroups: ["triggers.tekton.dev"]
     resources: ["clustertriggerbindings/status", "eventlisteners/status", "triggerbindings/status", "triggertemplates/status"]
     verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
+  - apiGroups: ["triggers.tekton.dev"]
```
dibyom, line 38: Instead of adding a new rule here, we can just add "triggers" to the rule in [line 33](https://github.com/tektoncd/triggers/blob/master/config/200-clusterrole.yaml#L33) and the status one in the following rule: https://github.com/tektoncd/triggers/blob/master/config/200-clusterrole.yaml#L36

dibyom:
> Do we know why the data in third_party was deleted? IIRC we need this to include license information in released images via kodata

FWIW, when I run `./hack/update-deps.sh` on master, I get the same result i.e. the same license files are deleted 🙃

dibyom:
I opened https://github.com/tektoncd/triggers/issues/653 to track the LICENSE removal bug.

bobcatfish:
Hey @dorismeixing @wlynch @dibyom  a few requests:
- Is there a design doc and/or issue that accompanies this?
- Can there be usage docs on how to use this?

dorismeixing:
Hi @bobcatfish, I think it's related to #624 and it will be used in Trigger CLI, but I am not sure if there is a design doc. I can add an usage doc if needed.

bobcatfish:
Hey @dorismeixing !

>  I think it's related to #624 and it will be used in Trigger CLI

Looking at #624 I don't see a design doc there either 🤔  I'm a bit surprised that we'd add a new CRD without a design doc - maybe @wlynch or @dibyom when he's back have more info?

> I can add an usage doc if needed.

yes plz :D usually we try to accompany every commit with the related docs changes (https://github.com/tektoncd/community/blob/master/standards.md#principles), so since we've added a new type here I'd expect to see some docs at https://github.com/tektoncd/triggers/tree/master/docs about how to use it

dibyom:
Triggers WG has been discussing a Trigger CRD in https://docs.google.com/document/d/1NX0ExhPad6ixTM8AdU0b6Vc3MVD5hQ_vIrOs9dIXq-I/edit# and https://docs.google.com/document/d/1zWpmEhtSNe8KAPKvTJE7Pjg5Uzk8mx9MUjSN24D1jUg/edit# 

We reached consensus on a Trigger CRD + selector approach in the Triggers WG and a TEP has been filed with the proposed approach: https://github.com/tektoncd/community/pull/148

This PR adds the type but it is not ready to be used by an end user yet. The main motivation for adding the type now was to simplify the run CLI tool while we worked out the open questions/TEP process for the overall Trigger/Selector approach. The CRD will be usable once we either have the run CLI implemented or we implement the approach from the TEP.

Issue #640, opened 2020-06-26
CEL feature: URL extension


andrey-carerix:
To make a secure integration with Bitbucket Cloud it is necessary to inspect the incoming webhook URL query parameters to extract pre-shared secret value (e.g. https://mywebhook/?token=icWo3SV9lbx0lm4n) to be validated with existing CEL extension function compareSecret.

dibyom:
Hi @andrey-carerix we recently added a `parseURL()` function in https://github.com/tektoncd/triggers/pull/591
I think that should do the trick?
/kind question

andrey-carerix:
@dibyom yes, I am aware of that function. But it's meant to parse some attributes within the json body. I am looking for a way to get a reference to the URL the webhook listener was invoked with by a callee (e.g. bitbucket cloud service). Basically URL extension that gives me access to the original HTTP request URL object similar to what body and header extensions provide.

dibyom:
ahh I see, we need to expose the url as well as the body and headers

andrey-carerix:
@dibyom exactly. Here is the real-life use case to support this feature request:

- bitbucket cloud has a webhook configured with a secret as query parameter, e.g. https://mywebhook/?token=icWo3SV9lbx0lm4n. Note: unlike bitbucket server there is no other way to secure your webhook endpoint other than bake a pre-shared secret as URL query parameter.
- bitbucket cloud invokes the above webhook URL whenever a change occurs in git repo
- tekton eventlistener with CEL interceptor receives the incoming HTTP request from bitbucket cloud 
- cel filter is configured to read the query parameter 'token' and match it against k8s secret to verify the authenticity of the webhook request

The desired behaviour as pseudo CEL expression:

`url.query('token').compareSecret('namespace', 'bitbucket-secret', 'tokenKey') `

dibyom:
yeah, currently we have `$(body)` and `$(header)`. We can add a `$(url)` or `$(fullUrl)` field that is accessible to the bindings.

PR #647, opened 2020-06-29
Expose the incoming request URL to CEL expressions.

bigkevmcd:
# Changes

This adds a `requestURL` key to the CEL expression context.

For example, `requestURL.parseURL().path` would get the path of the incoming request URL.

Fixes: #640 

# Submitter Checklist

review them:

# Release Notes

```
This adds a `requestURL` key to the CEL expression context.

For example, `requestURL.parseURL().path` would get the path of the incoming request URL.
```

### DIFF
--- a/docs/cel_expressions.md
+++ b/docs/cel_expressions.md
@@ -139,6 +139,20 @@ headers are also available.
       <pre>header['X-Test'][0] == 'test-value'</pre>
     </td>
   </tr>
+  <tr>
+    <th>
+      requestURL
+    </th>
+    <td>
+      string
+    </td>
+    <td>
+      This is the URL for the incoming HTTP request.
+    </td>
+    <td>
+      <pre>requestURL.parseURL().path</pre>
+    </td>
+  </tr>
 </table>
 
 NOTE: The header value is a Go `http.Header`, which is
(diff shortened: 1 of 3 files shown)


dibyom:
So, this adds `requestURL` to the CEL context. Wondering if it makes sense to add it to the TriggerBinding context as well -- I guess the only use case right now would be to use the entire url...unless we add CEL functions to TriggerBindings

/approve
/kind feature

bigkevmcd:
> So, this adds `requestURL` to the CEL context. Wondering if it makes sense to add it to the TriggerBinding context as well -- I guess the only use case right now would be to use the entire url...unless we add CEL functions to TriggerBindings

There's arguably value in having the requestURL in trigger bindings, _if_ CEL functions landed in TriggerBindings, I'd be looking to refactor and reuse the same context mechanism.

dibyom:
Yeah, let's just go with the CEL context for now.

lostiniceland:
I was hoping to parse query parameters using this function to easily trigger a pipeline in tekton (no need for curl or postman etc, just a browser).
I am able to use the cel filter in an EventListener but without support for this in a TriggerBinding it is useless for my case

bigkevmcd:
@lostiniceland you can expose the elements you need as overlays via a CEL interceptor and then use those in a binding https://tekton.dev/docs/triggers/interceptors/#cel-interceptors and https://tekton.dev/docs/triggers/triggerbindings/#accessing-data-added-by-interceptorsdocstriggersinterceptors ?

PR #694, opened 2020-07-31
Fix GitHub example bindings

dibyom:
# Changes

The old bindings were for the GitHub push event while this EventListener is
configured for pull_request event types.  See:
https://developer.github.com/webhooks/event-payloads/#pull_request

(I also moved the bindings and template to the same file to be consistent with the GitLab example)

# Submitter Checklist

review them:

# Release Notes

```
NONE
```


examples/github/github-eventlistener-interceptor.yaml
```diff
@@ -17,6 +17,53 @@ spec:
...
+                #! /bin/bash
+                ls -al $(inputs.resources.source.path)
+        inputs:
+          resources:
+            - name: source
+              resourceSpec:
+                type: git
+                params:
+                  - name: revision
+                    value: $(params.gitrevision)
```
savitaashture, line 67: Nice fix :+1: how its use to work earlier :thinking: 

small changes related to `tt.params`
```suggestion
                    value: $(tt.params.gitrevision)
```

dibyom: Done!
We were using a mock/fake body to test this example out. Turns out that body was wrong  -- it was for the push event not pull_request too. So, the example worked. When tried with a real GitHub PR, it failed 😅

PR #698, opened 2020-08-04
Add namespace to the commands in Nginx ingress doc

NavidZ:
Add namespace to the commands in Nginx ingress doc to get
them to work.

### DIFF
--- a/docs/exposing-eventlisteners.md
+++ b/docs/exposing-eventlisteners.md
@@ -34,7 +34,7 @@ services can be found
    ```
 2. Find the service name expose by the Eventlistener service:
    ```sh
-    kubectl get el <EVENTLISTENR_NAME> -o=jsonpath='{.status.configuration.generatedName}'
+    kubectl get el <EVENTLISTENR_NAME> -o=jsonpath='{.status.configuration.generatedName}{"\n"}'
    ```
 3. Create the Ingress resource. A sample Ingress is below. Check the docs
    [here](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/)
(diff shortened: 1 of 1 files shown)


docs/exposing-eventlisteners.md
```diff
@@ -32,9 +32,10 @@ services can be found
       # Check https://kubernetes.github.io/ingress-nginx/deploy for other installation options
       kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v0.34.1/deploy/static/provider/cloud/deploy.yaml
    ```
-2. Find the service name expose by the Eventlistener service:
+2. Find the service name expose by the Eventlistener service (assuming
+   it is under namespace `getting-started`):
    ```sh
-    kubectl get el <EVENTLISTENR_NAME> -o=jsonpath='{.status.configuration.generatedName}'
+    kubectl get el <EVENTLISTENR_NAME> -n getting-started -o=jsonpath='{.status.configuration.generatedName}{"\n"}'
```
dibyom, line 38: Thanks for adding this. Usually in most of our examples/snippets, we generally omit the namespace parameter since it is a common parameter that applies to almost all kubectl commands. However, I do realize that it is confusing for users following the `getting-started` tutorial since it uses a non-default namespace. What do you think of having both i.e. one snippet with no namespace with a note/comment below with an example mentioning that users should add a `-n <namespace>` param for their commands if needed?

NavidZ: I agree with you. I realized people might be coming here from other places beside the tutorial. But then I saw the yaml right after this that installs Ingress in getting-started namespace. So I thought maybe not. 
But I'm totally fine with your solution. what should we do with the Ingress resource?

dibyom: I'm fine with a quick note there to indicate that namespace might be needed too!

PR #701, opened 2020-08-04
Use Kaniko instead of img in tutorial to fix permission issue

### LINKED ISSUE #615: getting-started tutorial is failing
# Expected Behavior

The Pipeline in the getting-started guide work (it used to on this setup!) 

# Actual Behavior

The task fails with two  errors:

1. The condition fails with:

```
[condition-check-go-test] can't load package: package /workspace/source-repo: import "/workspace/source-repo": cannot import absolute path

```
I resolved this by setting adding `workingDir: "$(resources.source-repo.path)"` to the path

2. The build-docker-image task fails with:
```
nsenter: cannot change root filesystem propagation: Permission denied
```

Haven't been able to fix this yet

1. Go through the tutorial in https://github.com/tektoncd/triggers/tree/master/docs/getting-started . 

# Additional Info

Pipeline version: v0.13.2
Triggers version: v0.5.0
K8s versions:
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.4", GitCommit:"224be7bdce5a9dd0c2fd0d46b83865648e2fe0ba", GitTreeState:"clean", BuildDate:"2019-12-11T12:47:40Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.8-gke.15", GitCommit:"9cabee15e0922c3b36724de4866a98f6c2da5e6a", GitTreeState:"clean", BuildDate:"2020-05-01T21:47:04Z", GoVersion:"go1.13.8b4", Compiler:"gc", Platform:"linux/amd64"
```

NavidZ:
Use Kaniko instead of img in tutorial to fix permission issue

Fixes #615

### DIFF
--- a/docs/getting-started/pipeline.yaml
+++ b/docs/getting-started/pipeline.yaml
@@ -85,6 +85,10 @@ spec:
     description:
       The build directory used by img
     default: /workspace/source-repo
+  - name: pathToDockerFile
+    type: string
+    description: The path to the dockerfile to build
+    default: $(resources.inputs.source-repo.path)/Dockerfile
   resources:
     inputs:
       - name: source-repo
@@ -94,15 +98,13 @@ spec:
         type: image
   steps:
     - name: build-and-push
-      image: r.j3ss.co/img
+      image: gcr.io/kaniko-project/executor:v0.16.0
       command:
-        - /usr/bin/img
+        - /kaniko/executor
       args:
-        - build
-        - -t
-        - "$(resources.outputs.builtImage.url)"
-        - --no-cache
-        - "$(params.pathToContext)"
+        - --dockerfile=$(params.pathToDockerFile)
+        - --destination=$(resources.outputs.builtImage.url)
+        - --context=$(params.pathToContext)
 ---
 apiVersion: tekton.dev/v1alpha1
 kind: Condition
@@ -116,7 +118,8 @@ spec:
   check:
     image: golang
     command: ["go"]
-    args: ['test', "$(resources.source-repo.path)/..."]
+    workingDir: "$(resources.source-repo.path)"
+    args: ['test', "./..."]
 ---
 # Finally, we need something to receive our cloudevent announcing success!
 # That is this services only purpose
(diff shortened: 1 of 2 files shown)


NavidZ:
cc @dibyom

docs/getting-started/pipeline.yaml
```diff
@@ -94,15 +98,13 @@ spec:
         type: image
   steps:
     - name: build-and-push
-      image: r.j3ss.co/img
+      image: gcr.io/kaniko-project/executor:v0.16.0
       command:
-        - /usr/bin/img
+        - /kaniko/executor
```
dibyom, line 103: Do we need to do anything here to setup the docker registry creds for Kaniko? 
The pipeline tutorial has users setting up a secret and a service account: https://github.com/tektoncd/pipeline/blob/master/docs/tutorial.md#configuring-task-execution-credentials

Do we need to do the same?

NavidZ: Good point. It might come down to the k8s instance and the overall setup we run in. I used GKE and it seemed to work with this yaml. Maybe my default user/SA has the full access or something by default and no need to do that. But if you do use another service account maybe we would need that.

dibyom: Yeah, I think that's why it works. We should point that out in a note with maybe a link to the section on setting up registry auth in the pipelines tutorial?

NavidZ: Yup. I had a chat with @wlynch and he also mentioned on GKE without workload identity setup it goes through the default GCE that has the editor access and it just works. So I added a note to the readme to cover that case. Let me know if you think that is sufficient.

dibyom:
Thanks @NavidZ 
/approve
/lgtm

docs/getting-started/pipeline.yaml
dibyom: Yup, just verified the docs: https://cloud.google.com/container-registry/docs/using-with-google-cloud-platform#gke-permissions

PR #704, opened 2020-08-10
switch trigger sa ref from global to namespace scoped

gabemontero:
# Changes

So in getting caught up on some thing related to the trigger crd tep, I came across this type and https://github.com/tektoncd/triggers/pull/628

It appears to me that to some extent there was a simple copy done in https://github.com/tektoncd/triggers/pull/628 of https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_types.go#L77-L92
including my verbose comment and TODO around the `ServiceAccount` field.

Now, as already indicated in https://github.com/tektoncd/community/pull/148 and the use of a string type `serviceAccountName`, as well in the WG discussion, any SA assoicated with a Trigger should be in the same namespace as the trigger

So why make this correction now vs. just waiting until the TEP is implemented?

@dibyom 's comment https://github.com/tektoncd/triggers/pull/628#issuecomment-661103086 is what motivated me to submit this for consideration.

If the "run CLI tool" item noted there drops with sufficient time before the tep does, then we don't want the precedent of triggers using SAs from other namespaces being established.

Thoughts @khrm @dibyom @dorismeixing  ?

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/apis/triggers/v1alpha1/trigger_types.go
+++ b/pkg/apis/triggers/v1alpha1/trigger_types.go
@@ -32,14 +32,10 @@ type TriggerSpec struct {
 	Name         string                `json:"name,omitempty"`
 	Interceptors []*TriggerInterceptor `json:"interceptors,omitempty"`
 	// ServiceAccount optionally associates credentials with each trigger;
-	// more granular authorization for
-	// who is allowed to utilize the associated pipeline
-	// vs. defaulting to whatever permissions are associated
-	// with the entire EventListener and associated sink facilitates
-	// multi-tenant model based scenarios
-	// TODO do we want to restrict this to the event listener namespace and just ask for the service account name here?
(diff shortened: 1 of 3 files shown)


gabemontero:
BTW, reducing the scope of https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_types.go#L91 could also be a consderation, though maybe that is a bit muddied by any pre https://github.com/tektoncd/community/pull/148 attempts to have a single event listener in one namespace manage triggers / interceptors in multilpe namespaces

My thought was to tackle that possible discussion separately, but don't feel strongly about that.

dibyom:
SGTM, the CLI basically ignores the SA field at the moment but this makes it future proof

/approve

Issue #679, opened 2020-07-19
Config to turn-off Trigger's namespaced SA impersonation


skaegi:
Is there a config parameter or someway to prevent users from using the feature that let's cross-namespace SA impersonation.  This is very hard to secure and let's anyone who has privs to create an EventListener in one namespace the ability to impersonate any other user in any namespace like a cluster admin and steal their credentials. Am I misunderstanding this or is there a feature flag to turn this off?

dibyom:
> This is very hard to secure and let's anyone who has privs to create an EventListener in one namespace the ability to impersonate any other user in any namespace like a cluster admin and steal their credentials.

Can you elaborate on this? Is it that one user can in theory create a EL with a serviceAccountName that they might not have access to? In that case, how is it different from the Pipelines scenario where a user can create a Taskrun with a SA they don't have access to. In the Triggers sceanrio, all that a user can do is create a PipelineRun/TaskRun while in Pipelines, the user can actually run arbitrary code?

skaegi:
I’m totally fine with the ability for an EL to impersonate an SA “in the same namespace”. Anyone who can create a pod via TaskRun, Job, StatefulSet, etc can do that today. 

My concern is about the Trigger’s ability to impersonate an SA “in another namespace” . That is a cluster-wide super power and something no other built-in k8 CRD allows. I would consider removing this ability as I believe it is a mis-feature but at the very least let configure the feature off as it’s a security hazard.

PR #705, opened 2020-08-10
switch trigger sa based auth to impersonate; change secret ref to string and by extension same namespace as EventListener

gabemontero:
# Changes

As the efforts to progress support for multi-tenant event listeners has evolved, most notably around the notion of breaking out `EventListenerTrigger` into a Trigger CRD, or an analogous resource like a PipelineRecipe defined in the pipelines project, some evolution of the finer grained auth originally introduced in https://github.com/tektoncd/triggers/pull/454 seemed appropriate.

This PR currently has 2 commits with that goal in mind:
- switch the secret reference originally introduced in #454 from a `ObjectReference` to a `LocalObjectReference`, hence scoping the SA's available for use to the SA's present in the namespace of the trigger object itself; we originally scoped it as `ObjectReference` because the interceptors had a a multi-namespace scope for secrets per https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_types.go#L97-L104 however as the various TEPs/proposals have evolved that no longer seems like a valid motivator for having this SA reference be multi namespace
- as a result, the implementation for creating tekton objects under the permission of the referenced SA has been switched to use impersonation vs. construction of an actual client; this allowed for some positive changes in the roles required, but also should fit in well with how the upcoming TEPs / proposals will get implemented

And as a result this

Fixes https://github.com/tektoncd/triggers/issues/679

# Submitter Checklist

review them:

- [/ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ /] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ /] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

# Release Notes

- BREAKING CHANGE: the optional EventListenerTrigger based level of authentication for creating Tekton object has had its ServiceAccount reference changed from an ObjectReference to a LocalObjectReference, effectively enforcing that the ServiceAccount be in the same namespace as the EventListenerTrigger


gabemontero:
/assign @dibyom 
/assign @khrm 

details in the description gentlement

similar in spirit to https://github.com/tektoncd/triggers/pull/704 but a more comprehensive change as it evolves the support I added back in Feb / Mar of this year.

dibyom:
Thanks @gabemontero I think this should address https://github.com/tektoncd/triggers/issues/679 too?

gabemontero:
> Thanks @gabemontero I think this should address #679 too?

yes it does @dibyom

gabemontero:
forgot to

/hold cancel

gabemontero:
removed the WIP designation in one of the commit messages and re-pushed

pkg/sink/sink.go
```diff
@@ -164,12 +165,12 @@ func (r Sink) processTrigger(t *triggersv1.EventListenerTrigger, request *http.R
 
 	log.Infof("ResolvedParams : %+v", params)
 	resources := template.ResolveResources(rt.TriggerTemplate, params)
-	token, err := r.retrieveAuthToken(t.ServiceAccount, eventLog)
+	/*token, err := r.retrieveAuthToken(t.ServiceAccount, eventLog)
```
dibyom, line 168: remove commented line?

dibyom:
This is technically a breaking change. We should mention that in the release notes

gabemontero:
> This is technically a breaking change. We should mention that in the release notes

Agreed.

I've got a blurb in this PR's description release note section @dibyom 

What if any are the additional steps I need to take for end to end closure on release notes?  Or is the text in the PR subscription sufficient? thanks.

pkg/sink/sink.go
gabemontero: oops ... yep, will do just a sec

gabemontero:
removed commented out code @dibyom

dibyom:
> What if any are the additional steps I need to take for end to end closure on release notes? Or is the text in the PR subscription sufficient? thanks.

I think adding the words BREAKING CHANGE in the Release Notes section is sufficient so that the release captain can easily spot it.

/lgtm

dibyom:
Hmm, coverage dropped off a bit...I wonder if this is because the test for this is a e2e test..though we've had issues with the coverage tool not being accurate in the past so 🤷‍♂️

gabemontero:
> > What if any are the additional steps I need to take for end to end closure on release notes? Or is the text in the PR subscription sufficient? thanks.
> 
> I think adding the words BREAKING CHANGE in the Release Notes section is sufficient so that the release captain can easily spot it.
> 
> /lgtm

ok cool ... BREAKING CHANGE added to release notes section .. thanks @dibyom

gabemontero:
> Hmm, coverage dropped off a bit...I wonder if this is because the test for this is a e2e test..though we've had issues with the coverage tool not being accurate in the past so man_shrugging

maybe cause all the code I removed in auth_override.go was previously covered, and that messed with the overall percentage? ... just a guess

gabemontero:
is there anything else you think we need @dibyom before you apply the approve label and we merge this ?

dibyom:
@khrm was assigned so I wasn't sure if he wanted to take a look too 👼

gabemontero:
> /lgtm
> 
> I think we should move to `ServiceAccountName` string instead of taking it as `ServiceAccount` LocalObjectReference. Maybe a pr later on which deprecate `ServiceAccount` in favour of `ServiceAccountName`.
> 
> Then, this will have the same format as we do in other Tekton resources which requires ServiceAccount.

I'm fine with switching to `ServiceAccountName` with this PR.  I'll do that momentarily, and then between the two of you @khrm @dibyom unless anything else comes to mind, feel free to both  approve and lgtm.  Of course if @dibyom has issue with the move off of LocalObjectReference we can discuss further.

thanks

gabemontero:
switch from LocalObjectReference to string pushed

adjusted PR title

Also added Fixes #679 note in description

@dibyom @khrm PTAL

gabemontero:
oh and I removed more unneeded code from auth_override.go, hence the delta change

gabemontero:
tests back to green after the switch to `serviceAccountName string`

PTAL @khrm @dibyom

dibyom:
> oh and I removed more unneeded code from auth_override.go, hence the delta change

> we are now down to one method which we have to use the "fake" version of in the unit tests to bypass actual client access to a cluster

> the function is covered by the e2e that validates trigger level SA

Thanks for the clarification!

PR #708, opened 2020-08-11
Add createResource in TriggerRun CLI

dorismeixing:
# Changes

This is to add create resource feature to TriggerRun CLI. 
(related to issue #624 )

# Submitter Checklist

review them:

# Release Notes

```
NONE - experimental feature

```

### DIFF
--- a/cmd/triggerrun/cmd/root.go
+++ b/cmd/triggerrun/cmd/root.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 
@@ -31,13 +32,17 @@ import (
 	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	triggersclientset "github.com/tektoncd/triggers/pkg/client/clientset/versioned"
+	dynamicClientset "github.com/tektoncd/triggers/pkg/client/dynamic/clientset"
+	"github.com/tektoncd/triggers/pkg/client/dynamic/clientset/tekton"
 	"github.com/tektoncd/triggers/pkg/sink"
 	"github.com/tektoncd/triggers/pkg/template"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/runtime/serializer/streaming"
 
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/yaml"
 )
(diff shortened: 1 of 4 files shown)


dorismeixing:
add reviewers
/assign dibyom
/assign wlynch

dibyom:
Looks like a few commits on master that also got added here. 
I think rebasing from the latest upstream master should help :)

cmd/triggerrun/cmd/root.go
```diff
@@ -88,24 +93,57 @@ func trigger(triggerFile, httpPath string) error {
...
 
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	if err != nil {
+		return fmt.Errorf("Failed to build config from the flags: %w", err)
+	}
+
 	logger, _ := zap.NewProduction()
 	sugerLogger := logger.Sugar()
 	eventID := template.UID()
+	sinkClients, err := sink.ConfigureClients(config)
```
dibyom, line 104: A lot of the logic here is the same as https://github.com/tektoncd/triggers/blob/master/cmd/eventlistenersink/main.go#L52

It would be nice to have a NewSink function in the sink package that returns a Sink struct. The NewSink function should take in the config *rest.Config as an input and create a Sink with the right clients/logger.

cmd/triggerrun/cmd/root.go
```diff
@@ -50,27 +54,28 @@ var (
...
+	err := trigger(triggerFile, httpPath, action)
 	if err != nil {
 		return fmt.Errorf("fail to call trigger: %v", err)
 	}
 
 	return nil
 }
 
-func trigger(triggerFile, httpPath string) error {
+func trigger(triggerFile, httpPath, action string) error {
```
dibyom, line 78: the trigger function now contains quite a bit of logic...we should add some tests for it.

cmd/triggerrun/cmd/root.go
```diff
@@ -88,24 +93,57 @@ func trigger(triggerFile, httpPath string) error {
...
+		if action == "show" {
+			for _, resource := range resources {
+				s, err := yaml.Marshal(resource)
+				if err != nil {
+					return fmt.Errorf("fail to print out the resource: %w", err)
+				}
+				fmt.Println("-----------------------------------")
+				fmt.Printf("%s", s)
+			}
+		} else if action == "create" {
```
savitaashture, line 142: Will there be any other actions?
If yes can we make use of switch case instead of if/else WDYT?

dorismeixing: It only has two actions, which is "Create" and "Show" for now, but it will have more actions in in the future.

cmd/triggerrun/cmd/root_test.go
```diff
@@ -38,6 +39,40 @@ import (
...
+	}
+
+	gotValue := buf.Bytes()
+	wantValue := []uint8(nil)
+
+	if diff := cmp.Diff(wantValue, gotValue); diff != "" {
+		t.Errorf("Trigger mismatch -want +got: %s", diff)
+	}
+
+	buf = new(bytes.Buffer)
```
dibyom, line 59: this looks like duplicate of above...

cmd/triggerrun/cmd/root_test.go
```diff
@@ -38,6 +39,40 @@ import (
 	rtesting "knative.dev/pkg/reconciler/testing"
 )
 
+func TestTrigger(t *testing.T) {
```
dibyom, line 42: Since we are testing an error case, we should name this TestTrigger_Error

cmd/triggerrun/cmd/root_test.go
```diff
@@ -38,6 +39,40 @@ import (
...
+	}
+
+	gotValue := buf.Bytes()
+	wantValue := []uint8(nil)
+	if diff := cmp.Diff(wantValue, gotValue); diff != "" {
+		t.Errorf("Trigger mismatch -want +got: %s", diff)
+	}
+
+	//error case for create feature
+	err = trigger("../testdata/trigger.yaml", "../testdata/http.txt", "create", buf)
```
dibyom, line 60: Its nice that we are trying to add test cases for both show and create. But in this case, the error that we are expecting happens before we every get to the switch statement for show or create:
```
	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
```

So, in effect, we are just testing the same case twice. So I'd recommend just keeping one of the cases instead of two.

Issue #714, opened 2020-08-17
Increasing the el number of replicas requires deleting the deployment


afrittoli:
# Expected Behavior

When the number of replicas is updated in the event listener, the number of replicas should be updated in the deployment.

# Actual Behavior

Nothing happens unless the deployment is deleted.
```
$ k -o json get el/tekton-events | jq .spec.replicas
3
$ k -o json get deployment/el-tekton-events | jq .spec.replicas
1
```

After deleting the deployment:
```
k delete deployment/el-tekton-events
```

I get the right behaviour:
```
k -o json get deployment/el-tekton-events | jq .spec.replicas
3

k get pods | grep el-tekton-events
el-tekton-events-cb5b88c5b-cxp7r                                  1/1     Running     0          45s
el-tekton-events-cb5b88c5b-vplkj                                  1/1     Running     0          45s
el-tekton-events-cb5b88c5b-wqbg7                                  1/1     Running     0          45s
```

# Steps to Reproduce the Problem

1. Create an el
2. Change the number of replicas (I used `kubectl replace`)
3. Check that nothing happens

# Additional Info

```
Triggers v0.7.0

Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-21T14:51:23Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.11-gke.5", GitCommit:"baccd25d44f1a0d06ad3190eb508784efbb990a5", GitTreeState:"clean", BuildDate:"2020-06-25T22:55:26Z", GoVersion:"go1.13.9b4", Compiler:"gc", Platform:"linux/amd64"}
```

PR #715, opened 2020-08-17
Fix update deployment when there is a change in replicas

savitaashture:
# Changes
The changes fix update deployment operation when there is a change in replicas number.
With current code whenever user updates `replicas` as part of eventlistener update success but no change in behavior.

Fix: https://github.com/tektoncd/triggers/issues/714

# Submitter Checklist

review them:

# Release Notes

```
None
```

### DIFF
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -166,6 +166,8 @@ check out the guide on [exposing EventListeners](./exposing-eventlisteners.md).
 The `replicas` field is optional. By default, the number of replicas of EventListener is 1.
 If you want to deploy more than one pod, you can specify the number to this field. 
 
+**Note:** If user sets `replicas` field while creating eventlistener yaml then it won't respects replicas values edited by user manually or through any other mechanism (ex: HPA).
+
 ### PodTemplate
 
 The `podTemplate` field is optional. A PodTemplate is specifications for 
--- a/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
+++ b/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
@@ -348,7 +348,7 @@ func (c *Reconciler) reconcileDeployment(el *v1alpha1.EventListener) error {
 		el.Status.SetDeploymentConditions(existingDeployment.Status.Conditions)
 		// Determine if reconciliation has to occur
 		updated := reconcileObjectMeta(&existingDeployment.ObjectMeta, deployment.ObjectMeta)
-		if existingDeployment.Spec.Replicas == nil || *existingDeployment.Spec.Replicas == 0 {
+		if existingDeployment.Spec.Replicas != deployment.Spec.Replicas {
 			existingDeployment.Spec.Replicas = replicas
 			updated = true
 		}
--- a/pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
+++ b/pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
@@ -546,7 +546,7 @@ func TestReconcile(t *testing.T) {
 		endResources: test.Resources{
 			Namespaces:     []*corev1.Namespace{namespaceResource},
 			EventListeners: []*v1alpha1.EventListener{elWithStatus},
-			Deployments:    []*appsv1.Deployment{deploymentWithUpdatedReplicas},
+			Deployments:    []*appsv1.Deployment{elDeployment},
 			Services:       []*corev1.Service{elService},
 			ConfigMaps:     []*corev1.ConfigMap{loggingConfigMap},
 		},


pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -347,7 +347,7 @@ func (c *Reconciler) reconcileDeployment(el *v1alpha1.EventListener) error {
 		el.Status.SetDeploymentConditions(existingDeployment.Status.Conditions)
 		// Determine if reconciliation has to occur
 		updated := reconcileObjectMeta(&existingDeployment.ObjectMeta, deployment.ObjectMeta)
-		if existingDeployment.Spec.Replicas == nil || *existingDeployment.Spec.Replicas == 0 {
+		if existingDeployment.Spec.Replicas != deployment.Spec.Replicas {
```
khrm, line 350:
```suggestion
		if existingDeployment.Spec.Replicas == nil ||  existingDeployment.Spec.Replicas != deployment.Spec.Replicas {
```
I feel it should be above. It's quite possible for replica in deployment to be nil, then reconciler will panic. What if user manually changed deployment's replica to zero or empty by supplying yaml file?

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
```diff
@@ -451,20 +454,20 @@ func Test_reconcileDeployment(t *testing.T) {
 			name: "deployment-replica-update",
 			startResources: test.Resources{
 				Namespaces:     []*corev1.Namespace{namespaceResource},
-				EventListeners: []*v1alpha1.EventListener{eventListener1},
-				Deployments:    []*appsv1.Deployment{deployment3},
+				EventListeners: []*v1alpha1.EventListener{eventListener3},
```
dibyom, line 457: One issue here is that before we had `spec.replicas` in the EL itself, users had to manually update the replica on the deployment itself. And we had code and this test to verify that we do not overwrite any changes to the number of replicas on an EL. So, this is a change in our behavior.  I am thinking we can either:

1. Honor the old replica changes i.e. if someone changes the existing replica numbers in the deployment itself and `spec.Replicas` is not set on the EL, we should not set it back. This keeps the behavior backwards compatible

2. Or we deprecate the existing behavior (not sure how many people depend on this behavior) and only honor EL's `spec.replicas` -- we just call this out in the release notes. 


One thing I'm curious about (and don't know) is how this will work with Horizontal Pod Autoscalers. Do they change the deployment's spec.Replicas? In that case, if an HPA scales the deployment we do not want to inadvertently scale it back down.

khrm: Can we have the behavior where if el.spec.Replicas is set, then we track changes only in el.spec.Replicas and don't care what user set in deployment.
If el.spec.Replicas isn't set, then we don't change deployment replica and let k8s controller take care of it?
HPA uses deployment's number as you surmised. So design 2 will break them.

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
savitaashture: @khrm replica in deployment will never be nil because we always set replicas to `1` https://github.com/tektoncd/triggers/blob/master/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go#L262 so considering that we can assume like `existingDeployment.Spec.Replicas` will never be nil

khrm: What if user set it nil in Deployment YAML bypassing EL?

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
savitaashture: Ya this is something need to verify with HPA 
Thanks for pointing it out i will try and update here

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
savitaashture: setting `replicas` to **nil**  by `kubectl edit deployment` is not allowed which gives error like below
```
* : Invalid value: "The edited file failed validation": ValidationError(Deployment.spec.replicas): invalid type for io.k8s.api.apps.v1.DeploymentSpec.replicas: got "string", expected "integer"
```
and if we set `replicas` to 0 by doing `kubectl edit deployment`  it will rewrite to `1`

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
savitaashture: > One thing I'm curious about (and don't know) is how this will work with Horizontal Pod Autoscalers. Do they change the deployment's spec.Replicas? In that case, if an HPA scales the deployment we do not want to inadvertently scale it back down.

@dibyom Yes HPA uses deployment so with this PR change if HPA increases replicas on deployment whenever condition met But that will be again overridden by `el.Spec.Replicas` if set as part of eventlistener yaml

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
khrm: You can set nil by not giving or removing `replicas` during `kubectl edit deployment`.
But I check that defaults would initialize the pointer and put it to 0 if its nil. https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/pkg/apis/apps/v1beta1/defaults.go#L85

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
savitaashture: How about below opinion on supporting replicas

1. Lets keep `spec.replicas` which can be set by the user and we will add a note in README that if user sets `spec.replicas` as part of eventlistener yaml then it will override replicas values if user manually set by `kubectl edit deployment` or by setting HPA  
Which means if `spec.replicas`  set by event listener yaml then it won't respects replicas values edited by user manually or through any other mechanism (ex: HPA)

2. If user Don't set `spec.replicas` in eventlistener yaml then behavior will be same like earlier how it use to work before adding replicas field to eventlistener yaml

Few points from WG discussion

1. Don't allow `spec.replicas` as part of eventlistener yaml instead hardcode replicas to some number instead of `1` 
    -> With this approach if we hardcode `replicas` to say `3` then if HPA configured with `--min=2 --max=5` then again replicas number varies 

2. Some prior art / similar behavior for k8s addons - https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/addon-manager
    -> This is bit tricky to implement 

/cc @dibyom @khrm @wlynch

dibyom: Thanks for summarizing the discussion. It sounds good to me!

docs/eventlisteners.md
```diff
@@ -166,6 +166,8 @@ check out the guide on [exposing EventListeners](./exposing-eventlisteners.md).
 The `replicas` field is optional. By default, the number of replicas of EventListener is 1.
 If you want to deploy more than one pod, you can specify the number to this field. 
 
+**Note:** If user sets `replicas` field while creating eventlistener yaml then it won't respects replicas values edited by user manually or through any other mechanism (ex: HPA).
+
```
savitaashture, line 170: @dibyom Added a note as per the discussion here https://github.com/tektoncd/triggers/pull/715#discussion_r472392787

dibyom: Thanks!

Issue #716, opened 2020-08-17
Handle validation for replicas


savitaashture:
Now User can specify replicas information using eventlistener as part of [spec](https://github.com/tektoncd/triggers/blob/master/pkg/apis/triggers/v1alpha1/event_listener_types.go#L58) 

Following are few observation:
1. No validation for the replicas and user can set any negative value (ex: -1)
2. No replicas specified as part of spec But still EL will have below data after the creation
```
spec:
  podTemplate: {}
  replicas: 0
  serviceAccountName: tekton-triggers-github-sa
``` 
and EventListener pod will be UP and Running and which may lead confusion to the user .

# Expected Behavior

1. User shouldn't be allowed to provide `replicas` as negative number if provided it should give below error
```
Error from server (BadRequest): error when creating "examples/github/github-eventlistener-interceptor.yaml": admission webhook "validation.webhook.triggers.tekton.dev" denied the request: validation failed: invalid value: -1: spec.replicas
```
2. Created EL shouldn't show replicas field if user don't specify as part of Eventlistener yaml
i mean this one
```
spec:
  podTemplate: {}
  replicas: 0
  serviceAccountName: tekton-triggers-github-sa
``` 
So EL after creation should contain `replicas` only if it specified as part of eventlistener yaml

PR #717, opened 2020-08-17
Add validation for replicas

savitaashture:
# Changes

1. Add changes to validate invalid replicas specified by user as part of eventlistener spec.
2. Modified in order not to show replicas field after EL creation when replicas are not specified as part of EL spec.

Fix: https://github.com/tektoncd/triggers/issues/716

Signed-off-by: Savita Ashture sashture@redhat.com

# Submitter Checklist

review them:

# Release Notes

EventListener spec validation ensures proper validation when user specifies `spec.replicas` values as `negative or zero`
ex:

```
cat <<EOF | kubectl create -f -
kind: EventListener
metadata:
  name: bitbucket-listener
spec:
  serviceAccountName: tekton-triggers-bitbucket-sa
  replicas: 0
  triggers:
    - name: bitbucket-triggers
      bindings:
        - ref: bitbucket-binding
      template:
        name: bitbucket-template
EOF
```
Gives response like below
```
Error from server (BadRequest): error when creating "STDIN": admission webhook "validation.webhook.triggers.tekton.dev" denied the request: validation failed: invalid value: 0: spec.replicas
```

### DIFF
--- a/pkg/apis/triggers/v1alpha1/event_listener_types.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_types.go
@@ -55,7 +55,7 @@ type EventListenerSpec struct {
 	ServiceAccountName string                 `json:"serviceAccountName"`
 	Triggers           []EventListenerTrigger `json:"triggers"`
 	ServiceType        corev1.ServiceType     `json:"serviceType,omitempty"`
-	Replicas           int32                  `json:"replicas"`
+	Replicas           *int32                 `json:"replicas,omitempty"`
 	PodTemplate        PodTemplate            `json:"podTemplate,omitempty"`
 }
 
--- a/pkg/apis/triggers/v1alpha1/event_listener_validation.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_validation.go
@@ -28,10 +28,15 @@ import (
 
 // Validate EventListener.
 func (e *EventListener) Validate(ctx context.Context) *apis.FieldError {
-	return e.Spec.validate(ctx, e)
+	return e.Spec.validate(ctx)
 }
 
-func (s *EventListenerSpec) validate(ctx context.Context, el *EventListener) *apis.FieldError {
+func (s *EventListenerSpec) validate(ctx context.Context) *apis.FieldError {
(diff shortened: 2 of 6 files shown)


pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -55,7 +55,7 @@ type EventListenerSpec struct {
 	ServiceAccountName string                 `json:"serviceAccountName"`
 	Triggers           []EventListenerTrigger `json:"triggers"`
 	ServiceType        corev1.ServiceType     `json:"serviceType,omitempty"`
-	Replicas           int32                  `json:"replicas"`
```
dibyom, line 58: should this be a *int32 since it is an optional field?

khrm: Optional field here will use empty value as 0. In case of pointer, it's nil. So if we want 0 as a valid value for Replicas, we can use *int32, otherwise, int32 is also good enough.

savitaashture: even if user provide replicas as `0` it will be overridden to `1` so shall we keep `int32` or shall we move it to `*int32`?

khrm: After thinking about this, I realize that it would better to have it as `*int32` because it would cover some of the scenarios that we discussed in the Sync meeting. If want EL to not care about number of replicas if it isn't provided. Just up the deployment with a certain number which isn't reflected in the EL.Spec.

savitaashture: Modified `int32` to `*int32` this way we can avoid accepting `0` value as part of `spec.replicas`.
Handled same thing in validation

dibyom:
/test pull-tekton-triggers-integration-tests


Thanks @savitaashture 
This is a feature right? So, we should add something to the  release notes section.

savitaashture:
>This is a feature right? So, we should add something to the release notes section.

@dibyom Yes this is a feature
Updated release notes section

Thank you

Issue #687, opened 2020-07-24
Event listener does not accept events after some time


afrittoli:
# Expected Behavior
I can always deliver events to an event listener

# Actual Behavior
We enabled cloud events from taskruns in our dogfooding clusters. This means that every single task run and condition that runs generates three events (start, running, passed/failed), and these events are all sent to a single event listener in the dogfooding cluster.

A few of these events are selected to trigger pipelines:
- when start/pass/fail events are received from a taskrun part of a CI job, the github check status is updated
- when fail events are received from a taskrun part of a config map CD service, a notification is sent to slack

It happened twice in the past three weeks, since this service was enabled, that the event listener stopped accepting events.
Looking at the k8s events on one of the taskruns it shows:
```
Events:
  Type     Reason               Age                 From     Message
  ----     ------               ----                ----     -------
  Normal   Started              28m                 TaskRun
  Normal   Pending              28m                 TaskRun  Pending
  Normal   Pending              28m                 TaskRun  pod status "Initialized":"False"; message: "containers with incomplete status: [place-scripts place-tools]"
  Normal   Pending              28m                 TaskRun  pod status "Initialized":"False"; message: "containers with incomplete status: [place-tools]"
  Normal   Pending              28m                 TaskRun  pod status "Ready":"False"; message: "containers with unready status: [step-git-source-source-sh9dr step-kubeconfig-dlvwz step-ln-dir-targetcluster-vv5r7 step-fetch-current-config step-deploy]"
  Normal   Running              28m                 TaskRun  Not all Steps in the Task have finished executing
  Normal   Succeeded            28m                 TaskRun  All Steps have completed executing
  Warning  Cloud Event Failure  28m (x13 over 28m)  TaskRun  Post http://el-tekton-events.default:8080: dial tcp 10.121.3.142:8080: connect: cannot assign requested address (10x)
```

The pod and service associated to the event listener do not expose any obvious issue - they appear to be healthy.
The event listener itself seems to be fine. 
Deleting and recreating the event listeners solves the issue.

Unfortunately I do not have more data to share about this, I had to find a quick fix to restore CI, but perhaps the triggers team has an idea about what may be going wrong or what kind of data we may collect to catch this issue the next time it happens.

My guess is that a relatively high volume of events triggers the issue, but it's not clear in which component the issue happens exactly. 

# Steps to Reproduce the Problem

1.
2.
3.

# Additional Info
Triggers v0.5.0
Kube v0.16

[...]

PR #747, opened 2020-09-08
Add Timeout for EventListener Server

khrm:
Added Timeout to avoid running out of file descriptor and leaking connections.

# Changes

# Submitter Checklist

review them:

# Release Notes
```release-note
Add Read, Write, and Idle timeout for connections to the EventListener.
```

### DIFF
--- a/cmd/eventlistenersink/main.go
+++ b/cmd/eventlistenersink/main.go
@@ -123,9 +123,12 @@ func main() {
 	})
 
 	srv := &http.Server{
-		Addr:        fmt.Sprintf(":%s", sinkArgs.Port),
-		IdleTimeout: 120 * time.Second,
-		Handler:     mux,
+		Addr:         fmt.Sprintf(":%s", sinkArgs.Port),
+		ReadTimeout:  sinkArgs.ELReadTimeOut * time.Second,
+		WriteTimeout: sinkArgs.ELWriteTimeOut * time.Second,
+		IdleTimeout:  sinkArgs.ELIdleTimeOut * time.Second,
+		Handler: http.TimeoutHandler(mux,
+			sinkArgs.ELTimeOutHandler*time.Second, "EventListener Timeout!\n"),
 	}
 
 	if err := srv.ListenAndServe(); err != nil {
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -57,6 +57,10 @@ spec:
           "-stderrthreshold", "INFO",
           "-el-image", "ko://github.com/tektoncd/triggers/cmd/eventlistenersink",
           "-el-port", "8080",
+          "-el-readtimeout", "5",
+          "-el-writetimeout", "40",
+          "-el-idletimeout", "120",
+          "-el-timeouthandler", "30",
           "-period-seconds", "10",
           "-failure-threshold", "1"
         ]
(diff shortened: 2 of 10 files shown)


khrm:
This pr might resolve https://github.com/tektoncd/triggers/issues/687

cmd/eventlistenersink/main.go
```diff
@@ -101,5 +102,15 @@ func main() {
 		w.WriteHeader(200)
 		fmt.Fprint(w, "ok")
 	})
-	logger.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", sinkArgs.Port), nil))
+
+	srv := &http.Server{
+		Addr:         fmt.Sprintf(":%s", sinkArgs.Port),
+		ReadTimeout:  sinkArgs.ELReadTimeOut * time.Second,
```
dibyom, line 108: ReadTimeout includes reading both headers + body. Maybe we should use `ReadHeaderTimeout` instead?

cmd/eventlistenersink/main.go
```diff
@@ -101,5 +102,15 @@ func main() {
...
+		WriteTimeout: sinkArgs.ELWriteTimeOut * time.Second,
+		IdleTimeout:  sinkArgs.ELIdleTimeOut * time.Second,
```
dibyom, line 110: This is the one we should definitely set.

cmd/eventlistenersink/main.go
khrm: Or may we set both? What would be the sane values for all these?

dibyom:
@khrm How do you feel about adding just a IdleTimeout for the v0.8.1 release. We can then add the additional timeouts + config options to change it for v0.9?

khrm:
@dibyom Yea, that's good.

cmd/eventlistenersink/main.go
```diff
@@ -107,9 +107,12 @@ func main() {
...
 	srv := &http.Server{
-		Addr:        fmt.Sprintf(":%s", sinkArgs.Port),
-		IdleTimeout: 120 * time.Second,
-		Handler:     mux,
+		Addr:         fmt.Sprintf(":%s", sinkArgs.Port),
+		ReadTimeout:  sinkArgs.ELReadTimeOut * time.Second,
+		WriteTimeout: sinkArgs.ELWriteTimeOut * time.Second,
+		IdleTimeout:  sinkArgs.ELIdleTimeOut * time.Second,
+		Handler: http.TimeoutHandler(mux,
+			sinkArgs.ELTimeOutHandler*time.Second, "EventListener Timeout!\n"),
```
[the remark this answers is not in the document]
khrm: No, unless we write our own custom thing.

dibyom:
Summarizing discussion from last week's WG:
- There are a lot of controls, which give users a lot of power...but we'd need docs to explain how these timeouts work with each other and when to use what.

- As an alternative, maybe we can have a single timeout for the entire request processing time that the user can control (e.g. 30s) ? We'd use a context and cancel the context after this deadline. That might make this easier to explain/use.

- At any rate, we should keep the IdleTimeout at least so that we are not leaking connections.

cmd/eventlistenersink/main.go
dibyom: 5 seconds sounds reasonable to me.

khrm:
@dibyom @wlynch  Can we merge this pr? I have added relevant doc in eventlistener.md

dibyom:
/approve
/lgtm

(We should still consider adding a context based timeout to cancel trigger processing that takes more than x seconds)

Issue #656, opened 2020-07-02
EventListener return a 2xx status on failure


khrm:
When I test a trigger and it fails, the event listener pod logs messages with a level: error. But the status code returned by the event listener is in the 2xx range which is confusing. Github seems to think that the web hooks worked.

# Expected Behavior
In case of an error, the event listener should return a status code in the 500's

# Actual Behavior
In Github, I can see that the webhook was executed after a push but despite receiving messages like the following with a "level" of "error", the event listener returns a status code in the 200's which seems like a success until you dig into the logs.

{"level":"error","logger":"eventlistener","caller":"sink/sink.go:164","msg":"failed to ApplyEventValuesToParams: failed to marshal event: failed to unmarshal request body: invalid character 'p' looking for beginning of value","knative.dev/controller":"eventlistener","/triggers-eventid":"75t66","/trigger":"","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.processTrigger\n\t/opt/app-root/src/go/src/github.com/tektoncd/triggers/pkg/sink/sink.go:164\ngithub.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent.func1\n\t/opt/app-root/src/go/src/github.com/tektoncd/triggers/pkg/sink/sink.go:97"}
{"level":"error","logger":"eventlistener","caller":"sink/sink.go:164","msg":"failed to ApplyEventValuesToParams: failed to replace JSONPath value for param pusher-name:

{string $(body.name) []}
: name is not found","knative.dev/controller":"eventlistener","/triggers-eventid":"wz9kj","/trigger":"","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.processTrigger\n\t/opt/app-root/src/go/src/github.com/tektoncd/triggers/pkg/sink/sink.go:164\ngithub.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent.func1\n\t/opt/app-root/src/go/src/github.com/tektoncd/triggers/pkg/sink/sink.go:97"}

# Steps to Reproduce the Problem

1. Created a sample task. I replaced the $(input.params.appName) by a string to output hello world and removed the parameter appname
2. Created a pipeline with this task alone
3. Added a trigger using the "Add Trigger" option
4. Used the provided url in my Github repository that I use to test the trigger (https://github.com/joellord/triggertest)
5. Use format form-urlencoded in the Github web hook configuration to force an error
Test the hook

# Additional Info
[...]

PR #758, opened 2020-09-15
Add Error Msg and BadRequest for Parsing error

khrm:
This fixes EventListener Error Response during Parsing to return
400 when marshaling error occurs or JsonPath error is encountered.
Fixes: #656

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -64,7 +64,13 @@ type Response struct {
 	Namespace string `json:"namespace,omitempty"`
 	// EventID is a uniqueID that gets assigned to each incoming request
 	EventID string `json:"eventID,omitempty"`
+	// Error message contains any error that occur during processing event
+	Error string `json:"error,omitempty"`
 }
+
+const (
+	marshalErr = "failure to marshal event body correctly"
+)
 
 // HandleEvent processes an incoming HTTP event for the event listener.
 func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
@@ -100,6 +106,9 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 					result <- http.StatusForbidden
 					return
 				}
+				if template.IsParseErr(err) {
+					result <- http.StatusBadRequest
+				}
 				result <- http.StatusAccepted
 				return
 			}
@@ -110,6 +119,7 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 	//The eventlistener waits until all the trigger executions (up-to the creation of the resources) and
 	//only when at least one of the execution completed successfully, it returns response code 201(Created) otherwise it returns 202 (Accepted).
 	code := http.StatusAccepted
+	errMsg := ""
 	for i := 0; i < len(el.Spec.Triggers); i++ {
 		thiscode := <-result
 		// current take - if someone is doing unauthorized stuff, we abort immediately;
@@ -117,10 +127,16 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 		// below around accepted vs. created
 		if thiscode == http.StatusUnauthorized || thiscode == http.StatusForbidden {
 			code = thiscode
+			errMsg = ""
 			break
+		}
+		if thiscode == http.StatusBadRequest && code != http.StatusCreated {
+			code = http.StatusBadRequest
+			errMsg = marshalErr
 		}
 		if thiscode < code {
 			code = thiscode
+			errMsg = ""
 		}
 	}
 
@@ -130,6 +146,7 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 		EventListener: r.EventListenerName,
 		Namespace:     r.EventListenerNamespace,
 		EventID:       eventID,
+		Error:         errMsg,
 	}
 	if err := json.NewEncoder(response).Encode(body); err != nil {
 		eventLog.Errorf("failed to write back sink response: %w", err)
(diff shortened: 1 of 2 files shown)


pkg/template/event.go
```diff
@@ -97,11 +102,22 @@ func applyEventValuesToParams(params []triggersv1.Param, body []byte, header htt
...
 			pValue = strings.ReplaceAll(pValue, originals[i], val)
 		}
 		params[idx].Value = pValue
 	}
 	return params, nil
 }
+
+// IsParseErr determines if err is an error which indicates that parsing event failed
+func IsParseErr(err error) bool {
+	if strings.Contains(err.Error(), jsonPathErr) {
```
wlynch, line 116: I don't think we should determine error type based on the strings (e.g. could I embed extra data in interceptor responses that would cause this to pass?)

If we go with this style of error handling, we should likely look into error wrapping instead. https://blog.golang.org/go1.13-errors

wlynch: This would also allow us to not tie BadRequest statuses to marshal errors, since we could infer it from the error type instead.

pkg/sink/sink.go
```diff
@@ -64,8 +64,14 @@ type Response struct {
 	Namespace string `json:"namespace,omitempty"`
 	// EventID is a uniqueID that gets assigned to each incoming request
 	EventID string `json:"eventID,omitempty"`
+	// Error message contains any error that occur during processing event
+	Error string `json:"error,omitempty"`
 }
 
+const (
+	marshalErr = "failure to marhal event body correctly"
```
wlynch, line 72:
```suggestion
	marshalErr = "failure to marshal event body correctly"
```

pkg/sink/sink.go
```diff
@@ -119,6 +129,10 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 			code = thiscode
 			break
 		}
+		if thiscode == http.StatusBadRequest && code != http.StatusCreated {
+			code = http.StatusBadRequest
+			errMsg = marshalErr
+		}
 		if thiscode < code {
 			code = thiscode
```
wlynch, line 137: IIUC, this doesn't actually solve the problem, since the next successful trigger that returns 200 will overwrite the code, but keep the error.

pkg/sink/sink.go
```diff
@@ -119,6 +129,10 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
 			break
 		}
+		if thiscode == http.StatusBadRequest && code != http.StatusCreated {
+			code = http.StatusBadRequest
+			errMsg = marshalErr
```
wlynch, line 134: This doesn't layer well with the unauthorized error- if the next trigger is unauthorized, you'll have an unauthorized error with a marshal error detail.

pkg/template/event.go
```diff
@@ -97,11 +102,22 @@ func applyEventValuesToParams(params []triggersv1.Param, body []byte, header htt
 		for i, expr := range expressions {
 			val, err := ParseJSONPath(event, expr)
 			if err != nil {
-				return nil, fmt.Errorf("failed to replace JSONPath value for param %s: %s: %w", p.Name, p.Value, err)
+				return nil, fmt.Errorf(jsonPathErr+"%s: %s: %w", p.Name, p.Value, err)
```
wlynch, line 105:
```suggestion
				return nil, fmt.Errorf("%s %s: %s: %w", jsonPathErr, p.Name, p.Value, err)
```

pkg/template/event.go
```diff
@@ -87,7 +92,7 @@ func newEvent(body []byte, headers http.Header) (*event, error) {
 func applyEventValuesToParams(params []triggersv1.Param, body []byte, header http.Header) ([]triggersv1.Param, error) {
 	event, err := newEvent(body, header)
 	if err != nil {
-		return nil, fmt.Errorf("failed to marshal event: %w", err)
+		return nil, fmt.Errorf(marshalErr+": %w", err)
```
wlynch, line 95:
```suggestion
		return nil, fmt.Errorf("%s: %w", marshalErr, err)
```

pkg/sink/sink.go
```diff
@@ -64,8 +64,14 @@ type Response struct {
 	Namespace string `json:"namespace,omitempty"`
 	// EventID is a uniqueID that gets assigned to each incoming request
 	EventID string `json:"eventID,omitempty"`
+	// Error message contains any error that occur during processing event
+	Error string `json:"error,omitempty"`
```
wlynch, line 68: I agree with @dibyom on providing a structured per-trigger response type, instead of trying to infer individual trigger statuses from the overall result.

The same way it's misleading that a 2xx response results in an error, I'm worried that bubbling up more per-trigger errors at the event level misleads people into thinking nothing happened, when several triggers could have succeeded. Unlike the existing 2xx behavior, this might have more severe consequences (e.g. spawning many more Runs than intended).

I think https://google.aip.dev/193#partial-errors would be a great model to follow / draw inspiration from here (particularly the use of long running operations and returning subresource errors in metadata).

pkg/sink/sink.go
khrm: I changed the code so that we don't get any error in this case.

pkg/sink/sink.go
khrm: I put errMsg = "" if unauthorized error comes. So we won't show that.

kylewolfe:
> Fixes: #656

It looks as if this PR is only addressing marshalling issues, however the verbiage in #656 would cover my use case of also returning a non 2xx on both invalid payloads, as well as any failure in creating the resource, whether it be from a missing parameter (4xx) or some other scheduling reasons (5xx).

Should I open a new issue to discuss, or should this PR not close the original issue?

khrm:
/close
Will raise another pr which validates json body.

Issue #568, opened 2020-05-06
TriggerTemplate default value is not applied if the param is missing from Binding


lqy1992:
### [Summary](https://github.com/tektoncd/triggers/issues/568#issuecomment-648838427):

> I've been hitting this issue as well - it seems that if an omitted field in the JSON payload is used in a TriggerBinding, we error out before even reaching the code that merges in default values from our TriggerTemplate. It works fine as long as the field is present, even if it's a zero value like an empty string.
> 
> The relevant bit of code I could find: https://github.com/tektoncd/triggers/blob/master/pkg/template/event.go#L36-L41
> Here we exit with a JSONPath error, thus never reaching MergeInDefaultParams.
> 
> One way to solve this, would be to attempt to merge in our default params even when encountering a JSONPath error (field is missing) and finally return the error if no default is provided and the binding value can't be resolved.

triggers-binding.yaml
```
apiVersion: triggers.tekton.dev/v1alpha1
kind: TriggerBinding
metadata:
  name: triggers-nodejs-app-binding
spec:
  params:
  **- name: replicaset
    value: $(header.replicaset)**
  - name: gitrevision
    value: $(body.head_commit.id)
  - name: gitrepositoryurl
    value: $(body.repository.url)
  - name: dockerTag
    value: $(body.dockertag.version)
  - name: imageurl
    value: $(body.dockerhub.url)
```
triggertemplate.yaml
```
apiVersion: triggers.tekton.dev/v1alpha1
kind: TriggerTemplate
metadata:
  name: pipeline-template-nodejs-app
spec:
  params:
  - name: gitrevision
    description: The git revision
    default: master
  - name: gitrepositoryurl
    description: The git repository url
  - name: dockerTag
    default: latest
  **- name: replicaset
    default: "2"**
  - name: imageurl
    description: url of docker hub
  resourcetemplates:
  - apiVersion: tekton.dev/v1beta1
    kind: PipelineRun
    metadata:
      generateName: nodejs-app-elrun
    spec:
      pipelineRef:
        name: nodejs-app-pipeline
      serviceAccountName: tekton-pipelines-controller
      params:
      - name: dockerTag
        value: $(params.dockerTag)
      - name: replicaset
        value: $(params.replicaset)
      resources:
      - name: mygit
        resourceSpec:
          type: git
          params:
          - name: revision
            value: $(params.gitrevision)
          - name: url
            value: $(params.gitrepositoryurl)
      - name: myimage
        resourceSpec:
[...]

PR #761, opened 2020-09-18
apply triggertemplate param default value if body/header based triggerbinding param missing from body/header

gabemontero:
# Changes

Fixes https://github.com/tektoncd/triggers/issues/568

# Submitter Checklist

review them:

- [/] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [n/a] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [/] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

# Release Notes

    ```release-note
    NONE
    ```

/assign @dibyom

### DIFF
--- a/pkg/template/event.go
+++ b/pkg/template/event.go
@@ -28,17 +28,17 @@ import (
 // ResolveParams takes given triggerbindings and produces the resulting
 // resource params.
 func ResolveParams(rt ResolvedTrigger, body []byte, header http.Header) ([]triggersv1.Param, error) {
-	out, err := applyEventValuesToParams(rt.BindingParams, body, header)
-	if err != nil {
-		return nil, fmt.Errorf("failed to ApplyEventValuesToParams: %w", err)
-	}
-
 	var ttParams []triggersv1.ParamSpec
 	if rt.TriggerTemplate != nil {
 		ttParams = rt.TriggerTemplate.Spec.Params
 	}
 
-	return mergeInDefaultParams(out, ttParams), nil
+	out, err := applyEventValuesToParams(rt.BindingParams, body, header, ttParams)
+	if err != nil {
+		return nil, fmt.Errorf("failed to ApplyEventValuesToParams: %w", err)
+	}
+
+	return out, nil
 }
 
 // ResolveResources resolves a templated resource by replacing params with their values.
@@ -79,24 +79,40 @@ func newEvent(body []byte, headers http.Header) (*event, error) {
 
 // applyEventValuesToParams returns a slice of Params with the JSONPath variables replaced
 // with values from the event body and headers.
-func applyEventValuesToParams(params []triggersv1.Param, body []byte, header http.Header) ([]triggersv1.Param, error) {
+func applyEventValuesToParams(params []triggersv1.Param, body []byte, header http.Header,
+	defaults []triggersv1.ParamSpec) ([]triggersv1.Param, error) {
 	event, err := newEvent(body, header)
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal event: %w", err)
 	}
 
-	for idx, p := range params {
+	allParamsMap := map[string]string{}
+	for _, paramSpec := range defaults {
+		if paramSpec.Default != nil {
+			allParamsMap[paramSpec.Name] = *paramSpec.Default
+		}
+	}
+
+	for _, p := range params {
 		pValue := p.Value
 		// Find all expressions wrapped in $() from the value
 		expressions, originals := findTektonExpressions(pValue)
 		for i, expr := range expressions {
 			val, err := parseJSONPath(event, expr)
+			if defaults != nil && err != nil {
+				// if the header or body was not supplied or was malformed, go with a default if it exists
+				v, ok := allParamsMap[p.Name]
+				if ok {
+					val = v
+					err = nil
+				}
+			}
 			if err != nil {
 				return nil, fmt.Errorf("failed to replace JSONPath value for param %s: %s: %w", p.Name, p.Value, err)
 			}
 			pValue = strings.ReplaceAll(pValue, originals[i], val)
 		}
-		params[idx].Value = pValue
+		allParamsMap[p.Name] = pValue
 	}
-	return params, nil
+	return convertParamMapToArray(allParamsMap), nil
 }
--- a/docs/triggerbindings.md
+++ b/docs/triggerbindings.md
@@ -74,6 +74,11 @@ To access JSON keys that contain `.` character, we need to escape the `.` e.g.
 $(body.tekton\.dev) -> "triggers"
 ```
 
+#### Defaulting
+
+If the HTTP headers and body contents from an event fail to resolve the JSONPath expressions supplied, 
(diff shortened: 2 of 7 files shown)


savitaashture (commented):
How about adding sample example to [examples](https://github.com/tektoncd/triggers/tree/master/examples) to show usage of this

pkg/template/event.go
```diff
@@ -33,16 +34,16 @@ func ResolveParams(rt ResolvedTrigger, body []byte, header http.Header) ([]trigg
...
 	if rt.TriggerTemplate != nil {
 		ttParams = rt.TriggerTemplate.Spec.Params
 	}
 
+	out, err = applyEventValuesToParams(out, body, header, ttParams)
+	if err != nil {
+		return nil, fmt.Errorf("failed to ApplyEventValuesToParams: %w", err)
+	}
+
 	return MergeInDefaultParams(out, ttParams), nil
```
dibyom, line 47: do we still need to call `MergeInDefaultParams` since we are now adding default values in `applyEventValuesToParams` itself?

pkg/template/event.go
```diff
@@ -96,6 +97,18 @@ func applyEventValuesToParams(params []triggersv1.Param, body []byte, header htt
 		expressions, originals := findTektonExpressions(pValue)
 		for i, expr := range expressions {
 			val, err := ParseJSONPath(event, expr)
+			if defaults != nil && err != nil {
+				// if the header or body was not supplied or was malformed, go with a default if it exists
+				for _, ps := range defaults {
+					if ps.Name == p.Name && ps.Default != nil {
+						fmt.Fprintf(os.Stdout, "using default %s since failed to replace JSONPath value for param %s: %s: error: %s\n",
```
dibyom, line 104: This is useful information, but I'm not sure we need to print it out. If we do, maybe we can pass through the logger that we for other logging instead of using fmt.Fprintf?

gabemontero:
> How about adding sample example to [examples](https://github.com/tektoncd/triggers/tree/master/examples) to show usage of this or any doc update which provide this information that there is a support of default values if not found from `binding`

Per https://github.com/tektoncd/triggers/issues/568#issuecomment-627410993 the existing example could be tweaked to leverage this, where the fact that the default is getting used is highlighted.  Would that be amenable vs. creating a separate example?

I did scan the docs and was unclear on where to update.  Suggestions on how to cross reference https://github.com/tektoncd/triggers/blob/master/docs/triggertemplates.md with https://github.com/tektoncd/triggers/blob/master/docs/triggerbindings.md to connect the dots @savitaashture @dibyom ?

pkg/template/event.go
gabemontero: Yeah I looked at that and found it a bit onerous to pass the logger through all the call levels to get it here.

But if this is not an tenable compromise, then sure.  I can go one way or the other.  I don't have a strong opinion.  Will it be feasible for a user to figure out that if the default is used vs. the body/header reference, there was an issue with how it was constructed?

You got a final call here @dibyom ?

pkg/template/event.go
gabemontero: Yes we do.  Unit tests fail without it:

```
using default defaultVal since failed to replace JSONPath value for param p2: $(body.p2): error: p2 is not found
--- FAIL: TestResolveParams (0.00s)
    --- FAIL: TestResolveParams/add_default_values_for_params_with_missing_values (0.00s)
        event_test.go:351: didn't get expected params -want + got:   []v1alpha1.Param{
              	{Name: "p1", Value: "val1"},
            - 	{Name: "p2", Value: "defaultVal"},
              }
FAIL
FAIL	github.com/tektoncd/triggers/pkg/template	0.062s
FAIL
```

pkg/template/event.go
dibyom: Yeah, I think its fine to remove the log statement here given that this is documented/expected behavior.

dibyom:
> Per #568 (comment) the existing example could be tweaked to leverage this, where the fact that the default is getting used is highlighted. Would that be amenable vs. creating a separate example?

Sure as long as we have an example that shows that we substitute the default value when one from the event is missing, I'm ok!

> I did scan the docs and was unclear on where to update. Suggestions on how to cross reference https://github.com/tektoncd/triggers/blob/master/docs/triggertemplates.md with https://github.com/tektoncd/triggers/blob/master/docs/triggerbindings.md to connect the dots

In the [parameters](https://github.com/tektoncd/triggers/blob/master/docs/triggertemplates.md#parameters) section of TriggerTemplates, we can add something about the `default` field and when the default value is applied.

Similarly, in the TriggerBindings docs, we can add a something about the behavior maybe, as sub-section inside the Event Variable Interpolation section -- if we fail to resolve the param value from the event, we'll look to see if the trigger template param has a default value. If it doesn't we throw an error.

pkg/template/event.go
dibyom: Hmmm, interesting. I think that makes sense. But makes me wonder if we can simplify this code somehow.

pkg/template/event.go
gabemontero: ok just pushed removal of log statement as part of rebasing this PR on top of recent changes

pkg/template/event.go
gabemontero: ok @dibyom I've pushed a separate commit that consolidates `applyEventValuesToParams` and `mergeInDefaultParams`

it is possible the changes pulled in with the rebase helped things (the method was minimally switched from public to package only)

otherwise, the biggest takeaway is that an explicit for loop on the `default` array has been replaced with a golang map lookup.... that is positive.

If you like this approach better I'll  squash the commits.

gabemontero:
> > Per #568 (comment) the existing example could be tweaked to leverage this, where the fact that the default is getting used is highlighted. Would that be amenable vs. creating a separate example?
> 
> Sure as long as we have an example that shows that we substitute the default value when one from the event is missing, I'm ok!
> 

OK pushed commit https://github.com/tektoncd/triggers/pull/761/commits/64735d2a727a68df93363d31e05e3a223eb42b2e to try and capture this

gabemontero:
> In the [parameters](https://github.com/tektoncd/triggers/blob/master/docs/triggertemplates.md#parameters) section of TriggerTemplates, we can add something about the `default` field and when the default value is applied.
> 
> Similarly, in the TriggerBindings docs, we can add a something about the behavior maybe, as sub-section inside the Event Variable Interpolation section -- if we fail to resolve the param value from the event, we'll look to see if the trigger template param has a default value. If it doesn't we throw an error.

doc updates following these recommendations in https://github.com/tektoncd/triggers/pull/761/commits/774fb5c62124ab6b72f104792e7df71681b4648b

tests have been passing previously, so I'm assuming this doc only change will not affect things

PTAL @dibyom @savitaashture 

thanks!!

dibyom:
Nice cleanup! Would you mind squashing up the commits?

gabemontero:
commits squashed @dibyom

PR #767, opened 2020-09-23
Bump pipeline version to v0.16.3

savitaashture:
# Changes

Bumped to pipeline version v0.16.3

# Submitter Checklist

review them:

# Release Notes

The manual changes are:

1. Call `Promote` in tests, otherwise our generated reconcilers won't do anything (they aren't the leader!)
2. Fill namespaces in a few tests

```release-note
The minimum Kubernetes supported version is now 1.16.
```


savitaashture:
@dibyom PR is ready to review

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -190,6 +189,21 @@ func (els *EventListenerStatus) SetCondition(newCond *apis.Condition) {
...
 
+func (els *EventListenerStatus) MarkTrue() {
+	eventListenerCondSet.Manage(els).SetCondition(apis.Condition{
+		Type:   apis.ConditionReady,
+		Status: corev1.ConditionTrue,
+	})
+}
+
+// GetConditionSet retrieves the condition set for this resource. Implements
+// the KRShaped interface.
```
dibyom, line 200: Question: Do we have to implement the KRShaped interface as part of this PR? I think this adds a new `status.Condition` of type `Ready` -- which is what we'd want but we'd also have to add logic for when `Ready` is true (e.g. when deployment/service are ready).

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
```diff
@@ -771,6 +779,9 @@ func TestReconcile(t *testing.T) {
 			// Setup with startResources
 			testAssets, cancel := getEventListenerTestAssets(t, tt.startResources)
 			defer cancel()
+			if la, ok := testAssets.Controller.Reconciler.(pkgreconciler.LeaderAware); ok {
```
dibyom, line 782: Can this be moved to the `getEventListenerTestAssets` function?

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
```diff
@@ -782,6 +793,8 @@ func TestReconcile(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
+			fmt.Println("GOT", actualEndResources)
```
dibyom, line 796: remove

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
```diff
@@ -858,6 +871,9 @@ func TestReconcile_Delete(t *testing.T) {
 			// Setup with startResources
 			testAssets, cancel := getEventListenerTestAssets(t, tt.startResources)
 			defer cancel()
+			if la, ok := testAssets.Controller.Reconciler.(pkgreconciler.LeaderAware); ok {
```
dibyom, line 874: same as above

pkg/apis/triggers/v1alpha1/event_listener_types.go
dibyom: Pipeline seems to disable it for the existing types at the moment: https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L41

savitaashture: yes a new condition of type `Ready` handled here https://github.com/tektoncd/triggers/pull/767/files#diff-ac6d17e220f6c870c0bc0304bc7c53b4R115-R117

savitaashture: If we want to do it same like pipeline we can disable it on issues

Let me know

dibyom: Yeah, so if we are adding Ready we should ensure that it is only true iff both deployment and service are also ready. I think this will be a useful follow up. But since that's a feature and adds new fields to the type, we can add it in a follow up?

dibyom:
Thanks @savitaashture 

Could we also update the Release Notes to mention the updated minimum version of k8s needed?

pkg/apis/triggers/v1alpha1/event_listener_types.go
savitaashture: Ya that make sense :+1:

savitaashture:
> Thanks @savitaashture
> 
> Could we also update the Release Notes to mention the updated minimum version of k8s needed?

Updated Release Notes

dibyom:
Not 100% sure, but we might also be running into https://github.com/tektoncd/pipeline/issues/3237 with this change. Might be worth calling out in the release notes unless we get to upgrade again to a newer knative/pkg that has a fix.

Issue #617, opened 2020-06-17
More concise embedded TriggerBindings


dibyom:
# Expected Behavior
A trigger binding consists of name and value pairs. So, when we embed it we should be able to use something like:

```
 bindings:
    - ref: my-cool-binding # this one is a ref to an existing binding

    - name: commit_id # embedded binding
      value: "$(body.head_commit_id)

    - name: message # embedded binding with static values
      value: "Hello from an embedded binding"
```

# Actual Behavior
Today, the above example looks like:
```
 bindings:
        - ref: my-cool-binding # this one is a ref to an existing binding
       
        -  name: some-name
           spec:
              params:
                - name: commit_id # embedded binding
                  value: "$(body.head_commit_id)
                - name: message # embedded binding with static values
                  value: "Hello from an embedded binding"
```

# Additional Info

From: https://github.com/tektoncd/triggers/pull/548/files#r417024864
/kind feature

dibyom:
WG update: I'll write up a TEP!

PR #769, opened 2020-09-25
Fix Binding builder to only use ref

dibyom:
# Changes

The testbuilder `EventListenerTriggerBinding` allows users to specify both
`name` and `ref` fields. With #617, specifying both `name` and `ref` is no
longer allowed.

We fixed the builder usage for all the happy path tests to use `name` as an
empty string but did not do it for tests that are verifying error paths (such
as validation errors). So, these tests will now all fail at the same validation
check making most of these tests useless (causing the drop in coverage in

To fix, I changed the builder to only accept `ref` so that one cannot create
invalid structs. Later, we can migrate our tests to not use builders at all and
have our error tests verify the actual error string.

# Submitter Checklist

review them:

# Release Notes


savitaashture (commented):
/lgtm

khrm (approved):
/lgtm

Issue #616, opened 2020-06-17
Support embedded TriggerTemplates


dibyom:
# Expected Behavior

We should  be able to embed a TriggerTemplate spec inside a eventlistener's trigger just like we embedd a [TriggerBinding](https://github.com/tektoncd/triggers/blob/master/examples/eventlisteners/eventlistener-embedded-bindings.yaml)

# Actual Behavior

Users have to create a separate TriggerTemplate object and then reference it

# Additional Info
1. Other examples are embedding [tasks](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/taskruns/task-volume-args.yaml#L13) and [pipelines](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-with-pipelinespec-and-taskspec.yaml) inside the TaskRun/PipelineRun objects.

/kind feature

dibyom:
WG Update: I'll write up a (short) TEP

PR #783, opened 2020-10-02
Add support for embedded trigger templates

dibyom:
# Changes

This PR implements TEP-0023 by adding an optional `spec` field inside a
Trigger's `template` field. Users can specify either the `name` field to refer
to an embedded template or they can embed the template inside the `spec`
field.

Part of #616

/kind feature

# Submitter Checklist

review them:

# Release Notes
```release-note
Users can now specify embed a TriggerTemplate spec inside a Trigger.
```


dibyom:
/hold

Waiting on the TEP PR: https://github.com/tektoncd/community/pull/224

docs/triggers.md
```diff
@@ -50,6 +51,29 @@ spec:
     name: pipeline-template
 ```
 
+### template
+The `template` field inside a `Trigger` can either point to an existing `TriggerTemplate` object (using `name`) or the template spec can be embedded inside the Trigger using the `spec` field:
+
+```yaml
+# Embedded Template Sped
```
savitaashture, line 58: nit
```suggestion
# Embedded Template Spec
```

savitaashture (commented):
PR looks good to me :+1: 

Small nit fix for comment word

/lgtm

docs/triggers.md
dibyom: ahh nice catch again :)

pkg/template/resource.go
```diff
@@ -52,11 +52,22 @@ func ResolveTrigger(trigger triggersv1.EventListenerTrigger, getTB getTriggerBin
...
 		return ResolvedTrigger{}, fmt.Errorf("failed to resolve bindings: %w", err)
 	}
-	tt, err := getTT(context.Background(), trigger.Template.Name, metav1.GetOptions{})
-	if err != nil {
-		return ResolvedTrigger{}, fmt.Errorf("error getting TriggerTemplate %s: %w", trigger.Template.Name, err)
+
+	var resolvedTT *triggersv1.TriggerTemplate
+	if trigger.Template.Spec != nil {
+		resolvedTT = &triggersv1.TriggerTemplate{
+			ObjectMeta: metav1.ObjectMeta{}, // Unsed. TODO: Just return Specs from here.
```
wlynch, line 59:
```suggestion
			ObjectMeta: metav1.ObjectMeta{}, // Unused. TODO: Just return Specs from here.
```

pkg/apis/triggers/v1alpha1/trigger_validation_test.go
```diff
@@ -165,6 +187,9 @@ func Test_TriggerValidate(t *testing.T) {
 }
 
 func TestTriggerValidate_error(t *testing.T) {
+	var simpleResourceTemplate = runtime.RawExtension{
+		Raw: []byte(`{"kind":"PipelineRun","apiVersion":"tekton.dev/v1alpha1","metadata":{},"spec":{},"status":{}}`),
+	}
```
wlynch, line 192: nit: This can be brittle if we need to modify this . Let's break start with a PipelineRun type and define a func that will return the RawExtension?

e.g.

```go
func rawExtension(t *testing.T, a interface{}) runtime.RawExtension {
  t.Helper()
  o, err := json.Marshal(a)
  if err != nil {
    t.Fatal(err)
  }
  return runtime.RawExtension{Raw: b}
}
```

Then we can probably also get rid of this top level var and just inline it below (since this is only used once)

pkg/template/resource.go
```diff
@@ -52,11 +52,22 @@ func ResolveTrigger(trigger triggersv1.EventListenerTrigger, getTB getTriggerBin
...
+			Spec:       *trigger.Template.Spec,
+		}
+	} else {
+		ttName := trigger.Template.Name
+		resolvedTT, err = getTT(context.Background(), ttName, metav1.GetOptions{})
```
wlynch, line 64: We should plumb the context through, otherwise we won't respect timeouts/cancels applied to the context.

(I realize this is carry over from before - your choice if you want to handle it in this PR or if you want to differ to a different PR).

pkg/template/resource_test.go
```diff
@@ -469,6 +470,33 @@ func Test_ResolveTrigger(t *testing.T) {
...
 		},
 	},
+		{
+			name: "embedded trigger template",
+			trigger: triggersv1.EventListenerTrigger{
+				Template: &triggersv1.EventListenerTemplate{
+					Spec: &triggersv1.TriggerTemplateSpec{
+						ResourceTemplates: []triggersv1.TriggerResourceTemplate{{
+							RawExtension: runtime.RawExtension{
+								Raw: []byte(`{"kind":"PipelineRun","apiVersion":"tekton.dev/v1alpha1","metadata":{},"spec":{},"status":{}}`),
```
wlynch, line 480: Similar RawExtension func could be useful here as well.

wlynch (commented):
Mostly LGTM! Just a few small suggestions.

pkg/template/resource.go
dibyom: done

pkg/apis/triggers/v1alpha1/trigger_validation_test.go
dibyom: neat, yeah the string approach is pretty brittle

dibyom: I changed it here and opened #796 for cleaning up other usages.

pkg/template/resource.go
dibyom: yeah, this was recently added due to an upgrade to k8s v1.18 which changed the interfaces for clients to accept a context. 
I think I'd want to do this in a different PR (probably using an lister cache -- so that we are not calling the API server for each event).

pkg/template/resource_test.go
dibyom: done

wlynch (approved):
/lgtm

PR #787, opened 2020-10-05
Use template.ref instead of template.name

dibyom:
# Changes

Previously, we'd use the `name` field in `triggers` to refer to
TriggerTemplates. This commit deprecates the `name` field in favor of the `ref`
field. This makes the usage of `ref` consistent across `bindings` and
`template`.

`name` is still supported and will be removed in a future release.

Part of #616

I've split the changes to all examples/doc to use the `ref` field to a separate commit.

# Submitter Checklist

review them:

# Release Notes

```release-note
Use template.ref to refer to trigger templates in a Trigger instead of template.name. The name field is deprecated and will be removed in a future release.
```


dibyom:
/hold

Depends on https://github.com/tektoncd/triggers/pull/783

pkg/apis/triggers/v1alpha1/trigger_validation.go
```diff
@@ -55,13 +55,25 @@ func (t TriggerSpecTemplate) validate(ctx context.Context) (errs *apis.FieldErro
...
+
+	// Set Ref to Name
+	if t.Name != "" {
+		t.Ref = &t.Name
+	}
+
 	switch {
-	case t.Spec != nil && t.Name != "":
+	case t.Spec != nil && t.Ref != nil:
 		errs = errs.Also(apis.ErrMultipleOneOf("template.spec", "template.name"))
```
savitaashture, line 70: DO we need to change error message because we have `ref` here instead of `name` :thinking:

pkg/apis/triggers/v1alpha1/trigger_validation.go
```diff
@@ -55,13 +55,25 @@ func (t TriggerSpecTemplate) validate(ctx context.Context) (errs *apis.FieldErro
...
-	case t.Spec == nil && t.Name == "":
+	case t.Spec == nil && t.Ref == nil:
 		errs = errs.Also(apis.ErrMissingOneOf("template.spec", "template.name"))
```
savitaashture, line 72: Same here

pkg/apis/triggers/v1alpha1/trigger_validation.go
```diff
@@ -55,13 +55,25 @@ func (t TriggerSpecTemplate) validate(ctx context.Context) (errs *apis.FieldErro
...
 	case t.Spec != nil:
 		errs = errs.Also(t.Spec.validate(ctx))
+	case t.Ref == nil || *t.Ref == "":
+		errs = errs.Also(apis.ErrMissingField("template.name"))
```
savitaashture, line 76: here as well

pkg/apis/triggers/v1alpha1/trigger_defaults.go
```diff
@@ -58,3 +59,11 @@ func (t triggerSpecBindingArray) defaultBindings() {
 		}
 	}
 }
+
+func templateNameToRef(template *TriggerSpecTemplate) {
+	name := template.Name
+	if name != "" && template.Ref == nil {
+		template.Ref = &name
+		template.Name = ""
```
savitaashture, line 67: just a thought is it required to set "" to `name` can't we ignore it ?

pkg/apis/triggers/v1alpha1/trigger_validation.go
dibyom: Indeed we do 😅

pkg/apis/triggers/v1alpha1/trigger_defaults.go
dibyom: defaults get applied before validation, if we do not unset Name (to ""), we'll hit this line in trigger_validation.go:

```
      if t.Name != "" && t.Ref != nil {
		errs = errs.Also(apis.ErrMultipleOneOf("template.name", "template.ref"))
	}
```

savitaashture: ah i see got it :+1:

savitaashture (commented):
/lgtm

Thank you :+1:

wlynch (approved):
/lgtm

PR #804, opened 2020-10-15
Revert "dibyo's point about only EL sink/interceptors reading secrets"

### LINKED ISSUE #803: Webhook:  TLS handshake error from 10.128.0.8:47288: secret "triggers-webhook-certs" not found
# Expected Behavior

Users in fresh Triggers installation can create new triggers resources 

# Actual Behavior

Users in new triggers installs cannot create new triggers resources.
Triggers webhook logs if filled with:
``` sh
TLS handshake error from 10.128.0.8:47288: secret "triggers-webhook-certs" not found
```

# Additional Info

1. I think this only happens for new clusters where Triggers was not installed before and not if one is upgrading.

/kind bug

dibyom:
# Changes

This reverts commit 8eb2067bcc90cbfa3eca783497ab83b07da16d47.

Turns out, the webhook does need access to create secrets since it uses it
to create `tekton-webhook-certs` if it does not exist.

Fixes #803

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

/kind bug

### DIFF
--- a/config/200-role.yaml
+++ b/config/200-role.yaml
@@ -12,10 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# NOTE:  when multi-tenant EventListener progresses, moving this Role
+# to a ClusterRole is not the advisable path.  Additional Roles that
+# adds access to Secrets to the Namespaces managed by the multi-tenant
+# EventListener is what should be done.  While not as simple, it avoids
+# giving access to K8S system level, cluster admin privileged level Secrets
+
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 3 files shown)


khrm (approved):
/lgtm

gabemontero:
Finally got a second to look at this for a sec ... am I correct in that the knative webhook stuff is what is actually creating the secret @dibyom?

If so, perhaps because of that there is some precedence that we cannot overcome at this time (i.e. knative has mutiple places where it creates secrets dynamically), but a possible counter argument to adding secret creation back would be that we should be creating this secret as part of the resources applied from the `config` directory.

thoughts?

Issue #809, opened 2020-10-22
Eventlistener container crashes with eventlistener created with SA having no permissions to use eventlisteners.


sravankumar777:
# Expected Behavior

- Eventlistener shoud log error message & not crash, when event listener is created with a serviceaccount not having permissions for eventlisteners objects.
- Eventlistener could hang-in failure state provided with error message to check for right permissions of serviceaccount.

# Actual Behavior
Event Listener container crashes, when an eventlistener created with invalid(serviceaccount) & fired a bitbucket event to check for working condition with eventlistener service url. 

# Steps to Reproduce the Problem

1. Create EventListener with an invalid serviceaccount(not have permissions for serviceaccount to use eventlistener objects).
2. Using event listener service url, fire an event from any git repository(in this case, i have used bitbucket repository).
3. Check for eventlistener logs in that particular namespace.

# Additional Info
EventListener Logs:
```
{"level":"fatal","logger":"eventlistener","caller":"sink/sink.go:73","msg":"Error getting EventListener tekton-cnb-el in Namespace test-namespace: eventlisteners.triggers.tekton.dev \"tekton-cnb-el\" is forbidden: User \"system:serviceaccount:test-namespace:tenant-pipeline-default\" cannot get resource \"eventlisteners\" in API group \"triggers.tekton.dev\" in the namespace \"test-namespace\"","knative.dev/controller":"eventlistener","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:73\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2012\nnet/http.(*ServeMux).ServeHTTP\n\tnet/http/server.go:2387\nnet/http.serverHandler.ServeHTTP\n\tnet/http/server.go:2807\nnet/http.(*conn).serve\n\tnet/http/server.go:1895"}
```

```
$ kubectl -n tekton-pipelines get deployments tekton-triggers-controller -ojsonpath='{.metadata.labels.triggers\.tekton\.dev\/release}'
v0.8.0
```

PR #810, opened 2020-10-22
Don't log Fatal when handling events in the eventlistener

bigkevmcd:
# Changes

Don't log with Fatal when we can't fetch resources.

For this case, Fatalf causes the process to exit, and so the container cycles.

Fixes: #809 

I had a look through the other cases where we're using Fatalf on the logger, and they all look like non-user error cases 

It's not clear whether or not these should also be fixed for example this pattern is used twice:

https://github.com/tektoncd/triggers/blob/master/pkg/client/injection/reconciler/triggers/v1alpha1/eventlistener/controller.go#L55

This issue also doesn't look like a user could trigger it:

https://github.com/tektoncd/triggers/blob/master/pkg/client/injection/reconciler/triggers/v1alpha1/eventlistener/reconciler.go#L134

# Submitter Checklist

review them:

# Release Notes

```release-note
Previously, if the EventListener object could not be loaded by the eventlistener process when handling an event, the error logging caused the container to cycle, it should now be logged nomally.
```


pkg/sink/sink_test.go
```diff
@@ -87,7 +90,7 @@ func getSinkAssets(t *testing.T, resources test.Resources, elName string, auth A
 	clients := test.SeedResources(t, ctx, resources)
 	test.AddTektonResources(clients.Kube)
 
-	logger, _ := zap.NewProduction()
+	logger := zaptest.NewLogger(t)
```
dibyom, line 93: nice!

dibyom:
Thanks for the fix and adding the test! @bigkevmcd 

The other two places where we use `log.Fatalf` should be fine. They are generated code, and they basically fail to start the reconciler if the config is incorrect. I think that is ok.

Issue #807, opened 2020-10-20
separate SAs for controller/webhook deployment to allow for different permission sets


gabemontero:
# Expected Behavior

The controller and webhook deployments for triggers have differing permission needs, and should have only the permissions they need assigned to them

# Actual Behavior

Since the controller and webhook deployments utilized the same serviceaccount, the controller has the ability to create secrets, even though only the webhook deployment (which is the  wrapper around the knative based admission webhooks used by tekton) needs to be able to create a secret to house the cert needed for https based access to the related services

See https://github.com/tektoncd/triggers/pull/804 for more background

# Solution

If they used separate serviceaccounts, different roles and rolebindings could be applied for each

@dibyom per recent discussions

PR #818, opened 2020-10-30
separate SAs for controller/webhook deployment to allow for different permission sets

gabemontero:
Fixes https://github.com/tektoncd/triggers/issues/807

# Changes

See https://github.com/tektoncd/triggers/issues/807#issue-725940276

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

/assign @dibyom 

fyi - I have not tried this on a clean installed cluster yet ... will unhold after I do so

/hold

### DIFF
--- a/config/200-role.yaml
+++ b/config/200-role.yaml
@@ -31,6 +31,22 @@ rules:
     resources: ["podsecuritypolicies"]
     resourceNames: ["tekton-triggers"]
     verbs: ["use"]
+
+---
+
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: tekton-triggers-admin-webhook
+  namespace: tekton-pipelines
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-triggers
+rules:
+  - apiGroups: ["policy"]
+    resources: ["podsecuritypolicies"]
+    resourceNames: ["tekton-triggers"]
+    verbs: ["use"]
   - apiGroups: [""]
     resources: ["secrets"]
     verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
--- a/config/200-serviceaccount.yaml
+++ b/config/200-serviceaccount.yaml
@@ -20,3 +20,14 @@ metadata:
   labels:
     app.kubernetes.io/instance: default
     app.kubernetes.io/part-of: tekton-triggers
+
+---
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: tekton-triggers-webhook
+  namespace: tekton-pipelines
+  labels:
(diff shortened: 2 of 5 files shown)


gabemontero:
/hold cancel

ok @dibyom I've confirmed these changes in a new cluster that did not previously have pipelines/triggers

interestingly again, my prior commit with this PR did not work as is ... I had to make additional changes

as we talked about last time, there definitely does seem to be a clean up issue, most likely I would think around *Role and *RoleBindings, since removing permissions does not seem to get reflected in the CI runs, which were clean on my first push

config/200-role.yaml
```diff
@@ -26,11 +26,27 @@ metadata:
...
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-triggers
 rules:
   - apiGroups: ["policy"]
     resources: ["podsecuritypolicies"]
     resourceNames: ["tekton-triggers"]
     verbs: ["use"]
   - apiGroups: [""]
-    resources: ["secrets"]
```
gabemontero, line 35: wait - what? .... sigh

/hold

sorry for the churn @dibyom .... that is not what I was I thought I tested ... let me refresh everything and try again ... have to circle to something else though, so it will be tomorrow before I retry

gabemontero:
/hold cancel

ok @dibyom .... tested on a clean cluster with the right, non-overlapping set of role/rolebinding and clusterrole/clusterrolebinding

the high level jist:
- clusterrole / clusterrolebinding to allow for reading configmaps, services, and events applied to both controller/webhook SAs to facilitate reconciler/controllers' watches
- only the webhook can read secrets to facilitate the knative auto cert stuff
- the trigger controller comes up fine without secrets, which lines up with my searches finding no access to secrets there ... only in the sink/EL, which has a different set of RBAC

I'm still curious about 
a) how apparently RBAC seems to survive across e2e/pr runs
b) perhaps I am paranoid, but I half wonder if you should trying installing pipelines/triggers on a "clean cluster" in case the "flavor" of k8s you use has some unknown difference with mine

dibyom:
> a) how apparently RBAC seems to survive across e2e/pr runs

Hmm, my best guess is that we clean up the namespace but maybe not cluster resources (like cluster roles?). Though I was under the impression that we used fresh clusters for each e2e test....would have to look into it

> b) perhaps I am paranoid, but I half wonder if you should trying installing pipelines/triggers on a "clean cluster" in case the "flavor" of k8s you use has some unknown difference with mine

Sure, will try this out on my setup.

dibyom:
Tried it out on GKE. Works fine.

/lgtm
/approve

gabemontero:
> Tried it out on GKE. Works fine.
> 
> /lgtm
> /approve

thanks @dibyom

Issue #650, opened 2020-07-01
Support secure connection between service to EventListener pod


savitaashture:
# Expected Behavior
Supporting secure connection between EventListener service to pod.

# Actual Behavior
Currently we can achieve secure connection till ingress/service but cannot achieve secure connection from service to eventlistener pod. 

# Additional Info
Trigger should support end to end secure connection within and outside the cluster which helps a lot in security requirements.

/kind feature

verokarhu:
Wouldn't this be solved by deploying a service mesh?

savitaashture:
> Wouldn't this be solved by deploying a service mesh?

Hey @verokarhu sorry for the late reply

Yes that is one option but with that we will have dependency on thirdparty installation

And as part of this we are planning to support https connection to eventlistener pod as an inbuilt feature

csantanapr:
This is great to see that is being worked on. 👍 

I know about Istio but is a too heavy requirement for a simple use case where OpenShift Ingress Route comes with reencrypt and passthrough out of the box.

I'm using tekton via a downstream of OpenShift Pipelines, and the enterprise client wants a solution that goes along the same way OpenShift deals with this using Route reencrypt and passthrough

csantanapr:
Yay 🎉 haply to see implemented

PR #819, opened 2020-10-31
Allow secure connection to eventlistener pod

savitaashture:
# Changes

Issue: https://github.com/tektoncd/triggers/issues/650

As part of this PR Triggers eventlistener now support end to end secure connection to eventlistener pod.
For more info about the feature refer [TEP-0027](https://github.com/tektoncd/community/blob/master/teps/0027-https-connection-to-triggers-eventlistener.md) .

# Submitter Checklist

review them:

# Release Notes

`HTTPS` connection to eventlistener can be configured by tweaking eventlistener configuration
```
   kind: EventListener
   metadata:
     name: github-listener-interceptor
   spec:
     ...
     resources:
       kubernetesResource:
         spec:
           template:
             spec:
               serviceAccountName: tekton-triggers-github-sa
               containers:
               - env:
                 - name: TLS_CERT
                   valueFrom:
                     secretKeyRef:
                       name: tls-key-secret
                       key: tls.crt
                 - name: TLS_KEY
                   valueFrom:
                     secretKeyRef:
                       name: tls-key-secret
                       key: tls.key
``` 
**Note:**
`TLS_CERT`, `TLS_KEY` are **RESERVED** env

### DIFF
--- a/cmd/eventlistenersink/main.go
+++ b/cmd/eventlistenersink/main.go
@@ -136,7 +136,13 @@ func main() {
 			sinkArgs.ELTimeOutHandler*time.Second, "EventListener Timeout!\n"),
 	}
 
-	if err := srv.ListenAndServe(); err != nil {
-		logger.Fatalf("failed to start eventlistener sink: %v", err)
+	if sinkArgs.Cert == "" && sinkArgs.Key == "" {
+		if err := srv.ListenAndServe(); err != nil {
+			logger.Fatalf("failed to start eventlistener sink: %v", err)
+		}
+	} else {
+		if err := srv.ListenAndServeTLS(sinkArgs.Cert, sinkArgs.Key); err != nil {
+			logger.Fatalf("failed to start eventlistener sink: %v", err)
+		}
 	}
 }
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -42,6 +42,8 @@ using [Event Interceptors](#Interceptors).
     - [Multiple EventListeners (One EventListener Per Namespace)](#multiple-eventlisteners-one-eventlistener-per-namespace)
     - [Multiple EventListeners (Multiple EventListeners per Namespace)](#multiple-eventlisteners-multiple-eventlisteners-per-namespace)
     - [ServiceAccount per EventListenerTrigger](#serviceaccount-per-eventlistenertrigger)
+  - [EventListener Secure Connection](#eventlistener-secure-connection)
+    - [Prerequisites](#prerequisites)
 
 ## Syntax
 
@@ -277,8 +279,11 @@ Right now the allowed values as part of `podSpec` are
 ServiceAccountName
 NodeSelector
 Tolerations
+Volumes
 Containers
 - Resources
+- VolumeMounts
+- Env
 ```
 
 ### Logging
@@ -821,3 +826,12 @@ Except as otherwise noted, the content of this page is licensed under the
 [Creative Commons Attribution 4.0 License](https://creativecommons.org/licenses/by/4.0/),
 and code samples are licensed under the
 [Apache 2.0 License](https://www.apache.org/licenses/LICENSE-2.0).
+
+## EventListener Secure Connection
+
+Triggers now support both `HTTP` and `HTTPS` connection by adding few configuration to eventlistener.
+
+To setup TLS connection add two set of reserved environment variables `TLS_CERT` and `TLS_KEY` using `secretKeyRef` env type 
+where we need to specify the `secret` which contains `cert` and `key` files. See the full [example]((../examples/eventlistener-tls-connection/README.md)) for more details.
+
+Refer [TEP-0027](https://github.com/tektoncd/community/blob/master/teps/0027-https-connection-to-triggers-eventlistener.md) for more information on design and user stories.
--- a/examples/eventlistener-tls-connection/README.md
+++ b/examples/eventlistener-tls-connection/README.md
@@ -1,0 +1,78 @@
+## EventListener Secure Connection
+
+Triggers now support both `HTTP` and `HTTPS` connection by adding some configurations to eventlistener.
+
+### Prerequisites
+* Certificates with Key and Cert
+* Secret which includes those certificates
+
+### Try it out locally:
+
+#### Creating Prerequisites
+
+* #### Certificates with Key and Cert.
+
+##### 1. Steps to generate root key, cert
+1. Create Root Key
+   ```text
+   openssl genrsa -des3 -out rootCA.key 4096
+   ```
+2. Create and self sign the Root Certificate
+   ```text
+   openssl req -x509 -new -nodes -key rootCA.key -sha256 -days 1024 -out rootCA.crt
+   ```
+##### 2. Steps to generate certificate (for each server)
+1. Create the certificate key
+   ```text
+   openssl genrsa -out tls.key 2048
+   ```
+2. Create the signing (csr)
+
+* The CSR is where you specify the details for the certificate you want to generate. 
+This request will be processed by the owner of the root key to generate the certificate.
+
+* **Important:** While creating the csr it is important to specify the `Common Name` providing the IP address or domain name for the service, otherwise the certificate cannot be verified.
+   ```text
+   openssl req -new -key tls.key -out tls.csr
+   ```
+3. Generate the certificate using the tls csr and key along with the CA Root key
+   ```text
+   openssl x509 -req -in tls.csr -CA rootCA.crt -CAkey rootCA.key -CAcreateserial -out tls.crt -days 500 -sha256
+   ```
+##### 3. Follow same steps from 2 to generate certificates for client also
+
+* #### Create secret which includes those certificates
+   ```text
+   kubectl create secret generic tls-secret-key --from-file=tls.crt --from-file=tls.key
+   ```
+
+#### Configuring EventListener for TLS connection
+1. To create the TLS connection for EventListener and all related resources, run:
+
+   ```bash
+   kubectl apply -f examples/eventlistener-tls-connection/
+   ```
+
+1. Test by sending the sample payload.
+
+   ```bash
+   curl -v \
+   -H 'X-GitHub-Event: pull_request' \
+   -H 'X-Hub-Signature: sha1=ba0cdc263b3492a74b601d240c27efe81c4720cb' \
+   -H 'Content-Type: application/json' \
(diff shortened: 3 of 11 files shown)


cmd/eventlistenersink/main.go
```diff
@@ -112,7 +112,13 @@ func main() {
 		Handler:     mux,
 	}
 
-	if err := srv.ListenAndServe(); err != nil {
-		logger.Fatalf("faiiled to start eventlistener sink: %v", err)
+	if sinkArgs.Cert == "" && sinkArgs.Key == "" {
+		if err := srv.ListenAndServe(); err != nil {
+			logger.Fatalf("faiiled to start eventlistener sink: %v", err)
```
dibyom, line 117:
```suggestion
			logger.Fatalf("failed to start eventlistener sink: %v", err)
```

docs/eventlisteners.md
```diff
@@ -945,3 +950,55 @@ Except as otherwise noted, the content of this page is licensed under the
...
+* Certficates.
+* Secret which includes those certificates.
+
+There are two env var reserved for `HTTPS` connection to read certificates
+1. **SSL_CERT_FILE**: specify the location of mounted cert file.
+2. **SSL_KEY_FILE**: specify the location of mounted key file.
+
+As mentioned in the prerequisites users are asked to create their own certificates, secrets and mount as a volume in the 
+`resources` section of eventlistener as shown below.
+
```
dibyom, line 967: We definitely don't have to do this in this PR but it would be really nice to have an example/guide that runs through the whole flow i.e. creating certs, adding them to the secret, modifying the EL etc.

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
```diff
@@ -181,6 +181,14 @@ func Test_EventListenerValidate(t *testing.T) {
 										Effect:   "NoSchedule",
 									}},
 									NodeSelector: map[string]string{"beta.kubernetes.io/os": "linux"},
+									Volumes: []corev1.Volume{{
```
dibyom, line 184: could we move this to its own test case "Valid EventListener with TLS cert volumes" or something like that? Makes it easier to understand why this was added in the future.

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -261,120 +261,9 @@ func (r *Reconciler) reconcileDeployment(ctx context.Context, logger *zap.Sugare
...
-		Env: []corev1.EnvVar{{
-			Name: "SYSTEM_NAMESPACE",
-			ValueFrom: &corev1.EnvVarSource{
-				FieldRef: &corev1.ObjectFieldSelector{
-					FieldPath: "metadata.namespace",
-				},
-			},
-		}},
-	}
+	deployment := getDeployment(el)
```
dibyom, line 264: nice refactoring!

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
```diff
@@ -246,6 +246,121 @@ func makeDeployment(ops ...func(d *appsv1.Deployment)) *appsv1.Deployment {
 	return &d
 }
 
+// makeDeploymentForTLSConnection is a helper to build a Deployment that is created by an EventListener.
+// It generates a basic Deployment for the simplest EventListener and accepts functions for modification
+func makeDeploymentForTLSConnection(ops ...func(d *appsv1.Deployment)) *appsv1.Deployment {
```
dibyom, line 251: Instead of adding this function, is it possible to use makeDeployment() and then pass it a function that contains the modifications for the TLS certs? If we want to call it multiple times, we can extract that function and call it say, withTLSVolumes and then the call can look like `makeDeployment(withTLSVolumes)`. What do you think?

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
```diff
@@ -285,32 +400,40 @@ func logConfig(ns string) *corev1.ConfigMap {
...
+var withStatus = func(port ...int) bldr.EventListenerOp {
+	var p int
+	if len(port) == 0 {
+		p = *ElPort
+	} else {
+		p = port[0]
+	}
+	return bldr.EventListenerStatus(
+		bldr.EventListenerConfig(generatedResourceName),
+		bldr.EventListenerAddress(listenerHostname(generatedResourceName, namespace, p)),
```
dibyom, line 412: very minor: instead of modifying `withStatus` to take in an parameter, we can create a `withTLSPort` function that sets the EventListenerAddress to the right port. We'd call it using `makeEL(withStatus, withTLSPort, func...)`

pkg/sink/initialization.go
```diff
@@ -41,6 +41,10 @@ var (
 		"The namespace of the EventListener resource for this sink.")
 	portFlag = flag.String("port", "",
 		"The port for the EventListener sink to listen on.")
+	tlsCertFlag = flag.String("tls-cert", "",
+		"Filename for the TLS certificate.")
```
dibyom, line 45: Is this the filename or path to the file?

dibyom (commented):
Overall looks good. Just some minor comments. Also, how feasible would it be to add an e2e test for this?

(Also noticed some questions on https://github.com/tektoncd/community/pull/244 that we should resolve before merging this)

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
savitaashture: Actually we can make use of `makeDeployment()` but we need to pass lot of things to that 
ex: 
* ContainerPort
* Schema, Port to LivenessProbe and ReadinessProbe
* Modify port in the Args
* Modification in VolumeMount, Env and Volume

Basically i need to pass entire `Container` info so that is why created new deployment function.

Shall i go ahead with passing `container` info so that we will have single function `makeDeployment()`

dibyom: right, so what I was suggesting was:

```go

func WithTLSConfig(d *appsv1.Deployment) {
   d.Container.Port = ...
  // All of the TLS specific modifications to the container
}
```

and then we call:

```
makeDeployment(withTLSConfig)
```

Is this what you were thinking off too?

pkg/sink/initialization.go
savitaashture: earlier it was path to the file.
But after design change its now filename.

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
savitaashture: Modified :+1:

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
savitaashture: Yes i agree with you
Added new testcase for env related as we don't have volume support now

docs/eventlisteners.md
savitaashture: That's really good idea
Modified docs and added README to examples.

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
savitaashture: Yes 
Modified 
Thank you :+1:

dibyom:
/lgtm

Code looks fine. Pending design discussion on using `valueFrom/secretRef` instead of `value`: in https://github.com/tektoncd/community/pull/244

pkg/apis/triggers/v1alpha1/event_listener_validation.go
```diff
@@ -59,11 +67,82 @@ func validateKubernetesObject(orig *KubernetesResource) (errs *apis.FieldError)
...
+	// Allowed fields
+	out.SecretKeyRef = in.SecretKeyRef
+
+	// Disallowed fields
+	out.ConfigMapKeyRef = nil
+	out.FieldRef = nil
+	out.ResourceFieldRef = nil
+
+	return out
+}
```
savitaashture, line 127: I own to refactor changes related to masking fields in a follow up PR by creating separate file so that `_validation.go` will have changes related to validation

docs/eventlisteners.md
```diff
@@ -954,3 +959,11 @@ Except as otherwise noted, the content of this page is licensed under the
 [Creative Commons Attribution 4.0 License](https://creativecommons.org/licenses/by/4.0/),
 and code samples are licensed under the
 [Apache 2.0 License](https://www.apache.org/licenses/LICENSE-2.0).
+
+## EventListener Secure Connection
+
+Triggers now support both `HTTP` and `HTTPS` connection by adding few configurations to eventlistener.
```
dibyom, line 965: Could we add a little bit more detail here? e.g. :

even a like like -- To setup TLS, add set two environment variables `TLS_CERT`, and `TLS_KEY` with values set to a Secret. See the full example(LINK) for details`

dibyom (commented):
One small nit about docs otherwise LGTM!

/approve

docs/eventlisteners.md
savitaashture: Updated

examples/eventlistener-tls-connection/README.md
```diff
@@ -0,0 +1,102 @@
...
+   ```
+##### 3. Follow same steps from 2 to generate certificates for client also
+
+* #### Create secret which includes those certificates
+   ```text
+   kubectl create secret generic tls-secret-key --from-file=tls.crt --from-file=tls.key
+   ```
+
+#### Configuring EventListener for TLS connection
+```yaml
```
khrm, line 50: Nit: Why not put this in yaml file like other examples? It make it faster to test out examples. Otherwise, lgtm.

savitaashture: We can put but the secret will be given by user which contains cert and key files so even added sample example in README so that user can refer

khrm: We can probably assume `tls-key-secret` as secretKeyRef.

savitaashture: if we assume then we should create that before hand and for that we also need to create cert n keys

khrm: It's OK to not create key beforehand because those command will do it anyway.

khrm (approved):
/lgtm

khrm (approved):

PR #821, opened 2020-11-02
Use Listers to fetch data in the Sink

yaoxiaoqi:
# Changes

This PR uses Listers to fetch data from EventListeners, Triggers, TriggerBindings, TriggerTemplates, and
ClusterTriggerBindings, rather than directly making calls to the API server. Currently, multiple informers need to be
specified when creating a Sink object.
All Triggers in examples/ work under this change.

# Submitter Checklist

review them:

# Release Notes

```release-note
Use Listers to fetch data in the Sink
```

### DIFF
--- a/cmd/triggerrun/cmd/root.go
+++ b/cmd/triggerrun/cmd/root.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -37,6 +38,7 @@ import (
 	"github.com/tektoncd/triggers/pkg/sink"
 	"github.com/tektoncd/triggers/pkg/template"
 	"go.uber.org/zap"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/serializer/streaming"
 
 	"k8s.io/client-go/dynamic"
@@ -210,9 +212,15 @@ func processTriggerSpec(kubeClient kubernetes.Interface, client triggersclientse
 	}
 
 	rt, err := template.ResolveTrigger(el,
-		client.TriggersV1alpha1().TriggerBindings(tri.Namespace).Get,
-		client.TriggersV1alpha1().ClusterTriggerBindings().Get,
-		client.TriggersV1alpha1().TriggerTemplates(tri.Namespace).Get)
+		func(name string) (*triggersv1.TriggerBinding, error) {
+			return client.TriggersV1alpha1().TriggerBindings(tri.Namespace).Get(context.Background(), name, metav1.GetOptions{})
+		},
+		func(name string) (*triggersv1.ClusterTriggerBinding, error) {
+			return client.TriggersV1alpha1().ClusterTriggerBindings().Get(context.Background(), name, metav1.GetOptions{})
+		},
+		func(name string) (*triggersv1.TriggerTemplate, error) {
+			return client.TriggersV1alpha1().TriggerTemplates(tri.Namespace).Get(context.Background(), name, metav1.GetOptions{})
+		})
 	if err != nil {
 		log.Error("Failed to resolve Trigger: ", err)
 		return nil, err
--- a/examples/bitbucket/role.yaml
+++ b/examples/bitbucket/role.yaml
@@ -12,8 +12,8 @@ metadata:
 rules:
   # Permissions for every EventListener deployment to function
   - apiGroups: ["triggers.tekton.dev"]
-    resources: ["eventlisteners", "triggerbindings", "triggertemplates"]
-    verbs: ["get"]
+    resources: ["eventlisteners", "triggerbindings", "triggertemplates", "triggers"]
+    verbs: ["get", "list", "watch"]
   - apiGroups: [""]
     # secrets are only needed for GitHub/GitLab interceptors, serviceaccounts only for per trigger authorization
     resources: ["configmaps", "secrets", "serviceaccounts"]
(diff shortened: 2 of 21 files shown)


dibyom:
This is great! Thank you!

For the build tests, it looks like the issue is:

```
cmd/triggerrun/cmd/root.go:215:63: Error return value of `(github.com/tektoncd/triggers/pkg/client/clientset/versioned/typed/triggers/v1alpha1.TriggerBindingInterface).List` is not checked (errcheck)
	client.TriggersV1alpha1().TriggerBindings(tri.Namespace).List(context.Background(), metav1.ListOptions{})
```

So, we should handle the error value instead of ignoring it.

examples/bitbucket/role.yaml
```diff
@@ -33,4 +33,38 @@ subjects:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
-  name: tekton-triggers-bitbucket-minimal
\ No newline at end of file
+  name: tekton-triggers-bitbucket-minimal
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
dibyom, line 39: Instead of creating a new ClusterRoleBinding and Role, can we modify the existing Role to add the "list", and "watch" verbs? 
https://github.com/tektoncd/triggers/blob/master/examples/bitbucket/role.yaml#L16

In general, we'd want the ServiceAccount for the EventListener to have as few permissions as it can. With a ClusterRole, it will have access to all resources(triggers, bindings, etc.) in the cluster while with a Role it will only have access to the triggers within the resources within the namespace where it was created (this should be sufficient for our use cases).

cmd/triggerrun/cmd/root.go
```diff
@@ -208,11 +212,11 @@ func processTriggerSpec(kubeClient kubernetes.Interface, client triggersclientse
 	if tri.Namespace == "" {
 		tri.Namespace = "default"
 	}
-
+	client.TriggersV1alpha1().TriggerBindings(tri.Namespace).List(context.Background(), metav1.ListOptions{})
```
dibyom, line 215: hmm, looks like we aren't really doing anything with this line 😛  remove?

yaoxiaoqi: It's interesting that if I don't make any call directly using triggerClient(like Line#215), the Listers can't get any existing TriggerBinding or TriggerTemplate from the namespace. I'm new to Kubernetes but I guess the client does some preparation when called. I use about one day to find out this, but I can't figure out why Lister doesn't work without this line.

yaoxiaoqi: Besides, it only happens here. The Listers in cmd/eventlistenersink/main.go are totally fine.

dibyom: that interesting, maybe some kind of initialization that's missing? But one thing I just realized is that for cmd/triggerrun, it's ok to just make the direct API calls instead of using the lister since its a CLI tool. It'll only make the call once when the CLI tool is executed so using a Lister cache isn't all that useful.
I think passing in the direct triggersClient instead of the lister should be fine for this usage.

examples/bitbucket/role.yaml
yaoxiaoqi: Yes, I tried. But got this error `Failed to list *v1alpha1.TriggerTemplate: triggertemplates.triggers.tekton.dev is forbidden: User "system:serviceaccount:default:tekton-triggers-github-sa" cannot list resource "triggertemplates" in API group "triggers.tekton.dev" at the cluster scope` for every kind of resource that is involved, e.g. Trigger, TriggerTemplate, and TriggerBinding. 
Looks like the lister is used at the cluster scope. Do you mean the Lister should not be used at the cluster scope? I will take a look.

cmd/triggerrun/cmd/root.go
yaoxiaoqi: That makes sense. I will go back to the old way.

examples/bitbucket/role.yaml
dibyom: sorry missed this comment earlier. I think the issue here is that we are using a cluster scoped informer. We are setting up the informers here:  https://github.com/tektoncd/triggers/pull/821/files#diff-255de153352c6a922d11a7a76e0655d6dee57016f105177e0fbb7774bc1ae8b6R118


Instead what we can do is manually setup the Informers and restrict them to a namespace using  https://github.com/tektoncd/triggers/blob/v0.9.1/pkg/client/informers/externalversions/factory.go#L92

Something like:
```
	factory := informerfactory.NewSharedInformerFactoryWithOptions(clientset, 30*time.Second, informerfactory.WithNamespace(namespace)) // we can pass in the EL namespace here

```

Once we have the factory, we'll also have to start it:

```
	go func(ctx context.Context) {
		factory.Start(ctx.Done())
		<-ctx.Done()
	}(ctx)
```

and then return the Informers as we do now: ` factory.Triggers().V1alpha1().EventListeners(),`

yaoxiaoqi: ClusterTriggerBinding is a cluster scoped informer anyway. Should we exclude ClusterTriggerBinding when creating the Sink object? or it will emit the error without creating a new ClusterRoleBinding and ClusterRole.

dibyom: ahh for ClusterTriggerBinding, we'll need a cluster scoped informer. For the rest, we can get away with a namespace scoped one.

yaoxiaoqi: Thanks for explanation! This is what I'm doing currently. I was not sure it was proper or not before.

test/eventlistener_scale_test.go
```diff
@@ -138,4 +138,35 @@ func createServiceAccount(t *testing.T, c *clients, namespace, name string) {
 		t.Fatalf("Error creating RoleBinding: %s", err)
 	}
 
+	_, err = c.KubeClient.RbacV1().ClusterRoles().Create(context.Background(),
+		&rbacv1.ClusterRole{
```
dibyom, line 142: Since we are creating cluster-scoped resources, we should also delete them in a cleanup function. (By default we delete the namespace which deletes all the namespaced resources but not the cluster ones).

(e.g. we have  a cleanup https://github.com/tektoncd/triggers/blob/master/test/eventlistener_test.go#L608 that is then called in the test: https://github.com/tektoncd/triggers/blob/master/test/eventlistener_test.go#L127)

dibyom (commented):
Looking good. Just a couple of minor comments.
/lgtm

test/eventlistener_scale_test.go
yaoxiaoqi: Got it. Besides, I'm curious about https://github.com/tektoncd/triggers/blob/master/test/eventlistener_test.go#L608. I would like to know why we need to delete eventlistener after the whole namespace is already deleted. It seems that the eventlistener lives along with the namespace.

dibyom: hmm, that's a good question 🤔 

We probably should just delete the EL first and then the namespace? Or just the namespace.

dibyom: So, if TEST_KEEP_NAMESPACES env var is set, we will not delete the namespace but will delete the EL and the servcie . I wonder if this is intentional behavior.

yaoxiaoqi: Thanks for the explanation. It makes sense. Should I also delete the eventlistener in test/eventlistener_scale_test.go after tear down the namespace? I can open another PR.

dibyom: Actually, maybe we can just remove the lines that delete the EventListener and instead rely on the cleanup that already deletes the namespace to clean it up?

Issue #777, opened 2020-09-29
Backslash quote in json will crash the trigger parser


chmouel:
On a git repo which is plugged to a triggers, i have made a standard git revert of a commit to submit to a PR.

The title of git revert would look like this : 

```
    Revert "Detect tkn version"
```

This will shows up  like this in the webhook json payload : 

```
    "title": "Revert \"Detect tkn version\"",
```

but this title will crash trigger's parser : 

```
{"level":"error","logger":"eventlistener","caller":"sink/sink.go:173","msg":"couldn't unmarshal json: invalid character 'D' after object key:value pair","knative.dev/controller":"eventlistener","/triggers-eventid":"ff46j","/trigger":"github-listene
r","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.processTrigger\n\t/opt/app-root/src/go/src/github.com/tektoncd/triggers/pkg/sink/sink.go:173\ngithub.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent.func1\n\t/opt/app-root/src/go/src/githu
b.com/tektoncd/triggers/pkg/sink/sink.go:94"}

```

Changing the title and removing the quote made triggers accepting the json payload and go on with its life...

/label bug

dibyom:
@chmouel what version of Triggers are you running?

khrm:
@dibyom
He is running version 0.6.1.
I can recreate this on v0.8.1.
/assign

khrm:
Error in v0.8.1
```
{"level":"error","logger":"eventlistener","caller":"sink/sink.go:266","msg":"problem creating obj: &errors.errorString{s:\"couldn't unmarshal json: invalid character 'R' after object key:value pair\"}","knative.dev/controller":"eventlistener","/triggers-eventid":"9dvkl","/trigger":"foo-trig","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.CreateResources\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:266\ngithub.com/tektoncd/triggers/pkg/sink.Sink.processTrigger\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:183\ngithub.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent.func1\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:94"}
[...]

PR #823, opened 2020-11-05
Drop escaping of strings in the JSON.

bigkevmcd:
# Changes

Add support for the old escaping mechanism through an annotation on the
TriggerTemplate.

This removes the old replacement of `"` with `\"` in parameters, which was yielding
invalid JSON when the strings were already quoted.

Fixes: #777

# Submitter Checklist

review them:

# Release Notes

```release-note
Change the escaping of parameters into TriggerTemplates.

Previously, parameters were escaped as they were being replaced into a TriggerTemplate, by replacing double-quotes " with an escaped version \", this functionality has been removed, as it was breaking quoted strings, and in some cases, rendering the resulting output unparseable.

action required: If you were relying on the escaping, you can retain the old behaviour by adding an annotation to an affected TriggerTemplate, `tekton.dev/old-escape-quotes: true"`
```

### DIFF
--- a/docs/triggertemplates.md
+++ b/docs/triggertemplates.md
@@ -120,3 +120,64 @@ As of Tekton Pipelines version
 embed resource specs. It is a best practice to embed each resource specs in the
 PipelineRun or TaskRun that uses the resource spec. Embedding the resource spec
 avoids a race condition between creating and using resources.
+
+## Templating Params
+
+When templating parameters into resources, a simple replacement on the string
+with the parameter name e.g. `$(tt.params.name)` is carried out.
+
+This means that for simple string / number values, replacements in the
+YAML should work fine.
+
+If the string could begin with a number e.g. `012abcd`, it might be misinterpreted by YAML as a
+number, which could cause an error, in which case you can put quotes around the
+templated parameter key, and it should solve the problem.
+
+## Escaping quoted strings.
+
+TriggerTemplate parameter values were previously escaped by simply replacing
+`"` with `\"` this could lead to problems when strings were already escaped, and
+generating invalid resources from the TriggerTemplate.
+
+No escaping is done on the templated variables, if you are inserting a JSON
+object as a template var, then you should not put it within a quoted string.
+
+This behaviour has been deprecated, if this breaks your templates, you can add
+an annotation to the TriggerTemplate.
+
+For example with the following JSON body:
+
+```json
+{
+  "title": "this is \"demo\" body",
+  "object": {
+    "name": "testing"
+  }
+}
+```
+
+If you have a TriggerBinding that extracts `$(body.title)` then when it's
+inserted into a TriggerTemplate it will be embedded as `this is a \"demo\"
+body`.
+
+By annotating the TriggerTemplate.
+
+```yaml
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: TriggerTemplate
+metadata:
+  name: escaped-tt
+  annotations:
+    triggers.tekton.dev/old-escape-quotes: "true"
+spec:
+  params:
+  - name: title
+    description: The title from the incoming body
+```
+
+This would pass the same body through as `this is a \""demo\"" body`, which is
+invalid JSON, but, if you were to use a value with `$(body.object)` in a
+template, and want it passed through as a quoted string, then this will work.
+
+This might be useful if you want a string of JSON that you want to parse in a
+command.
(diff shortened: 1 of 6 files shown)


dibyom:
```
        {"level":"error","ts":"2020-11-10T16:58:02.248Z","logger":"eventlistener","caller":"sink/sink.go:274","msg":"problem creating obj: &errors.errorString{s:\"couldn't unmarshal json: invalid character 'k' after object key:value pair\"}","knative.dev/controller":"eventlistener","/triggers-eventid":"742ms","/trigger":"","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.CreateResources\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:274\ngithub.com/tektoncd/triggers/pkg/sink.Sink.processTrigger\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:191\ngithub.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent.func1\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:101"}
        {"level":"error","ts":"2020-11-10T16:58:02.248Z","logger":"eventlistener","caller":"sink/sink.go:192","msg":"couldn't unmarshal json: invalid character 'k' after object key:value pair","knative.dev/controller":"eventlistener","/triggers-eventid":"742ms","/trigger":"","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.processTrigger\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:192\ngithub.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent.func1\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:101"}
```

pkg/template/event.go
```diff
@@ -22,9 +22,15 @@ import (
 	"net/http"
 	"strings"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 )
 
+const (
+	OldEscapeAnnotation = "tekton.dev/old-escape-quotes"
```
wlynch, line 31:
```suggestion
	OldEscapeAnnotation = "triggers.tekton.dev/old-escape-quotes"
```

wlynch, line 31: Add a comment explaining what this controls, when it can be removed, etc.

docs/triggertemplates.md
```diff
@@ -120,3 +120,28 @@ As of Tekton Pipelines version
 embed resource specs. It is a best practice to embed each resource specs in the
 PipelineRun or TaskRun that uses the resource spec. Embedding the resource spec
 avoids a race condition between creating and using resources.
+
+
+## Quote-Escaping
```
wlynch, line 125: This primarily focuses on what the old behavior was, but doesn't really explain what the current behavior is. Let's add some more details here.

test/eventlistener_test.go
```diff
@@ -62,13 +62,13 @@ const (
 	examplePRJsonFilename = "pr.json"
 )
 
-func loadExamplePREventBytes() ([]byte, error) {
+func loadExamplePREventBytes(t *testing.T) []byte {
```
wlynch, line 65: It'll likely be useful to add a `t.Helper()` here.

wlynch (commented):
Overall looks good, just some small tweaks.

/lgtm

docs/triggertemplates.md
bigkevmcd: I'm not sure what to put here, the strings are no longer escaped at all, which means that it's up to you to put the correct quoting around params if you expect the values to be quoted or not.

dibyom: One thing that could be useful is an example of what a templated JSON body would look like now and with the annotation. something like- previously or with the annotation `{\"foo\": \"bar\"}` and while now it will be `{"foo": "bar"}`

dibyom:
Looks good. One question -- if I am using a TriggerTemplate that is embedded inside a Trigger resource, can I still use the annotation on the Trigger resource itself or does this only apply TriggerTemplates used via reference?

dibyom:
> can I still use the annotation on the Trigger resource itself or does this only apply TriggerTemplates used via reference?

It doesn't work for embedded trigger templates yet but at the WG we decided this would be fine to start off with. We can add the annotation support for embedded templates if users ask for it.

PR #828, opened 2020-11-13
TEP-0022: Switch to immutable input event bodies

dibyom:
# Changes
There are two commits in this PR:

1. TEP-0022: Add interface to support event bodies

This commit adds two new types for InterceptorRequest and
InterceptorResponse. Instead of just forwarding the incoming request,
EventListeners will use these types to send requests to interceptors.

This commit introduces an `extensions` field where Interceptors can add
additional data that is available to TriggerBindings. This replaces the
previous mechanism where Interceptors could freely modify the input
bodies. With this change, the input bodies are immutable, and any extra
fields added by Interceptors can only be done within the `extensions`
field.

This commit simply adds the new interface types and plumbing. Future
commits will change the Interceptors to adopt the new Interface.

2. Migrate CEL to new Interceptor Interface

This commit updates the CEL interceptor to use the new interface that supports
immutable input event bodies. This represents a BREAKING CHANGE for users using
`overlays`. Overlays will no longer modify the request body. Instead `keys`
from overlays will be added to a new top-level `extensions` field that is
accessible to TriggerBindings.

# Submitter Checklist

review them:

# Release Notes

```release-note
action required: If you are using `overlays` in the CEL Interceptor, please update your bindings to use $(extensions.<overlay-key>)  instead of $(body.<overlay-key>)

BREAKING CHANGE:
CEL overlays now add fields to a new top level `extensions` field instead of the modifying the incoming event body. TriggerBindings can access values within this new `extensions` field using `$(extensions.<key>)` syntax.
```


bigkevmcd:
Looks good to me.

pkg/interceptors/cel/cel.go
```diff
@@ -57,52 +60,136 @@ var (
...
-	env, err := makeCelEnv(request, w.EventListenerNamespace, w.KubeClientSet)
+// We need to keep this method around till we migrate the other interceptors to the new interface
+func (w *Interceptor) ExecuteTrigger(_ *http.Request) (*http.Response, error) {
+	return nil, fmt.Errorf("executeTrigger() is deprecated. Call Process() instead")
+}
+
+func evaluate(expr string, env *cel.Env, data map[string]interface{}) (ref.Val, error) {
+	parsed, issues := env.Parse(expr)
+	if issues != nil && issues.Err() != nil {
+		return nil, fmt.Errorf("failed to parse expression %#v: %s", expr, issues.Err())
```
wlynch, line 81: Return a wrapped error (same with other occurrences).
```suggestion
		return nil, fmt.Errorf("failed to parse expression %#v: %w", expr, issues.Err())
```

pkg/sink/sink.go
```diff
@@ -195,24 +196,30 @@ func (r Sink) processTrigger(t *triggersv1.EventListenerTrigger, request *http.R
...
+		Header:     in.Header.Clone(),
+		Extensions: map[string]interface{}{}, // Empty extensions for the first interceptor in chain
+		//InterceptorParams: ip, // To be added by the initial interceptor
+		Context: &triggersv1.TriggerContext{
+			EventURL: in.URL.String(),
+			EventID:  eventID,
+			// t.Name might not be fully accurate until we get rid of triggers inlined within EventListener
+			// that can only be
+			TriggerID: fmt.Sprintf("namespaces/%s/triggers/%s", r.EventListenerNamespace, t.Name),
+		},
```
wlynch, line 219:
```suggestion
		Context: &triggersv1.TriggerContext{
			EventURL: in.URL.String(),
			EventID:  eventID,
			// t.Name might not be fully accurate until we get rid of triggers inlined within EventListener
			// that can only be
			TriggerID: fmt.Sprintf("namespaces/%s/triggers/%s", r.EventListenerNamespace, t.Name),
		},
		// InterceptorParams will be added by the initial interceptor

```

pkg/sink/sink.go
```diff
@@ -195,24 +196,30 @@ func (r Sink) processTrigger(t *triggersv1.EventListenerTrigger, request *http.R
 	return nil
 }
 
-func (r Sink) ExecuteInterceptors(t *triggersv1.EventListenerTrigger, in *http.Request, event []byte, log *zap.SugaredLogger) ([]byte, http.Header, error) {
+// ExecuteInterceptor executes all interceptors for the Trigger and returns back the body, header, and extensions to use.
+// When TEP-0026 is fully implemented, this function will only return the added extensions.
+func (r Sink) ExecuteInterceptors(t *triggersv1.EventListenerTrigger, in *http.Request, event []byte, log *zap.SugaredLogger, eventID string) ([]byte, http.Header, map[string]interface{}, error) {
```
wlynch, line 201: We might want to return `([]byte, http.Header, *InterceptorResponse, error)` so that we can distinguish "we failed to execute the interceptor" vs "the interceptor returned an error, but we processed it correctly" errors.

pkg/sink/sink.go
```diff
@@ -223,33 +230,57 @@ func (r Sink) ExecuteInterceptors(t *triggersv1.EventListenerTrigger, in *http.R
...
-		request = &http.Request{
-			Method: http.MethodPost,
-			Header: resp.Header,
-			URL:    in.URL,
-			Body:   ioutil.NopCloser(resp.Body),
+		// Immutable Interface check
+		if interceptorInterface, ok := interceptor.(triggersv1.InterceptorInterface); ok {
+			// Set per interceptor config params to the request
+			request.InterceptorParams = interceptors.GetInterceptorParams(i)
+			interceptorResponse = interceptorInterface.Process(context.Background(), &request)
```
wlynch, line 244: We should add the context as a parameter so that we can accurately pass through timeouts and tracing information (if any).

wlynch (changes requested):
Overall looking good!

pkg/sink/sink.go
khrm: We can differentiate that based on error also but InterceptorResponse along with Error is probably better. Do we need to document this behaviour so that other interceptor also return InterceptorResponse like this?

pkg/sink/sink.go
dibyom: done

pkg/interceptors/cel/cel.go
```diff
@@ -57,52 +60,136 @@ var (
...
+	if err := json.Unmarshal(b, &p); err != nil {
+		// Should never happen since Unmarshal only returns err if json is invalid which we already check above
+		return &triggersv1.InterceptorResponse{
+			Continue: false,
+			Status:   status.Newf(codes.InvalidArgument, "invalid json: %v", err),
+		}
+	}
+	ns, _ := triggersv1.ParseTriggerID(r.Context.TriggerID)
+
+	env, err := makeCelEnv(nil, ns, w.KubeClientSet)
```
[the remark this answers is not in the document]
dibyom: done!

pkg/interceptors/cel/cel.go
dibyom: done.

pkg/sink/sink.go
dibyom: done.

pkg/sink/sink.go
dibyom: ack. the only reason I did not do it here was because there was no existing context to plumb through. Was planning on adding it a follow up PR where we move the interceptors to run in a separate HTTP service. What do you think?

wlynch: sgtm

wlynch (approved):
/lgtm

Issue #271, opened 2019-12-11
FR: Core Interceptor plugins


wlynch:
# Summary

* TEP: https://github.com/tektoncd/community/blob/master/teps/0026-interceptor-plugins.md#work-plan

* Work Plan/Progress: https://github.com/tektoncd/triggers/issues/271#issuecomment-724830692

# Expected Behavior

Core Interceptors should be pluggable so that users can choose which are available by default.

# Actual Behavior

Interceptors are currently static.

# Additional Info

https://github.com/hashicorp/go-plugin is a potential candidate for supporting in-binary plugins.

ncskier:
Waiting for design doc

dibyom:
I wrote up a design proposal: https://docs.google.com/document/d/1zIG295nyWonCXhb8XcOC41q4YVP9o2Hgg8uAoL4NdPA/edit#heading=h.8z1iarlx6yyx

Planning on filing TEP after discussion.

dibyom:
Pending the TEP getting merged, here is the implementation plan:

- [ ]  Migrate core interceptor to new interface structs - We'll migrate the core interceptors to use the new API. In this step, the core interceptors will still run inside the EventListener but will use go structs from the new API. 

- [ ] Implement [TEP-0022](https://github.com/tektoncd/community/pull/218) i.e. modify the CEL interceptor to return extensions instead of modifying the body.

- [ ]  Move the core interceptors to their own deployments - This will involve implementing the core interceptors to their own deployment and using HTTP to communicate. In addition, we will also be adding the InterceptorConfiguration CRD as part of this.

- [ ]  Add TLS support - This will involve adding suport for the `caBundle` field to allow for TLS communication between EventListener and Interceptor.

- [ ] Implement additional features - extra context for debugging, versioning etc.

dibyom:
Shipped in v0.13. Closing.

PR #854, opened 2020-12-04
Use our own Status type for InterceptorResponse

dibyom:
We were using GRPC's `status.Status` as the type for the `Status` field in
InterceptorResponse. This struct contains an embedded proto status message.
This means that we'd have to use a custom Marshal/Unmarshal function for the
Status to correctly contain the Code and Message fields. Instead, this commit
creates our own Status struct that mimics the GRPC status. We will still use
the same Status codes from GRPC but having our own type makes the JSON
encoding/decoding process significantly less complex.

In the future, we'd have to come up with a way to represent the Details field
(which is of proto.Any type) as well.

Part of #271

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

/kind misc


pkg/apis/triggers/v1alpha1/interceptor_types.go
```diff
@@ -46,7 +47,26 @@ type InterceptorResponse struct {
 	// Continue indicates if the EventListener should continue processing the Trigger or not
 	Continue bool `json:"continue,omitempty"`
 	// Status is an Error status containing details on any interceptor processing errors
-	Status *status.Status `json:"status,omitempty"`
+	Status Status `json:"status"`
+}
+
+type Status struct {
+	// The status code, which should be an enum value of [google.rpc.Code][google.rpc.Code].
+	Code int32 `json:"code,omitempty"`
```
wlynch, line 55: Should we do this instead? This would make it so we don't need to do the codes.Code(code) / int32() casting below.

```suggestion
	Code codes.Code `json:"code,omitempty"`
```

wlynch (commented):
/lgtm

One minor comment.

pkg/apis/triggers/v1alpha1/interceptor_types.go
dibyom: done!

savitaashture (commented):
For new `status` need to add test scenario in interceptor_types_test.go
Otherwise
/lgtm

dibyom:
> For new status need to add test scenario in interceptor_types_test.go

Yes, right now there is just a simple constructor so I didn't add any. The next PR has test cases that test that the Status returns the right info.

wlynch (approved):
/lgtm

PR #855, opened 2020-12-04
refactor: Do not use %w unless in fmt.Errorf

### LINKED ISSUE #853: 0.10.x: cryptic error message when not changed over to `extensions` in CEL overlays
After upgrading to 0.10.1, I missed an overlay CEL that referenced `body.var` and not `extensions.body.var`  this message maybe could be clearer as to what the issue is?

```
{"level":"info","ts":"2020-12-03T18:07:41.355Z","logger":"eventlistener","caller":"sink/sink.go:214","msg":"interceptor stopped trigger processing: %!w(*status.Error=&{0xc000a38cc0})","knative.dev/controller":"eventlistener","/triggers-eventid":"c8km5","/trigger":"slack-deploy-interceptor"}
```

dibyom:
# Changes

We were using %w in some Zap logger calls. %w is only supported in `fmt.Errorf`
calls and calling it in `fmt.Sprintf` will cause `go vet` to fail.
Unfortunately, since the call to Sprintf happens inside the zap library, we do
not trigger the go vet failure.

Also, cleaned up some old usages of xerrors.Errorf to just fmt.Errorf.

Fixes #853

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

/kind bug

### DIFF
--- a/cmd/triggerrun/cmd/root.go
+++ b/cmd/triggerrun/cmd/root.go
@@ -254,7 +254,7 @@ func newSink(config *rest.Config, sugerLogger *zap.SugaredLogger) sink.Sink {
 	dynamicCS := dynamicClientset.New(tekton.WithClient(dynamicClient))
 	kubeClient, _, err := getKubeClient(kubeconfig)
 	if err != nil {
-		log.Fatal("fail to get clients: %w", err)
+		log.Fatalf("fail to get clients: %v", err)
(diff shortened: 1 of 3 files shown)


savitaashture (commented):
/lgtm

khrm (approved):

PR #858, opened 2020-12-07
Adds a server for serving core interceptors.

dibyom:
# Changes

There are two commits:

* interceptors: small changes to types

1. Change body from []byte to json.RawMessage
2. Remove omitempty tag for boolean type so that it marshals when false
3. Change context.url to context.event_url to match struct type 

*  Add a HTTP handler for serving core interceptors

This commit packages the 4 core interceptors into a single HTTP server.
Each interceptor is available at a different path e.g. /cel for CEL etc.

This does not wire up the implementation of the server into the
EventListener which will be done in a future commit.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

/kind feature

### DIFF
--- a/pkg/apis/triggers/v1alpha1/interceptor_types.go
+++ b/pkg/apis/triggers/v1alpha1/interceptor_types.go
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -9,6 +10,10 @@ import (
 )
 
 type InterceptorInterface interface {
+	// Process executes the given InterceptorRequest. Simply getting a non-nil InterceptorResponse back is not sufficient
+	// to determine if the interceptor processing was successful. Instead use the InterceptorResponse.Status.Continue to
+	// see if processing should continue and InterceptorResponse.Status.Code to distinguish between the kinds of errors
+	// (i.e user errors vs system errors)
 	Process(ctx context.Context, r *InterceptorRequest) *InterceptorResponse
 }
 
@@ -16,7 +21,7 @@ type InterceptorInterface interface {
 // +k8s:deepcopy-gen=false
 type InterceptorRequest struct {
 	// Body is the incoming HTTP event body
-	Body []byte `json:"body,omitempty"`
+	Body json.RawMessage `json:"body,omitempty"`
 	// Header are the headers for the incoming HTTP event
 	Header map[string][]string `json:"header,omitempty"`
 	// Extensions are extra values that are added by previous interceptors in a chain
@@ -26,12 +31,12 @@ type InterceptorRequest struct {
 	InterceptorParams map[string]interface{} `json:"interceptor_params,omitempty"`
 
 	// Context contains additional metadata about the event being processed
-	Context *TriggerContext
+	Context *TriggerContext `json:"context"`
 }
 
 type TriggerContext struct {
 	// EventURL is the URL of the incoming event
-	EventURL string `json:"url,omitempty"`
+	EventURL string `json:"event_url,omitempty"`
 	// EventID is a unique ID assigned by Triggers to each event
 	EventID string `json:"event_id,omitempty"`
 	// TriggerID is of the form namespace/$ns/triggers/$name
@@ -42,10 +47,9 @@ type TriggerContext struct {
 // +k8s:deepcopy-gen=false
 type InterceptorResponse struct {
 	// Extensions are additional fields that is added to the interceptor event.
-	// See TEP-0022. Naming TBD.
 	Extensions map[string]interface{} `json:"extensions,omitempty"`
 	// Continue indicates if the EventListener should continue processing the Trigger or not
-	Continue bool `json:"continue,omitempty"`
+	Continue bool `json:"continue"` //Don't add omitempty -- it  will remove the continue field when the value is false.
 	// Status is an Error status containing details on any interceptor processing errors
 	Status Status `json:"status"`
 }
--- a/pkg/interceptors/cel/cel.go
+++ b/pkg/interceptors/cel/cel.go
@@ -61,7 +61,7 @@ var (
 )
 
 // NewInterceptor creates a prepopulated Interceptor.
-func NewInterceptor(k kubernetes.Interface, l *zap.SugaredLogger) *Interceptor {
+func NewInterceptor(k kubernetes.Interface, l *zap.SugaredLogger) interceptors.Interceptor {
 	return &Interceptor{
 		Logger:        l,
 		KubeClientSet: k,
--- a/pkg/interceptors/cel/cel.go
+++ b/pkg/interceptors/cel/cel.go
@@ -61,16 +61,11 @@ var (
 )
 
 // NewInterceptor creates a prepopulated Interceptor.
-func NewInterceptor(k kubernetes.Interface, l *zap.SugaredLogger) interceptors.Interceptor {
+func NewInterceptor(k kubernetes.Interface, l *zap.SugaredLogger) *Interceptor {
 	return &Interceptor{
 		Logger:        l,
 		KubeClientSet: k,
 	}
-}
-
-// We need to keep this method around till we migrate the other interceptors to the new interface
-func (w *Interceptor) ExecuteTrigger(_ *http.Request) (*http.Response, error) {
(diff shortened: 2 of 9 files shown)


pkg/interceptors/server/server.go
```diff
@@ -0,0 +1,123 @@
...
+		case Error:
+			is.Logger.Infof("HTTP %d - %s", e.Status(), e)
+			http.Error(w, e.Error(), e.Status())
+		default:
+			is.Logger.Errorf("Non Status Error: %s", err)
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		}
+	}
+	w.Header().Add("Content-Type", "application/json")
+	if _, err = w.Write(b); err != nil {
```
wlynch, line 57:
```suggestion
	if _, err := w.Write(b); err != nil {
```

pkg/interceptors/server/server.go
```diff
@@ -0,0 +1,123 @@
...
+		is.Logger.Errorf("failed to write response: %s", err)
```
wlynch, line 58: I think we need to explicitly set the response code - I'm not sure if this is set automatically if the write fails.

pkg/interceptors/server/server.go
```diff
@@ -0,0 +1,123 @@
...
+	defer r.Body.Close()
+	if _, err := io.Copy(&body, r.Body); err != nil {
+		return nil, internal(fmt.Errorf("failed to read body: %w", err))
+	}
+	var ireq v1alpha1.InterceptorRequest
+	if err := json.Unmarshal(body.Bytes(), &ireq); err != nil {
+		return nil, badRequest(fmt.Errorf("failed to parse body as InterceptorRequest: %w", err))
+	}
+	// Decorate ctx with eventID
+	iresp := ii.Process(ctx, &ireq)
```
wlynch, line 117: We should add documentation detailing how integrators should handle errors, both system and user - e.g. just getting an OK message doesn't mean that the request was successfully processed - they should inspect the response for errors as well.

pkg/interceptors/server/server.go
```diff
@@ -0,0 +1,123 @@
...
+	interceptors map[string]v1alpha1.InterceptorInterface
+}
+
+func NewWithCoreInterceptors(k kubernetes.Interface, l *zap.SugaredLogger) *Server {
+	i := map[string]v1alpha1.InterceptorInterface{
+		"bitbucket": bitbucket.NewInterceptor(k, l).(v1alpha1.InterceptorInterface),
+		"cel":       cel.NewInterceptor(k, l).(v1alpha1.InterceptorInterface),
+		"github":    github.NewInterceptor(k, l).(v1alpha1.InterceptorInterface),
+		"gitlab":    gitlab.NewInterceptor(k, l).(v1alpha1.InterceptorInterface),
+	}
```
wlynch, line 35: No action needed, but something to think about: I wonder if we should have libraries Register interceptors at init so that it's easier for other core interceptors to be added opaquely. 🤔  

(this is also making me think we should look into embedding the logger into the context)

pkg/apis/triggers/v1alpha1/interceptor_types.go
```diff
@@ -45,7 +46,7 @@ type InterceptorResponse struct {
 	// See TEP-0022. Naming TBD.
 	Extensions map[string]interface{} `json:"extensions,omitempty"`
 	// Continue indicates if the EventListener should continue processing the Trigger or not
-	Continue bool `json:"continue,omitempty"`
+	Continue bool `json:"continue"` // Adding omitempty will remove the continue field when the value is false.
```
wlynch, line 49: nit: Reworded to make it clear to not do this (on first pass I read this like a TODO 😅 )

```suggestion
	Continue bool `json:"continue"` // Don't add omitempty - this will remove the continue field when the value is false.
```

pkg/interceptors/cel/cel.go
```diff
@@ -61,7 +61,7 @@ var (
 )
 
 // NewInterceptor creates a prepopulated Interceptor.
-func NewInterceptor(k kubernetes.Interface, l *zap.SugaredLogger) *Interceptor {
+func NewInterceptor(k kubernetes.Interface, l *zap.SugaredLogger) interceptors.Interceptor {
```
wlynch, line 64: I think we want to still return *Interceptor here - by doing this you're masking the `triggersv1.InterceptorInterface` and other methods which limits how this can be used by callers.

If you want static type checking use a similar trick as the `triggersv1.InterceptorInterface` var above.

wlynch: If you do this I think you can also get rid of the type assertion in server.go below.

pkg/interceptors/server/server.go
```diff
@@ -0,0 +1,123 @@
...
+	Logger       *zap.SugaredLogger
+	interceptors map[string]v1alpha1.InterceptorInterface
+}
+
+func NewWithCoreInterceptors(k kubernetes.Interface, l *zap.SugaredLogger) *Server {
+	i := map[string]v1alpha1.InterceptorInterface{
+		"bitbucket": bitbucket.NewInterceptor(k, l).(v1alpha1.InterceptorInterface),
+		"cel":       cel.NewInterceptor(k, l).(v1alpha1.InterceptorInterface),
+		"github":    github.NewInterceptor(k, l).(v1alpha1.InterceptorInterface),
+		"gitlab":    gitlab.NewInterceptor(k, l).(v1alpha1.InterceptorInterface),
```
wlynch, line 34: We should check to make sure these are not nil.

pkg/interceptors/server/server.go
dibyom: The first call to Write() will automatically call WriteHeader(200) if WriteHeader has not been called before: 

```
	// If WriteHeader is not called explicitly, the first call to Write
	// will trigger an implicit WriteHeader(http.StatusOK).
	// Thus explicit calls to WriteHeader are mainly used to
	// send error codes.
```

https://pkg.go.dev/net/http#ResponseWriter

pkg/apis/triggers/v1alpha1/interceptor_types.go
dibyom: done.

pkg/interceptors/server/server.go
wlynch: Yup! I don't know if the same holds true if the call to w.Write fails though - we should probably return a 500 or similar. I forget what the behavior is if we reach the end of the handler without and successful writes - I want to make sure we don't accidentally return OK if the write fails.

pkg/interceptors/server/server.go
dibyom: done.

pkg/interceptors/server/server.go
dibyom: yes, we need a way to dynamically register interceptors at startup. Once we have the interceptor config crd, we can have an informer/lister to fetch all known interceptors either on startup or on the fly. 
If we need to, we could also add a way to register onto the core interceptor deployment too using a Register func that is called at init.

dibyom: on the logging from context -- yes, I plan on plumbing it through the context once we actually wire this up in the new binary.

pkg/interceptors/cel/cel.go
dibyom: So, I added this for this to be consistent with the other interceptors which all return interceptors.Interceptor (i.e. the old interface). I could flip it to return the new interface but that will be a slightly larger change -- we'll have to change some type assertions in sink.go since the webhook interceptor will still be using the old interface at the moment.

dibyom: ok, switched all of the interceptors to the new interface.

pkg/interceptors/server/server.go
dibyom: There isn't user facing docs yet but I've added this to the docstrings for the InterceptorInterface.

pkg/interceptors/server/server.go
dibyom: oh sorry, I missed this comment yesterday -- my understanding is that the headers and status code gets flushed first before the body is written. So, if we the body write fails somewhere in the middle, we don't really have a way to send 500s back anymore (since the status code has already been sent). The best we can do is log the error on the server side.

wlynch (approved):
/lgtm

PR #859, opened 2020-12-07
Merge extensions into body for webhook interceptors

### LINKED ISSUE #857: CEL - Webhook Interceptor Chaining broken
# Expected Behavior

It should be possible to add a field using a CEL overlay and then use that overlay value in a Webhook Interceptor.

# Actual Behavior

This behavior used to work but broke in v0.10.0/1 . This was due to the CEL interceptor moving to the new Interface where it can only write to the new `extensions` field while webhooks are still in the old interface and have no access to the `extensions` field.

# Additional Info

The easiest fix for this seems to be to address https://github.com/tektoncd/triggers/blob/master/pkg/interceptors/cel/cel.go#L233

i.e. we keep the old behavior of merging extensions in the body (temporarily as well).

dibyom:
With CEL interceptor writing to the `extensions` field, it is impossible to
chain added fields from a CEL interceptor to a webhook interceptor today. This
is because we do not pass the extensions field to the webhook interceptor yet.
This commit attempts to fix this by merging any extensions to the body before
sending it over to the Webhook interceptor. This is temporary until we fully
move webhook interceptors to the new pluggable interface. One implication of
this change is that the body may now contain an `extensions` field separate
from the top level extensions field if one uses the Webhook interceptor.

In addition, the CEL environment also did not have access to extensions which I
fixed.

Fixes #857

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
 Extensions added by a CEL Interceptor will be passed on to webhook interceptors by merging the extension fields into the event body under a `extensions` field. See docs/eventlisteners.md##chaining-interceptors for more details.
```

### DIFF
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -34,6 +34,7 @@ using [Event Interceptors](#Interceptors).
     - [Bitbucket Interceptors](#bitbucket-interceptors)
     - [CEL Interceptors](#cel-interceptors)
       - [Overlays](#overlays)
+    - [Chaining Interceptors](#chaining-interceptors)
   - [EventListener Response](#eventlistener-response)
   - [How does the EventListener work?](#how-does-the-eventlistener-work)
   - [Examples](#examples)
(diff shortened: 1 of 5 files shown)


pkg/sink/sink.go
```diff
@@ -354,3 +360,20 @@ func (r Sink) CreateResources(triggerNS, sa string, res []json.RawMessage, trigg
 	}
 	return nil
 }
+
+// extendBodyWithExtensions merges the extensions into the given body.
+func extendBodyWithExtensions(body []byte, extensions map[string]interface{}) ([]byte, error) {
+	newBody := body
```
bigkevmcd, line 366: I don't think you need to reassign body here?

bigkevmcd (commented):
Couple of typos that should be tidied up, but otherwise looks good to me.

docs/eventlisteners.md
```diff
@@ -679,6 +680,41 @@ spec:
...
+      - webhook:
+        objectRef:
+          kind: Service
+          name: some-interceptor
+          apiVersion: v1
+      - cel:
+          filter: "body.extensions.truncated_sha == \"abcde\"" # Can also be extensions.truncated_sha == \"abcde\"
+  ```
+
+In the above snipped, the first CEL interceptor adds the `truncated_sha` field. To ensure the following webhook interceptor can use this field, the EventListener will add it to the body. So, the body recevied by the webbhok interceptor will look as follows:
```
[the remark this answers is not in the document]
dibyom: done

pkg/sink/sink.go
dibyom: done

wlynch (approved):
/lgtm

Issue #781, opened 2020-09-30
chdir to cwd "/home/nonroot" set in config.json failed: permission denied": unknown


rannox:
Hello everyone, 

updating the version of the Tekton Triggers from 0.6.1 to either 0.7.0 or 0.8.1 fails with:
 `"chdir to cwd (\"/home/nonroot\") set in config.json failed: permission denied": unknown`

# Expected Behavior

Installation of the Tekton Triggers should work. 

# Actual Behavior

The start of the webhook and controller pods fails:

```
$ kubectl get pods --namespace tekton-pipelines
NAME                                          READY   STATUS             RESTARTS   AGE
tekton-triggers-controller-77fcb944f9-vws29   0/1     CrashLoopBackOff   259        21h
tekton-triggers-webhook-dc7c546b9-9nbjv       0/1     CrashLoopBackOff   259        21h
```

```
$ kubectl describe pod tekton-triggers-controller-77fcb944f9-vws29 -n tekton-pipelines
...
Events:
  Type     Reason     Age                 From               Message
  ----     ------     ----                ----               -------
  Normal   Scheduled  2m7s                default-scheduler  Successfully assigned tekton-pipelines/tekton-triggers-controller-77fcb944f9-vws29 to [HOST]
  Normal   Created    84s (x4 over 2m6s)  kubelet, [HOST]  Created container tekton-triggers-controller
  Warning  Failed     83s (x4 over 2m6s)  kubelet, [HOST]  Error: failed to start container "tekton-triggers-controller": Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "chdir to cwd (\"/home/nonroot\") set in config.json failed: permission denied": unknown
  Warning  BackOff    44s (x7 over 2m4s)  kubelet, [HOST]  Back-off restarting failed container
  Normal   Pulling    37s (x5 over 2m6s)  kubelet, [HOST]  Pulling image "[PRIVATE_REGISTRY]/gcr.io/tekton-releases/github.com/tektoncd/triggers/cmd/controller:v0.7.0"
  Normal   Pulled     37s (x5 over 2m6s)  kubelet, [HOST]  Successfully pulled image "[PRIVATE_REGISTRY]/gcr.io/tekton-releases/github.com/tektoncd/triggers/cmd/controller:v0.7.0"

```

# Steps to Reproduce the Problem

[...]

PR #789, opened 2020-10-07
Set runAsUser uid for nonroot distroless image

dibyom:
# Changes

Port of tektoncd/pipeline#3342:

The distroless nonroot image define a user with the uid 65532 and not 1001. The
deployment should use that uid to make sure it works anywhere.

Fixes #781

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [x] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/triggers/blob/master/CONTRIBUTING.md) for more details._

# Release Notes
```release-note
Fix inconsistent uid for the controller and webhook deployment, resulting in failure of installing tekton pipeline on minikube (and other platforms.)
```

PR #862, opened 2020-12-14
Add SecurityContext to reconciled Deployment

MarcelMue:
# Changes

Adds a SecurityContext to the created Deployment - changes match those suggested here: https://github.com/tektoncd/triggers/issues/781

# Submitter Checklist

review them:

~- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#docs) (if user facing)~

# Release Notes

```release-note
Add a security context on event listener deployments to prevent issues with restrictive default PSPs.
```

### DIFF
--- a/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
+++ b/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
@@ -364,6 +364,10 @@ func (r *Reconciler) reconcileDeployment(ctx context.Context, logger *zap.Sugare
 				existingDeployment.Spec.Template.Spec.Volumes = deployment.Spec.Template.Spec.Volumes
 				updated = true
 			}
+			if !reflect.DeepEqual(existingDeployment.Spec.Template.Spec.SecurityContext, deployment.Spec.Template.Spec.SecurityContext) {
+				existingDeployment.Spec.Template.Spec.SecurityContext = deployment.Spec.Template.Spec.SecurityContext
+				updated = true
+			}
 		}
 		if updated {
 			if _, err := r.KubeClientSet.AppsV1().Deployments(el.Namespace).Update(ctx, existingDeployment, metav1.UpdateOptions{}); err != nil {
@@ -461,6 +465,10 @@ func getDeployment(el *v1alpha1.EventListener) *appsv1.Deployment {
 					ServiceAccountName: serviceAccountName,
 					Containers:         []corev1.Container{container},
 					Volumes:            vol,
+					SecurityContext: &corev1.PodSecurityContext{
+						RunAsNonRoot: ptr.Bool(true),
+						RunAsUser:    ptr.Int64(65532),
+					},
 				},
 			},
 		},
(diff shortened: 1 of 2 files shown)


MarcelMue:
Sorry for the dumb questions but this is my first time contributing to tekton upstream:
1. Is this user facing? I would say yes because the behaviour does change slightly - not sure though.
2. CI is taking really long to run - is this expected or is something wrong with my PR?
3. I want to squash commits when I think changes are ready to merge - is that okay?

dibyom:
Thanks for opening the PR @MarcelMue :)

> Is this user facing? I would say yes because the behaviour does change slightly - not sure though.

Yeah I'd agree. We should add a release note to the PR description like in https://github.com/tektoncd/triggers/pull/789

> CI is taking really long to run - is this expected or is something wrong with my PR?

It should take a few mins. Looks like its reporting that its still running after an hour. I'll retrigger the tests (I've seen this happen a few times before)

> I want to squash commits when I think changes are ready to merge - is that okay? 

Yes, that's ok!

vdemeester:
@dibyom so… this breaks `OpenShift`. As this is user facing, I would have suggested to add a way to configure this behavior somehow (and disable it in some cases)

MarcelMue:
> @dibyom so… this breaks `OpenShift`. As this is user facing, I would have suggested to add a way to configure this behavior somehow (and disable it in some cases)

Interesting - can you elaborate to me how it breaks `OpenShift` (I have never worked with it)?

I assumed this was safe because it is on master `yaml`s without being configurable - e.g.:
https://github.com/tektoncd/triggers/blob/0a9df14bf5731468e5cfc49d065a41cebfe82de6/config/controller.yaml#L78-L81

vdemeester:
Hey @MarcelMue, more detail in #864, but in a gist, OpenShift needs some "extra rights" to be able to set a given uid to a Pod. By default, OpenShift does use a random uid, and it works just fine for the eventlistener pods. This changes forces to attach to the eventlistener pod a serviceAccount with `anyuid` — which adds a little bit of burden to *operators* of tektoncd/triggers instances (and even more to the user if one wants to add its own ServiceAccount if that's possible).

vdemeester:
> I assumed this was safe because it is on master `yaml`s without being configurable - e.g.:
> https://github.com/tektoncd/triggers/blob/0a9df14bf5731468e5cfc49d065a41cebfe82de6/config/controller.yaml#L78-L81

Indeed 😉 This is another *small* problem that we may have, but when we package Tekton for OpenShift, we do control the payload that we apply and thus we can safely remove that part (`runAsUser`) — but we do not control the code that is in `tektoncd/triggers`, so there is no easy way to remove the `runAsUser` set by the code in that PR.

@MarcelMue note that, I am 100% for this change in standard k8s. We just need to find a way to disable this behavior on demand (but having it enabled by default in a standard distribution of tektoncd).

dbazhal:
definately it broken our triggers in openshift with "unable to validate against any security context constraint: [spec.containers[0].securityContext.runAsUser: Invalid value: 65532: must be in the ranges: [1000750000, 1000759999]]"

PR #863, opened 2020-12-15
Bind admin role so that SA can use PSP

fiunchinho:
# Changes

Examples fail because the `ServiceAccount` used has no permissions to use the `PodSecurityPolicy`. We need to bind the admin role to the `ServiceAccount` so that it can use the PSP.

# Submitter Checklist

review them:

### DIFF
--- a/examples/rbac.yaml
+++ b/examples/rbac.yaml
@@ -24,6 +24,10 @@ rules:
 - apiGroups: [""]
   resources: ["serviceaccounts"]
   verbs: ["impersonate"]
+- apiGroups: ["policy"]
+  resources: ["podsecuritypolicies"]
+  resourceNames: ["tekton-triggers"]
+  verbs: ["use"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding


examples/rbac.yaml
```diff
@@ -37,6 +37,18 @@ roleRef:
   kind: Role
   name: tekton-triggers-example-minimal
 ---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
```
dibyom, line 41: Instead of adding a binding to the admin role, could we add the [rule](https://github.com/tektoncd/triggers/blob/master/config/200-role.yaml#L30-L33) of using pod security policies to the role defined above in this file?

dibyom:
Question: Is this needed due to https://github.com/tektoncd/triggers/pull/862?

Also, /cc @gabemontero our resident PSP expert!

gabemontero:
> Question: Is this needed due to #862?
> 

Yes I believe it is @dibyom .... the slack discussion thread we've had with these folks on their issue is https://tektoncd.slack.com/archives/CKUSJ2A5D/p1607949867225500

Subsequent to getting this to work for them, I am curious how long triggers has been broke with PSP turned on.  Or if it ever worked with PSP turned on.  I would think it had worked with PSP sometime in the past but am not certain.  If it did, that might mean something changed in the image or controller to necessitate these changes.

By comparison, the PSP analogous feature in openshift, SCC, (which is what I typically run with), has not seen this (unless we need to retry with the very latest ... it has been a couple of weeks for me).  @savitaashture attested to the same experience.

> Also, /cc @gabemontero our resident PSP expert!

oooh, that is a very dangerous characterization man ;-) .... "expert" at best is a relative term here :-)

dibyom:
> Subsequent to getting this to work for them, I am curious how long triggers has been broke with PSP turned on. Or if it ever worked with PSP turned on. I would think it had worked with PSP sometime in the past but am not certain. If it did, that might mean something changed in the image or controller to necessitate these changes.

well, #862 got merged yesterday so I'm guessing since then 😛 The conroller/webhook already had roles to use PSPs. The EL sink deployment did not have this role, but the deployment spec also did not have a `securityContext` (until yesterday).

gabemontero:
> > Subsequent to getting this to work for them, I am curious how long triggers has been broke with PSP turned on. Or if it ever worked with PSP turned on. I would think it had worked with PSP sometime in the past but am not certain. If it did, that might mean something changed in the image or controller to necessitate these changes.
> 
> well, #862 got merged yesterday so I'm guessing since then stuck_out_tongue The conroller/webhook already had roles to use PSPs. The EL sink deployment did not have this role, but the deployment spec also did not have a `securityContext` (until yesterday).

the opposite actually ... I suggested that to them, along with this PR,  to help get past this ... basically mimic what we do for the controller / webhook ... they were seeing the issue before that merged

dibyom:
Makes sense. controllers/webhook had the right policies/securityContext while EL sinks did not. So in environments requiring PSPs, the controller/webhook would work but not the EL.

MarcelMue:
I will take over this PR from @fiunchinho - he is currently on vacation but I'd like to push this forward in the meantime.

dibyom:
> I will take over this PR from @fiunchinho - he is currently on vacation but I'd like to push this forward in the meantime.

Thank you 🙏

fiunchinho:
Addressed review comment, please take a look. Thanks.

dibyom:
Looks good. One small thing -- could you squash up the two commits? Thanks!

/approve

fiunchinho:
> Looks good. One small thing -- could you squash up the two commits? Thanks!
> 
> /approve

Done

MarcelMue:
Nice :tada:

Issue #864, opened 2020-12-16
Add a configuration for SecurityContext to reconciled eventlistener Deployment


vdemeester:
With #862 is merged, eventlistener deployment are no more able to be created, *by default* on OpenShift.
The reason for it is that, by default, you cannot set `uid` for your Pods, OpenShift takes care of this. To be able to use `runAsUser: 65532` on OpenShift, you need to use `anyuid` which is a *higher* privilege than you usually need to. 

Before this change, the eventlistener pod(s), in OpenShift, would happily run with a random uid. With this change, the serviceAccount that runs the eventlistener pod(s) need to have the `anyuid` scc attached (see [here](https://docs.openshift.com/container-platform/4.6/authentication/managing-security-context-constraints.html))

The problem with #862 is that the behavior is not configurable, aka we can't disable it if we don't need it.
I propose to add a way to configure this behavior, most likely through a feature-flag (that would be enabled by default)

MarcelMue:
I will try to provide a PR with a fix as I introduced the change.

MarcelMue:
PR which adds a flag is open now. My main concern is that I am unable to test this properly - so would rely on someone trying this on openshift.

MarcelMue:
I added some more unit tests and am fairly confident now, that this can be closed.

PR #865, opened 2020-12-16
Add flag to disable security context on EL Deployment

MarcelMue:
# Changes

This is a follow up to #864 - it introduces a flag which removes the security context from all created deployments. This enables `triggers` to work correctly on OpenShift.

# Submitter Checklist

review them:

# Release Notes

```release-note
Add a flag to disable SecurityContext on the reconciled eventlistener Deployment
```

### DIFF
--- a/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
+++ b/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
@@ -61,12 +61,15 @@ const (
 )
 
 var (
-	// The container that we use to run in the EventListener Pods
-	elImage = flag.String("el-image", "override-with-el:latest",
+	// ELImage defines the container that we use to run in the EventListener Pods
+	ELImage = flag.String("el-image", "override-with-el:latest",
 		"The container image for the EventListener Pod.")
 	// ElPort defines the port for the EventListener to listen on
 	ElPort = flag.Int("el-port", 8080,
 		"The container port for the EventListener to listen on.")
+	// ELSetSecurityContext defines if the security context is set.
+	ELSetSecurityContext = flag.Bool("el-security-context", true,
+		"Add a security context to the event listener deployment.")
 	// ELReadTimeOut defines the read timeout for EventListener Server
 	ELReadTimeOut = flag.Int64("el-readtimeout", 5,
 		"The read timeout for EventListener Server.")
@@ -364,7 +367,7 @@ func (r *Reconciler) reconcileDeployment(ctx context.Context, logger *zap.Sugare
 				existingDeployment.Spec.Template.Spec.Volumes = deployment.Spec.Template.Spec.Volumes
 				updated = true
 			}
-			if !reflect.DeepEqual(existingDeployment.Spec.Template.Spec.SecurityContext, deployment.Spec.Template.Spec.SecurityContext) {
+			if !reflect.DeepEqual(existingDeployment.Spec.Template.Spec.SecurityContext, deployment.Spec.Template.Spec.SecurityContext) && *ELSetSecurityContext {
 				existingDeployment.Spec.Template.Spec.SecurityContext = deployment.Spec.Template.Spec.SecurityContext
 				updated = true
 			}
@@ -402,6 +405,7 @@ func getDeployment(el *v1alpha1.EventListener) *appsv1.Deployment {
 		tolerations                          []corev1.Toleration
 		nodeSelector, annotations, podlabels map[string]string
 		serviceAccountName                   string
+		securityContext                      corev1.PodSecurityContext
 	)
 	podlabels = mergeMaps(el.Labels, GenerateResourceLabels(el.Name))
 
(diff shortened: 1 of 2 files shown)


pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -61,12 +61,15 @@ const (
 )
 
 var (
-	// The container that we use to run in the EventListener Pods
-	elImage = flag.String("el-image", "override-with-el:latest",
+	// ELImage defines the container that we use to run in the EventListener Pods
```
MarcelMue, line 64: I renamed this flag to capitalize it like the other flags. Mainly for consistency.

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -61,12 +61,15 @@ const (
...
+	ELImage = flag.String("el-image", "override-with-el:latest",
 		"The container image for the EventListener Pod.")
 	// ElPort defines the port for the EventListener to listen on
 	ElPort = flag.Int("el-port", 8080,
 		"The container port for the EventListener to listen on.")
+	// ELSetSecurityContext defines if the security context is set.
+	ELSetSecurityContext = flag.Bool("el-security-context", true,
```
MarcelMue, line 71: I would like to unit test this but have no idea how. Are there any examples where it's easy to tell how flags can be invoked for unit tests in this project?

MarcelMue: Current idea for me is to re-factor the flag setup in this controller to allow unit testing different values.

savitaashture, line 71: I forgot to mention this during WG call 
If we keep default value as true then for `openshift` platform we need to update the controller deployment to set the flag to `false`
I believe which is extra step
so considering this shall we make default as `false`
@dibyom @bigkevmcd @gabemontero WDYT?

dibyom: One way might be to refactor one of the functions e.g. reconcileDeployment (or getDeployment) to take in a parameter called `isSecurityContexEnabled` and then we write unit tests for those functions (we lack those smaller unit tests today but maybe we should add them)

Another way might be to add these as fields in the Reconciler struct and we can set the right values in the test Reconcilers.

dibyom: Yeah, my understanding is that openshift already generates/modifies some of the YAML so it should be able to change/add the flag here: https://github.com/tektoncd/triggers/blob/master/config/controller.yaml#L62

Would that work?

gabemontero: Yeah @savitaashture I think upstream needs to default to being able to function securely out of the box, which means being able to function if PSP is set to more restrictive default policies like what @MarcelMue and Jose experienced.

Which means default to true, no?

And then if openshift has to tweak more yaml like we do already when dealing with SCC vs. PSP, that that is what we have to do.

@vdemeester @piyush-garg FYI

savitaashture: Yes :+1:

vdemeester: Agreed, default to true is ok for me 👍🏼

MarcelMue: > Another way might be to add these as fields in the Reconciler struct and we can set the right values in the test Reconcilers.

This is what I want to go for currently. I would like to do this as a follow up to this PR to prevent this from getting too big.

khrm: We will need to modify release.yaml for openshift because controller and webhook also have hardcoded UUID. So we can do for this flag also. We already do this for pipelines also.

MarcelMue:
I think this should be ready to merge now. Unit tests currently don't exist and will be added in a follow up. I can not find any place where we document these flags - correct me if I am wrong.

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
vdemeester: The *only* downside of the flag approach (compared to a configmap approach) is that a user that wants to run `upstream` triggers on OpenShift needs to edit the deployment, where, otherwise only a configmap edition is required (which is.. simpler..). But I am definitely fine with the flag 😉

dibyom: yeah in general, the number of flags are increasing...so it'd be nice to move to a configMap at some point.

dibyom: sounds good to me!

Issue #867, opened 2020-12-16
Interceptors: Migrate CEL and source control interceptors to a separate server


dibyom:
The work for this issue is to migrate the core interceptors (minus webhook) to run on their own deployment. The EL will send HTTP requests using the interface specified in [TEP-0026](https://github.com/tektoncd/community/blob/master/teps/0026-interceptor-plugins.md#http-interface).

Part of [TEP-0026](https://github.com/tektoncd/community/blob/master/teps/0026-interceptor-plugins.md#http-interface)

PR #877, opened 2020-12-22
Change InterceptorRequest.Body to string

dibyom:
# Changes

Storing the body as `json.RawMessage` can lead to loss of information about
spaces in the incoming body when the InterceptorRequest is marshaled as JSON
before sending it over HTTP.

This is because Go's `json.Marshal` will encode `[]byte` as a base64 string and
base64 does not preserve spaces. Adding a custom `MarshalJSON` does not help
here since `MarshalJSON` will return a `[]byte` that will then get compacted as
base64 by `json.Marshal`.

This loss of spaces can be problematic for some use cases. For instance, the
GitHub payload signature validation requires us to compare using the exact body
as it was sent. Otherwise, the validation fails. Note that this will only be a
problem when we marshal the InterceptorRequest i.e. when we move the core
interceptors out to its own server.

Using a string means that the string will be quoted e.g. `{\"foo\": \"bar\"}`.
Go should take care of unquoting it during the unmarshaling process. However,
intercetpor authors using a different language will have to manually unquote
the string by parsing it as a JSON object.

Part of #271, #867

# Submitter Checklist

review them:

# Release Notes

```release-note
InterceptorRequest.Body is now of type string. No immediate action is required but once #271 is fully implemented, interceptor authors will have to parse the Body separately as a JSON object.
```

### DIFF
--- a/pkg/apis/triggers/v1alpha1/interceptor_types.go
+++ b/pkg/apis/triggers/v1alpha1/interceptor_types.go
@@ -2,7 +2,6 @@ package v1alpha1
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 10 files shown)


pkg/apis/triggers/v1alpha1/interceptor_types.go
```diff
@@ -20,10 +19,14 @@ type InterceptorInterface interface {
 // Do not generate DeepCopy(). See #827
 // +k8s:deepcopy-gen=false
 type InterceptorRequest struct {
-	// Body is the incoming HTTP event body
-	Body json.RawMessage `json:"body,omitempty"`
+	// Body is the incoming HTTP event body. We use a "string" representation of the JSON body
+	// in order to preserve the body exactly as it was sent (including spaces etc.). This is necessary
```
wlynch, line 23: Let's also mention the json.Marshal/base64 behavior - while`[]byte` can be an exact representation of a string, it's the marshal behavior that makes it unusable.

wlynch (commented):
/lgtm

pkg/apis/triggers/v1alpha1/interceptor_types.go
dibyom: done!

wlynch (approved):
/lgtm

PR #878, opened 2020-12-23
Move core interceptors to their own server

### LINKED ISSUE #867: Interceptors: Migrate CEL and source control interceptors to a separate server

dibyom:
# Changes

The EventListener will forward events over HTTP to the interceptor.  Each core
interceptor is available at a different path e.g. /cel for CEL.

Fixes #867

Depends on #877

# Submitter Checklist

review them:

# Release Notes
```release-note
Core interceptors now run in their own server that is deployed in the tekton-pipelines namespace.
```


pkg/interceptors/interceptors.go
```diff
@@ -161,3 +170,52 @@ func UnmarshalParams(ip map[string]interface{}, p interface{}) error {
 	}
 	return nil
 }
+
+// ResolveURL returns the URL for the given core interceptor
+func ResolveURL(i *triggersv1.TriggerInterceptor) string {
+	// This is temporary until we implement #868
+	base := fmt.Sprintf("http://%s.%s.svc/", CoreInterceptorsHost, system.GetNamespace())
```
wlynch, line 177: Let's build this from a url.URL so we're confident that things are URL encoded properly - I'm paranoid about how weird values being injected here might interact (e.g. `http://example.com/my/other/path/.foo.svc/`)

(while I'm pretty sure that we're safe here since we're using const / system derived values, better to be safe).

pkg/sink/sink_test.go
```diff
@@ -100,13 +104,29 @@ func getSinkAssets(t *testing.T, resources test.Resources, elName string, auth A
...
+	coreInterceptors, err := server.NewWithCoreInterceptors(clients.Kube, logger.Sugar())
+	if err != nil {
+		t.Fatalf("failed to initialize core interceptors: %v", err)
+	}
+	srv := httptest.NewServer(coreInterceptors)
+	t.Cleanup(func() {
+		srv.Close()
+	})
+	httpClient := srv.Client()
+	u, _ := url.Parse(srv.URL)
```
wlynch, line 117: check error

pkg/sink/sink_test.go
```diff
@@ -1074,57 +1091,74 @@ func (f *echoInterceptor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	echoServer := &echoInterceptor{}
-	srv := httptest.NewServer(echoServer)
+	r.MatcherFunc(match).Handler(echoServer)
+	r.PathPrefix("/").Handler(coreInterceptors)
+
+	srv := httptest.NewServer(r)
 	defer srv.Close()
 	client := srv.Client()
-	// Redirect all requests to the fake server.
 	u, _ := url.Parse(srv.URL)
```
wlynch, line 1118: Even though this wasn't modified, we should check the error here as well

test/init_test.go
```diff
@@ -70,7 +70,7 @@ func tearDown(t *testing.T, cs *clients, namespace string) {
 		if err != nil {
 			t.Error(err)
 		} else {
-			t.Log(string(bs))
+			t.Logf("%s", bs)
```
wlynch, line 73:
```suggestion
			t.Log(bs)
```

wlynch (commented):
looks good! few minor comments.

pkg/interceptors/interceptors.go
dibyom: done.

pkg/interceptors/interceptors_test.go
```diff
@@ -14,24 +14,28 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package interceptors
+package interceptors_test
```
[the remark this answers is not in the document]
dibyom: Import cycle:

```
# github.com/tektoncd/triggers/pkg/interceptors
import cycle not allowed in test
package github.com/tektoncd/triggers/pkg/interceptors (test)
	imports github.com/tektoncd/triggers/pkg/interceptors/server
	imports github.com/tektoncd/triggers/pkg/interceptors/bitbucket
	imports github.com/tektoncd/triggers/pkg/interceptors
```

The only extra thing we needed to export is the RequestCacheKey -- its unused and I'm going to remove it in a followup PR.

pkg/interceptors/interceptors.go
wlynch: I think this suffers from the same issue, since the parse wouldn't be able to distinguish which pieces were meant for the host. Let's do something like this instead?

```go
func ResolveURL(i *triggersv1.TriggerInterceptor) *url.URL, error {
  path := ""
  switch {
    ...
  }
  return &url.URL{
    Scheme: "http",
    Host: fmt.Sprintf("%s.%s.svc", CoreInterceptorsHost, system.GetNamespace()),
    Path: path,
  }
}

dibyom: ahh ok, I think I get the specific case you were talking about above. Will make the change.

dibyom: done.

wlynch (approved):
/lgtm

PR #880, opened 2020-12-24
Update tekton eventlistener portname owing to istio naming conventions

sravankumar777:
https://istio.io/docs/ops/deployment/requirements/

# Changes
we should follow service port naming convention as stated here -> https://istio.io/docs/setup/kubernetes/additional-setup/requirements/
Update new latest release for ```tektoncd/trigger v0.10.2```. Creation of underlying eventlistener svc gives error about service portname as per istio naming conventions.
```
Ports in your service are missing name in the format <protocol>[-<suffix>]. Supported protocols - istio.io/docs/setup/kubernetes/spec-requirements : Operation cannot be fulfilled on deployments.apps "el-valid-eventlistener": the object has been modified; please apply your changes to the latest version and try again
```

# Release Note

```release-note
Event listener port is now called `http-listener` instead of `listener`.
```

Issue #586, opened 2020-05-27
Change event listener port


sebgoa:
question:

Is it possible to change the event listener port ?

Right now it serves on 8080, I would like to change it to 80.  Editing the services and setting the port does not work (I guess the controller is overriding it)

ncskier:
Unfortunately, I don't think this is an officially supported feature. However, you could modify your Tekton Triggers controller deployment called `tekton-triggers-controller` to change the [`-el-port`](https://github.com/tektoncd/triggers/blob/master/config/controller.yaml#L59) to 80:
```yaml
          "-el-port", "8080",
```

I don't think that this has been thoroughly tested, so I'll mark this issue as a feature request 👍 

/kind feature

This is related to https://github.com/tektoncd/triggers/issues/505, which is also about customizing the EventListener Service/Deployment.

Issue #318, opened 2020-01-07
Allow config listen port of Eventlistener


cccfeng:
# Expected Behavior
When I deploy the triggers project, I expect to be able to configure the port that EventListener listens to, which helps to configure EventListener's service more flexibly. 

# Actual Behavior
We internally register headless service in k8s, and the ingress service perform health check on port 80 by default. So it's not enough for us to only listen to port 8080.

# Steps to Reproduce the Problem

1.  Deploy triggers and apply eventListener
2. You will find that el-container and el-service listen 8080 

# Additional Info

PR #881, opened 2020-12-30
Fix behaviour which allows EL ports to be configured

MarcelMue:
# Changes

This was a lot more work than I first anticipated. Settled for this solution after trying around a few times.

The previous behaviour did not only prevent the port from being configured but also introduced some nasty side effects during reconciliation because the port value was basically written do a global variable.

# Submitter Checklist

review them:

# Release Notes

```release-note
Fix behaviour which allows EL ports to be configured.
```

### DIFF
--- a/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
+++ b/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
@@ -157,6 +157,8 @@ func (r *Reconciler) reconcileService(ctx context.Context, logger *zap.SugaredLo
 		serviceType = el.Spec.Resources.KubernetesResource.ServiceType
 	}
 
+	port := getPort(el, r.config)
+
 	service := &corev1.Service{
 		ObjectMeta: generateObjectMeta(el, r.config.StaticResourceLabels),
 		Spec: corev1.ServiceSpec{
@@ -165,9 +167,9 @@ func (r *Reconciler) reconcileService(ctx context.Context, logger *zap.SugaredLo
 			Ports: []corev1.ServicePort{{
 				Name:     eventListenerServicePortName,
 				Protocol: corev1.ProtocolTCP,
-				Port:     int32(*r.config.Port),
+				Port:     int32(port),
 				TargetPort: intstr.IntOrString{
-					IntVal: int32(*r.config.Port),
+					IntVal: int32(port),
 				}},
 			},
 		},
@@ -178,7 +180,7 @@ func (r *Reconciler) reconcileService(ctx context.Context, logger *zap.SugaredLo
 		// Determine if reconciliation has to occur
 		updated := reconcileObjectMeta(&existingService.ObjectMeta, service.ObjectMeta)
 		el.Status.SetExistsCondition(v1alpha1.ServiceExists, nil)
-		el.Status.SetAddress(listenerHostname(service.Name, el.Namespace, *r.config.Port))
+		el.Status.SetAddress(listenerHostname(service.Name, el.Namespace, port))
 		if !reflect.DeepEqual(existingService.Spec.Selector, service.Spec.Selector) {
 			existingService.Spec.Selector = service.Spec.Selector
 			updated = true
@@ -209,7 +211,7 @@ func (r *Reconciler) reconcileService(ctx context.Context, logger *zap.SugaredLo
 			logger.Errorf("Error creating EventListener Service: %s", err)
 			return err
 		}
-		el.Status.SetAddress(listenerHostname(service.Name, el.Namespace, *r.config.Port))
+		el.Status.SetAddress(listenerHostname(service.Name, el.Namespace, port))
 		logger.Infof("Created EventListener Service %s in Namespace %s", service.Name, el.Namespace)
 	default:
 		logger.Error(err)
@@ -491,7 +493,6 @@ func getContainer(el *v1alpha1.EventListener, c Config) corev1.Container {
 	}
 
 	if elCert != "" && elKey != "" {
-		*c.Port = 8443
 		scheme = corev1.URISchemeHTTPS
 		vMount = append(vMount, corev1.VolumeMount{
 			Name:      "https-connection",
@@ -499,7 +500,6 @@ func getContainer(el *v1alpha1.EventListener, c Config) corev1.Container {
 			MountPath: "/etc/triggers/tls",
 		})
 	} else {
-		*c.Port = 8080
 		scheme = corev1.URISchemeHTTP
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -555,6 +557,40 @@ func getContainer(el *v1alpha1.EventListener, c Config) corev1.Container {
 	}
 }
 
+func getPort(el *v1alpha1.EventListener, c Config) int {
```
MarcelMue, line 560: This whole function is not super nice / concise but it seems to me like the only way to safely determine which port to use.

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -555,6 +557,40 @@ func getContainer(el *v1alpha1.EventListener, c Config) corev1.Container {
...
+	}
+	if v, ok := certEnv["TLS_KEY"]; ok {
+		elKey = v.SecretKeyRef.Key
+	} else {
+		elKey = ""
+	}
+
+	if elCert != "" && elKey != "" {
+		// We always set port 8443 if TLS is enabled - regardless of the configured port.
+		return 8443
```
MarcelMue, line 588: This was discussed on slack to be the intended behaviour. Please let me know if you think this should be different.

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
```diff
@@ -1157,6 +1202,66 @@ func TestReconcile_Delete(t *testing.T) {
 	}
 }
 
+func Test_getPort(t *testing.T) {
```
MarcelMue, line 1205: Very short unit test - other `get` functions are not unit tested though.

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
```diff
@@ -1045,6 +1072,24 @@ func TestReconcile(t *testing.T) {
 			ConfigMaps:     []*corev1.ConfigMap{loggingConfigMap},
 		},
 	},
+		{
```
MarcelMue, line 1075: This test case verifies the intended behaviour initially stated here:
https://github.com/tektoncd/triggers/issues/318

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
dibyom: nice. the bit below about checking for cert Env seems repeated a couple of times...could we extract it to a function `isTLSEnabled` or something?

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
dibyom: hmmm..do we know why we can't/shouldn't change the port if TLS is enabled? @savitaashture

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -165,9 +167,9 @@ func (r *Reconciler) reconcileService(ctx context.Context, logger *zap.SugaredLo
 			Ports: []corev1.ServicePort{{
 				Name:     eventListenerServicePortName,
 				Protocol: corev1.ProtocolTCP,
-				Port:     int32(*r.config.Port),
+				Port:     int32(port),
```
dibyom, line 170: I'm wondering -- can we get away we just changing the port on the service (and leaving the TargetPort of the EL deployment as is?). That should simplify a lot of the logic. On the other hand, it will change a few things e.g. if someone is directly port-forwarding to the EL deployment.

Thoughts @MarcelMue @savitaashture @bigkevmcd ?

pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
dibyom: Nice, I think this is also related to #586

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
MarcelMue: I thought about this as well. Unfortunately we need to extract different information each time where this is done. I can try to refactor this later on but would prefer to do so in a separate PR.

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
MarcelMue: I agree that we could simplify this logic but can not tell what the intended setup is. Do we know of any users which would be impacted by such a change?

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
savitaashture: No specific reason but I see TLS connections usually enabled on `443` or `8443`
and for `443` port getting error related to permission bind
```
{"level":"fatal","ts":"2021-01-04T17:05:15.398Z","logger":"eventlistener","caller":"eventlistenersink/main.go:145","msg":"failed to start eventlistener sink: listen tcp :443: bind: permission denied","knative.dev/controller":"eventlistener","stacktrace":"main.main\n\tgithub.com/tektoncd/triggers/cmd/eventlistenersink/main.go:145\nruntime.main\n\truntime/proc.go:204"}
```
So just used `8443`

MarcelMue: I would then prefer to create a separate default for the TLS connection?
So users have the ability to overwrite it.

khrm (commented):
lgtm and can be merged after https://github.com/tektoncd/triggers/pull/881/files#r551844316

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
MarcelMue: I can change it but would prefer to do it in a follow up, if that's okay. I don't want to mix all these changes together :)

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
savitaashture: Sure i will send a PR to have default port for TLS

then we can also merge this https://github.com/tektoncd/triggers/pull/880 and i will add different PortName for TLS

savitaashture: @MarcelMue do you want to handle this as part of this PR itself ?
Or you want me to send separate PR

MarcelMue: I will take care in this PR. Change coming up in a few minutes :)

MarcelMue: @savitaashture Can you check? Changes are in now.

savitaashture (approved):

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
dibyom: ok, this looks fine for now. I'll add an item to discuss the separate TLS port behavior in the next working group.

PR #882, opened 2020-12-30
Refactor sink test

dibyom:
# Changes

There are two commits in this PR. It's pretty big but a lot of it is deleting duplicated code. The largest change is to take the many TestHandle* tests that all had the same setup steps and combine them into a single table driven test with multiple test cases.

There were a lost of TestHandle* tests with the same copy pasta making 
it hard to read/modify e.g. they all copied the same code to create
PipelineResource templates, used the same eventBody etc. 

1. combine most TestHandle into a table driven test
2. use v1beta1 TaskRuns instead of PipelineResources for all tests
3. add a func for creating test servers with core/webhook interceptors
4. refactor test.HMACHeader to take in testing.TB
5. use switch instead of if/else to reduce verbosity in TestHandle_OverrideAuth
6. use test.RawExtension helper for creating triggertemplates
 
# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/sink/sink_test.go
+++ b/pkg/sink/sink_test.go
@@ -29,25 +29,12 @@ import (
 	"sync"
 	"testing"
 
-	"knative.dev/pkg/ptr"
-
-	"github.com/tektoncd/triggers/pkg/interceptors/server"
-
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
-	"go.uber.org/zap/zaptest"
-	"go.uber.org/zap/zaptest/observer"
-	discoveryclient "k8s.io/client-go/discovery"
-	"k8s.io/client-go/dynamic"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
-
 	"github.com/gorilla/mux"
-	pipelinev1alpha1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
-	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	pipelinev1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
-	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	dynamicclientset "github.com/tektoncd/triggers/pkg/client/dynamic/clientset"
 	"github.com/tektoncd/triggers/pkg/client/dynamic/clientset/tekton"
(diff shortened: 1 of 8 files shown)


test/resources.go
```diff
@@ -45,10 +45,10 @@ func ToUnstructured(t *testing.T, in interface{}) *unstructured.Unstructured {
 // expected to be able to interact with.
 func AddTektonResources(clientset *fakekubeclientset.Clientset) {
 	nameKind := map[string]string{
-		"triggertemplates":  "TriggerTemplate",
 		"pipelineruns":      "PipelineRun",
 		"taskruns":          "TaskRun",
 		"pipelineresources": "PipelineResource",
+		"triggertemplates":  "TriggerTemplate",
```
savitaashture, line 51: any reason to change the position of these attributes

savitaashture (commented):
looks good to me :+1: 

minor question on attribute position change

test/resources.go
dibyom: I removed it thinking it was not necessary but then added it back after tests started failing. Will revert it.

savitaashture (commented):
/lgtm

pkg/sink/sink_test.go
```diff
@@ -141,25 +109,57 @@ func getSinkAssets(t *testing.T, resources test.Resources, elName string, auth A
...
-		prs = append(prs, pr)
+		trs = append(trs, tr)
 	}
-	return prs
+	return trs
 }
 
-// checkSinkResponse checks that the sink response status code is 201
-// and that the body returns the EventListener, namespace, and eventID.
+// checkSinkResponse checks that the sink response status code is 201 and that the body returns the EventListener, namespace, and eventID.
```
MarcelMue, line 162: Quick question: Why was this put into one line? Are there any coding guidelines for comments?

I usually see comments split consistently but here it was reverted.

pkg/sink/sink_test.go
```diff
@@ -179,687 +179,758 @@ func checkSinkResponse(t *testing.T, resp *http.Response, elName string) {
...
+		// resources are the K8s objects to setup the test env.
+		resources test.Resources
+		// webhookInterceptor is a handler that implements any webhook interceptors referenced in the triggers.
+		webhookInterceptor http.HandlerFunc
+		eventBody          []byte
+		headers            map[string][]string
+		// want is the resulting TaskRun(s) created by the EventListener
+		want []pipelinev1.TaskRun
+	}{{
+		name: "single trigger embedded within EventListener",
```
MarcelMue, line 330: Are there any conventions on test case naming? I generally like this style of naming but have noticed that it is inconsistent throughout the `triggers` project.

pkg/sink/sink_test.go
```diff
@@ -179,687 +179,758 @@ func checkSinkResponse(t *testing.T, resp *http.Response, elName string) {
...
+			req.Header.Add("Content-Type", "application/json")
+			resp, err := http.DefaultClient.Do(req)
+			if err != nil {
+				t.Fatalf("error sending request: %s", err)
+			}
+			checkSinkResponse(t, resp, elName)
+			// Check right resources were created.
+			got := toTaskRun(t, dynamicClient.Actions())
+			// Sort actions (we do not know what order they executed in)
+			if diff := cmp.Diff(tc.want, got, cmpopts.SortSlices(func(x, y pipelinev1.TaskRun) bool {
```
MarcelMue, line 697: Could this be split into two lines? Otherwise the comment is not really true as we sort and compare in a single line. It would make the code more readable / easier to follow IMO.

pkg/sink/sink_test.go
```diff
@@ -179,687 +179,758 @@ func checkSinkResponse(t *testing.T, resp *http.Response, elName string) {
...
-	secret := &corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "secret",
-			Namespace: namespace,
-		},
-		Data: map[string][]byte{
-			"secretKey": []byte("secret"),
-		},
-	}
+// Setup for TestHandleEvent_AuthOverride
```
MarcelMue, line 706: Why are these declared in the middle of the file as compared to either at the top or in inside the test func?

MarcelMue (commented):
Just some comments as I try to get into the project more.

pkg/sink/sink_test.go
dibyom: yeah this looks inadvertent. will revert.

pkg/sink/sink_test.go
dibyom: no hard and fast rule. Ideally the name would be both descriptive and short but if not possible, I prefer a more descriptive but longer name over a shorter one (though you are right we haven't been very consistent about this).

pkg/sink/sink_test.go
dibyom: sure!

pkg/sink/sink_test.go
dibyom: We can't put it inside the TestHandleEvent_AuthOverride since these variables are used in the OverrideAuthentication method that we implement for the `fakeAuth` type below. We could put these at the top of the file too but given that these are only used for the test below, we declare them close to where they are being used.

pkg/sink/sink_test.go
dibyom: Updated the comment -- we don't technically sort the taskRuns ourselves, `cmp.Diff` has an option to sortSlices, we just pass it a function that it can use to compare the TaskRuns when sorting.

khrm (commented):
/lgtm

Issue #868, opened 2020-12-16
Interceptors: Implement CRD to provide a single way to install Interceptors


dibyom:
The work for this will involve implementing the InterceptorConfig CRD as described in [TEP-0026](https://github.com/tektoncd/community/blob/master/teps/0026-interceptor-plugins.md#installing-interceptors)

PR #891, opened 2021-01-08
Use system.DefaultNamespace for interceptor urls

dibyom:
# Changes

SYSTEM_NAMESPACE is set to the namespace of the EventListener by the reconciler
which means that with the current logic the interceptor url is wrong. Use
default namespace to always use `tekton-pipelines` as the namespace. This is
hardcoded and the logic is temporary till we implement #868.

/kind bug

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/cmd/interceptors/main.go
+++ b/cmd/interceptors/main.go
@@ -81,6 +81,7 @@ func main() {
 		ReadTimeout:  readTimeout,
 		WriteTimeout: writeTimeout,
 		IdleTimeout:  idleTimeout,
+		Handler:      mux,
 	}
 
 	logger.Infof("Listen and serve on port %d", Port)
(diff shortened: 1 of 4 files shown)


savitaashture (approved):

wlynch (approved):
/lgtm

MarcelMue:
@dibyom Can you explain why this was necessary? This currently completely breaks the usecase of running in a different namespace because it will _always_ resolve the service name into a wrong namespace.

dibyom:
> Can you explain why this was necessary? This currently completely breaks the usecase of running in a different namespace because it will always resolve the service name into a wrong namespace.

I thought SYSTEM_NAMESPACE was the namespace Triggers was installed in but it turns out we (also) use it for the namespace that the Event Listener was installed in :(

PR #898, opened 2021-01-13
Remove the template.Name field

MarcelMue:
# Changes

Simple PR which removes the already deprecated field.

Closes: #791 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#tests) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#docs) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commits)
- [x] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/triggers/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

```release-note
Remove the Name field from the trigger spec. Use Ref instead.
```

PR #912, opened 2021-01-19
Revert removal of template.Name field and handle Deprecation of template.Name field

savitaashture:
**Issue:**  [slack discussion](https://tektoncd.slack.com/archives/CL3T51NRF/p1610982177013500) 
**RootCause:**  https://tektoncd.slack.com/archives/CL3T51NRF/p1611039325020300?thread_ts=1610982177.013500&cid=CL3T51NRF

# Changes
Reverted back changes done as part of this [PR](https://github.com/tektoncd/triggers/pull/898) and handled Deprecation for `template.Name` field
Will remove `template.Name` field may be in the next release `v0.12.0`

We may need to do patch release to handle this issue

Signed-off-by: Savita Ashture sashture@redhat.com

# Submitter Checklist

review them:

# Release Notes

```release-note
Deprecate template.Name in favor of template.Ref
```


pkg/apis/triggers/v1alpha1/trigger_types.go
```diff
@@ -39,9 +39,12 @@ type TriggerSpec struct {
 }
 
 type TriggerSpecTemplate struct {
-	Ref        *string              `json:"ref,omitempty"`
-	APIVersion string               `json:"apiversion,omitempty"`
-	Spec       *TriggerTemplateSpec `json:"spec,omitempty"`
+	// Deprecated: Use Ref instead
+	// To be removed in a later release #911
+	DeprecatedName string               `json:"name,omitempty"`
```
MarcelMue, line 44: Not blocking just wondering:
Why does the `omitempty` fix the problem described in slack?

khrm (approved):
/lgtm

pkg/apis/triggers/v1alpha1/trigger_types.go
khrm: Because omitempty ensures that when this struct is store, it stores nothing related to DeprecatedName field but Ref using defaults. Previously, default wasn't working properly because it stored "".

pkg/apis/triggers/v1alpha1/event_listener_defaults.go
```diff
@@ -30,6 +30,7 @@ func (el *EventListener) SetDefaults(ctx context.Context) {
 		for i := range el.Spec.Triggers {
 			triggerSpecBindingArray(el.Spec.Triggers[i].Bindings).
 				defaultBindings()
+			templateNameToRef(el.Spec.Triggers[i].Template)
```
khrm, line 33: Call this function only if there's Template isn't nil. I think that's why test is failing.

savitaashture: Yes its failing because of that

Updating

pkg/apis/triggers/v1alpha1/trigger_types.go
MarcelMue: What I mean is: We have removed the name field completely already, creating a new CR should therefore not have the field anymore. So even if it was stored with "" we should fix anything going wrong in that case.

Readding the field here kind of implies that users first need to upgrade to this specific version before updating to the next minor.

khrm: I think we can recommend the user to directly upgrade to v0.11.0 from v0.10.2

MarcelMue: So then it would make sense to put this change into `v0.10.3` instead of `v0.11.1`, right?

dibyom:
Summarizing the issue here:

We had logic in triggers_default.go to change template.Name to template.Ref. But because the Name field did not have a omitempty tag, the resource still contained a `Name: ""` field. This makes it impossible to upgrade running trigger resources from v0.10.2 to v0.11.0

pkg/apis/triggers/v1alpha1/trigger_types.go
dibyom: yeah doing a 0.10.3 is also an option. Though that would mean users would first have to move to v0.10.3 and then to v0.11.0/v0.11.1

dibyom:
> EasyCLA Expected — Waiting for status to be reported

🙄

MarcelMue:
Fine with me - I think it's the easiest way forward for users.

/lgtm

dibyom:
Gonna try to close/re-open this PR for trigger the CLA again

savitaashture:
> Gonna try to close/re-open this PR for trigger the CLA again

I have done force push

PR #921, opened 2021-01-22
Add a CRD for Interceptors

dibyom:
# Changes

Broken down into 2 commits for easier review:
1. Add Interceptor CRD type 
The `spec` currently only contains a URL field to locate where the interceptor
is running. Other fields will be added as they are implemented in follow ups.

1. Migrate core interceptors to use Interceptor CRD
    1. Add YAML definitions for each core interceptor using the Interceptor CRD.
    2. Modify the EventListener to use the Interceptor CRD to find the Interceptor's URL.
    3. Update roles/RBAC/unit tests for the above 2 changes.

1. Fix core interceptors installation errors 
We need to wait for the Interceptors CRD to become established before we can
create the core interceptors. Otherwise, we run the risk of trying to create
the core-interceptors too early leading to installation errors.

# Submitter Checklist

review them:

# Release Notes

```release-note
Adds a new CRD type called Interceptors for installing pluggable interceptors. Moves current set of core-inerceptors to be installed using the new CRD.
```


config/interceptors/core-interceptors.yaml
```diff
@@ -0,0 +1,70 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: Interceptor
+metadata:
+  name: cel
+spec:
+  clientConfig:
+    url: "http://tekton-triggers-core-interceptors.tekton-pipelines.svc/cel"
```
wlynch, line 21: Letting the users set the path definitely seems useful, but it's unclear to me how users would know what Service URL this would be associated with, particularly if there are multiple interceptor services at play.

Few ideas:
1) Is the full URL something we can expose in the status and let the Interceptor controller handle filling this in for clients?
2) Should these CRDs be produced by Interceptor providers instead (e.g. on startup/control loop)? This has the added benefit of not requiring users to apply the interceptors themselves, sidestepping the apply ordering issues.

dibyom: end users do not have to know what the URL is. Only the interceptor provider. I implemented `url` first since that's the simplest but we will [support a ref to an object too ](https://github.com/tektoncd/community/blob/master/teps/0026-interceptor-plugins.md#installing-interceptors). 

> Is the full URL something we can expose in the status and let the Interceptor controller handle filling this in for clients?

Yes, was planning on adding this as part of the followup that adds service ref support. That being said, it might still make sense to have the same URL resolution happen at the EL level in case the EL receives an event before the Interceptor controller does a reconcile loop.

> Should these CRDs be produced by Interceptor providers instead (e.g. on startup/control loop)? This has the added benefit of not requiring users to apply the interceptors themselves, sidestepping the apply ordering issues.

Interesting idea. Maybe? I think it will solve the issue only for the core interceptors that are bundled with the Triggers installation. Otherwise, it will be up to the interceptor provider to ship both the deployment running the logic and the CRD to go with it. They could decide to ship a deployment that creates the CRD on startup (but then that deployment will need to have RBAC roles to create Interceptors)

wlynch: > but then that deployment will need to have RBAC roles to create Interceptors

mmmmm good point. Perhaps this is better then.

wlynch (commented):
/lgtm

Holding off on approval - I think it'd be useful to get a 2nd pair of eyes on this as a sanity check to validate the "user" (interceptor author?) facing API.

pkg/interceptors/interceptors.go
```diff
@@ -171,28 +171,40 @@ func UnmarshalParams(ip map[string]interface{}, p interface{}) error {
...
+		name = "gitlab"
 	}
-	return &url.URL{
-		Scheme: "http",
-		Host:   fmt.Sprintf("%s.%s.svc", CoreInterceptorsHost, os.Getenv("TEKTON_INSTALL_NAMESPACE")),
-		Path:   path,
+	return name
+}
+
+// Resolve resolves the given CEL name to the
```
bigkevmcd, line 191: This comment doesn't seem quite right?

pkg/interceptors/interceptors.go
```diff
@@ -171,28 +171,40 @@ func UnmarshalParams(ip map[string]interface{}, p interface{}) error {
...
+type InterceptorGetter func(name string) (*triggersv1.Interceptor, error)
+
+var ErrNoURL = errors.New("interceptor URL was not found")
```
bigkevmcd, line 194: I wonder if this shouldn't be an error type, so that we can have the URL that was not found in the message to help with debugging?

pkg/interceptors/interceptors_test.go
```diff
@@ -305,48 +306,76 @@ func TestResolvePath(t *testing.T) {
...
+			}, nil
+		}
+
+		got, err := interceptors.ResolveToURL(fakeGetter, "cel")
+		if err != nil {
+			t.Fatalf("ResolveToURL() error: %s", err)
+		}
+		want := "http://some-host/cel"
+		if got.String() != want {
+			t.Fatalf("ResolveToURL want: %s; got: %s", want, got.String())
```
bigkevmcd, line 362: nitpick here, `%s` will trigger calling `.String()` on values.

bigkevmcd (commented):
I think we should improve that error message in the event of not found.

pkg/interceptors/interceptors.go
dibyom: oh we return this when the URL field is nil...so we literally have nothing to return. I can see this being confusing. I'll change the error message to URL was nil or something like that

pkg/interceptors/interceptors.go
```diff
@@ -171,28 +171,40 @@ func UnmarshalParams(ip map[string]interface{}, p interface{}) error {
...
+var ErrNilURL = errors.New("interceptor URL was nil")
+
+// ResolveToURL finds an Interceptor's URL.
+func ResolveToURL(getter InterceptorGetter, name string) (*apis.URL, error) {
+	ic, err := getter(name)
+	if err != nil {
+		return nil, err
+	}
+	url := ic.Spec.ClientConfig.URL
+	if url == nil {
```
savitaashture, line 202:
```suggestion
	if url := ic.Spec.ClientConfig.URL; url == nil {
```

docs/interceptors.md
```diff
@@ -0,0 +1,51 @@
...
+processing of a trigger to modify the behavior or payload of Triggers. The
+custom resource describes how an EventListener can connect to a workload that
+is running the interceptor business logic (and in the future what extra
+paramters the interceptor accepts).
+
+**NOTE**: This doc is a WIP. Please also see the [Interceptors section](./eventlisteners.md#interceptors) in the EventListener doc.
+
+- [Interceptors](#interceptors)
+  - [Syntax](#syntax)
+  -- [clientConfig](#clientConfig)
```
savitaashture, line 19:
```suggestion
- [Interceptors](#interceptors)
  - [Syntax](#syntax)
    - [clientConfig](#clientConfig)
```

config/interceptors/core-interceptors.yaml
```diff
@@ -0,0 +1,70 @@
+# Copyright 2020 The Tekton Authors
```
savitaashture, line 1:
```suggestion
# Copyright 2021 The Tekton Authors
```

config/300-interceptor.yaml
```diff
@@ -0,0 +1,54 @@
+# Copyright 2019 The Tekton Authors
```
savitaashture, line 1:
```suggestion
# Copyright 2021 The Tekton Authors
```

savitaashture (commented):
few nits

savitaashture:
@dibyom 

`ko apply -f config/` won't read `interceptors` folder
So we may 
* need to update [Install Triggers](https://github.com/tektoncd/triggers/blob/master/DEVELOPMENT.md#install-triggers) part of DEVELOPMENT.md

OR

We can create a symlink  for the files present in `interceptors` folder

config/300-interceptor.yaml
```diff
@@ -0,0 +1,54 @@
...
+    version: "devel"
+spec:
+  group: triggers.tekton.dev
+  scope: Cluster
+  names:
+    kind: Interceptor
+    plural: interceptors
+    singular: interceptor
+    shortNames:
+      - i
```
MarcelMue, line 32: I personally don't think this shortname is a good idea. I would prefer having none than a 1 character one.

config/interceptors/core-interceptors.yaml
```diff
@@ -0,0 +1,70 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: Interceptor
+metadata:
+  name: cel
```
MarcelMue, line 18: Putting them in namespace `tekton-pipelines` would make sense to me. This is where basically everything else is, right? Especially since they are shipped together.

config/interceptors/core-interceptors.yaml
```diff
@@ -0,0 +1,70 @@
...
+spec:
+  clientConfig:
```
MarcelMue, line 20: WDYT about omitting `Config` from the name and simply naming the field `client`?
Every CR is config to an extend and it's clear that this simply defines the client which handles the interceptor.

pkg/apis/triggers/v1alpha1/interceptor_types.go
```diff
@@ -6,8 +6,43 @@ import (
...
+// Interceptor describes a interceptor's configuration such as the fields it accepts
+// and its deployment address
+type Interceptor struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec InterceptorSpec `json:"spec"`
+}
+
```
MarcelMue, line 26: Godoc should be added.

MarcelMue (commented):
Some minor comments.

config/interceptors/core-interceptors.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - name: "overlays"
+      description: "CEL expressions whose values get mapped to user defined keys"
+---
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: Interceptor
+metadata:
+  name: bitbucket
+spec:
+  clientConfig:
+    url: "http://tekton-triggers-core-interceptors.tekton-pipelines.svc/bitbucket"
```
savitaashture, line 35: just thinking at high level

These interceptors are core interceptors which will be installed as part of triggers installation what if Tekton installs in different namespace other than `tekton-pipelines` :thinking: 

In that case will this be responsibility of user/admin who installs Triggers ?

dibyom: yeah this is a bit confusing right now. The idea is instead of url we'd also support a `service` field that contains a reference to the service for the deployed core interceptors. Custom installs would have to change the namespace of the service but would not have to modify a URL.

(I was thinking of adding that as a separate PR but maybe I'll add it to this one itself to make it less confusing).

pkg/interceptors/interceptors.go
dibyom: kept this but fixed the return statement to return `url` which is why I orginally extracted this to a variable.

config/300-interceptor.yaml
dibyom: fair enough. I am also thinking of renaming the actual type from Interceptors to InterceptorType to make it clearer that this CR defines a new "type" or kind of interceptor that can then be used by a end user.

config/interceptors/core-interceptors.yaml
dibyom: Right now this is a cluster wide type i.e. you install the InterceptorType GitHub for your entire cluster. Similar to how you install pipelines for the entire cluster.

config/interceptors/core-interceptors.yaml
dibyom: I was basing this off of the webhook client config type that k8s webhook types : https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#webhookclientconfig-v1-admissionregistration-k8s-io

On second thought, maybe I can just use that type itself instead of defining my own type for this.

dibyom:
/close

Closing in favor of #960 and #976

Issue #901, opened 2021-01-14
Proposal: Use UUID for Trigger Event IDs


wlynch:
Short proposal I wanted to get feedback on before making a change.

Currently we generate Event IDs similar to `generateName` by [randomly generating a 5 character string](https://github.com/tektoncd/triggers/blob/9bc264f2cefd7b7ade512841ba4ae04c81da582c/pkg/template/resource.go#L143) to associate to the event. While this has worked so far, we might find that the current 5 character solution isn't sufficient in the future, particularly if/when we start recording events that don't result in runs.

Instead of slowly growing the Event ID over time, I think we should switch over to [UUIDs](https://tools.ietf.org/html/rfc4122) for Event ID generation. This is how k8s objects generate UIDs, and would be a fairly straight forward change for us since we treat these IDs as opaque strings today.

Let me know if anyone has any objections!

dibyom:
Sounds reasonable to me!

MarcelMue:
Sounds reasonable - some notes:
- Is the current ID used in any length limited fields (e.g. name)? Following RFC4122 we would have 36 characters instead of 5
- Do we expect the hyphens in the UUID to have any impact? It shouldn't in k8s but I am not sure where this ID is currently used
- I would like to introduce more debug logging in `triggers`, logging the full UUID might be very spammy. There are multiple solutions to shorten UUIDs for that purpose. We should at least think about it.

afrittoli:
We used to have `$(uid)` in resource names in trigger templates - we still do in the getting started, which with the new uid is not possible, as the resource name becomes too long:
```
[...]

PR #926, opened 2021-01-25
Switch to UUID for event IDs

MarcelMue:
# Changes
Switches the 5 character random string UIDs to RFC 4122 UUIDs.

Closes: #901

# Submitter Checklist

review them:

# Release Notes

```release-note
Change the event ID representation from a 5 character random string to a UUID. 
```


pkg/template/resource.go
```diff
@@ -136,13 +135,13 @@ func applyParamToResourceTemplate(param triggersv1.Param, rt json.RawMessage, ol
 	return bytes.Replace(rt, []byte(paramVariable), []byte(param.Value), -1)
 }
 
-// UID generates a random string like the Kubernetes apiserver generateName metafield postfix.
-var UID = func() string { return rand.String(5) }
+// UUID generates a Universally Unique IDentifier following RFC 4122.
+var UUID = func() string { return uuid.New().String() }
```
MarcelMue, line 139: I have currently changed the naming everywhere from `uid` to `uuid` - I am not certain if this is really required.

`uuid` is more clear as to what it is.
`uid` is a less breaking change (and is the name k8s uses AFAIK).

MarcelMue: cc @wlynch in case you have suggestions.

wlynch: I agree - I'm leaning towards leaving things as `uid` based on your reasoning + so we don't have to bother with deprecation.

wlynch (commented):
Thanks for picking this up!

pkg/template/resource.go
MarcelMue: This is done now - the change is not only very internal, renaming only the internal functions properly.

MarcelMue:
I am not aware of any docs which should be changed. None that I could find mention the ID structure.

wlynch (approved):
/lgtm

PR #946, opened 2021-02-05
Allow Nested triggers to be invoked after interceptor filter chain completes

### LINKED ISSUE #945: Feature: Allow Triggers to invoke other triggers after interceptor processing
### Feature request

Screen Shot 2021-02-05 at 11 06 40 AM

Triggers processed in the sink right now can currently only resolve to a single template and set of bindings. As outlined in #820, there is often a goal to do some initial interceptor processing before invoking additional triggers. In this implementation, a `Triggers` field would be added to the current `TriggerSpec` that would be a list of refs to Trigger objects accessible to the EventListener sink. After interceptor processing, the sub-triggers would be invoked alongside any existing binding/template pair set on the trigger.

Some changes that would have to happen to make this work:

* Template/Bindings would become optional in the Trigger CRD - update the openapi spec for Triggers to force one-of either template/bindings or triggers.
* The sink Trigger resolution function would have to distinguish between "exposed" triggers and "internal" triggers (Propose using a label for this)
* Need to prevent infinite loop of triggers somehow
* Determine what triggers are available for "internal" processing. In my mind, all triggers would be available to use as a trigger ref, we are just filtering some of them from the sink.
* Determine if a Trigger can have both a template/binding and trigger refs.
* Determine how the event listener will respond to these requests. Should it wait for all the "sub"-triggers to be invoked to determine if it should response 201 or 202? For our use case its not essential, but I may not be aware of other use cases that take advantage of that status code.
 
### Use case

The use case is fairly well outlined in #820 as well. My team's use case is that we are passing additional metadata along with each repository and having a webhook add that metadata via a webhook extension. We want to run that webhook once for each invocation of the eventlistener, then separate processing into a set of subsequent triggers after that initial extension is processed.

jmcshane:
# Changes

Closes #945

In this PR:
- [x] proposing simple API spec to get started here
- [x] Update logic of sink to exclude Triggers from sink processing
- [x] Branch logic after interceptors to be able to invoke triggers
- [ ] Update eventlistener response to check if a pipeline was triggered through nesting
- [x] protect eventlistener from infinite nesting

# Submitter Checklist

review them:

# Release Notes

```release-note
Enable sink triggers to invoke downstream triggers to branch validation logic from pipeline logic
```


jmcshane:
Will squash after implementation

jmcshane:
I created a simple test in a gist if you are interested in testing on your local installation: https://gist.github.com/jmcshane/fb810c6b13da53c6de168b5a8c9a3294

If you download the trigger.yaml and hack.sh script to the same directory, you can run `sh hack.sh` and view the full demo

jmcshane:
Closing in favor of #1025

PR #949, opened 2021-02-07
Resolve TriggerTemplate and TriggerBinding Ref

### LINKED ISSUE #948: Feature: Allow triggerbinding and triggertemplate refs to be populated by tekton expressions
### Feature request

Currently, triggerbinding and triggertemplate references must be hardcoded strings. In this feature requests, these strings can be populated by an expression, possibly adding a new field in the `TriggerSpecBinding` and `TriggerSpecTemplate` spec besides `Ref` to account for this difference in how the values would be populated.

### Use case

This would allow Trigger objects to be more flexible to support a range of options based on the inputs. A couple small use cases:

- Github PR event has Repository owner in the body. This field could be used to select pipelines or binding metadata that are specific to that set of repositories. For example, a triggerbinding that associates the GH repository owner to information about that owner stored in a set of values on that binding.
- If trigger processing can determine whether or not a binding needs a workspace with a new volume claim, existing volume claim, or no workspace (via an extension), the Trigger processing could select the TriggerTemplate that has each one of these.

jmcshane:
This PR demonstrates a capability to pre-process TriggerTemplate and TriggerBinding
objects with the event context to allow the Ref field to include tekton  expressions
in order to allow body and header context to be populated into these names.

# Changes

Closes #948

# Submitter Checklist

review them:

Tests and documentation in progress.

# Release Notes

```release-note 
TriggerTemplate and TriggerBindings can now use Tekton variable substitution to resolve resource names
```

### DIFF
--- a/docs/triggers.md
+++ b/docs/triggers.md
@@ -49,6 +49,31 @@ spec:
     ref: pipeline-template
 ```
 
+Triggers can reference templates and bindings based on data in the event body, headers, or processed extensions using the `dynamicRef` field. This allows contextual information to be pulled in from the event to determine which binding or template should be populated. The `dynamicRef` field for `bindings` and `templates` can be populated in the same manner as triggertemplate params.
+
+This can be useful when different pipelines require different resources or workspaces to run. `TriggerTemplate` objects by themselves must have a static configuration structure, even if individual strings are able to be populated.
+
+These dynamic resolutions also allow for a fallback option if the dynamic resolution does not succeed. If the `dynamicRef` fields are not present in the event body, the Trigger processing will use the `ref` field if populated.
+
+<!-- FILE: examples/triggers/trigger-with-event-reference.yaml -->
+```YAML
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: Trigger
+metadata:
+  name: trigger
+spec:
+  interceptors:
+    - cel:
+        overlays:
+        - key: extensions.truncated_sha
+          expression: "body.pull_request.head.sha.truncate(7)"
+  bindings:
+  - dynamicRef: pipeline-$(body.repository.owner.login)-binding
+  template:
+    dynamicRef: pipeline-$(header.X-GitHub-Event)-trigger
+    ref: pipeline-default-trigger
+```
+
 ## Specifying the corresponding `TriggerTemplate`
 
 In the `template` field,  you can do one of the following:
--- a/cmd/triggerrun/cmd/root.go
+++ b/cmd/triggerrun/cmd/root.go
@@ -203,6 +203,7 @@ func processTriggerSpec(kubeClient kubernetes.Interface, client triggersclientse
 	if iresp != nil {
 		if !iresp.Continue {
 			log.Errorf("interceptor stoppped trigger processing: %w", iresp.Status.Err())
+			return nil, iresp.Status.Err()
 		}
 	}
 
@@ -210,7 +211,12 @@ func processTriggerSpec(kubeClient kubernetes.Interface, client triggersclientse
 		tri.Namespace = "default"
 	}
 
-	rt, err := template.ResolveTrigger(*tri,
+	extensions := map[string]interface{}{}
+	if iresp != nil && iresp.Extensions != nil {
+		extensions = iresp.Extensions
+	}
+
+	resources, err := template.Resolve(*tri,
 		func(name string) (*triggersv1.TriggerBinding, error) {
 			return client.TriggersV1alpha1().TriggerBindings(tri.Namespace).Get(context.Background(), name, metav1.GetOptions{})
 		},
@@ -219,23 +225,13 @@ func processTriggerSpec(kubeClient kubernetes.Interface, client triggersclientse
 		},
 		func(name string) (*triggersv1.TriggerTemplate, error) {
 			return client.TriggersV1alpha1().TriggerTemplates(tri.Namespace).Get(context.Background(), name, metav1.GetOptions{})
-		})
-	if err != nil {
-		log.Error("Failed to resolve Trigger: ", err)
-		return nil, err
-	}
-	extensions := map[string]interface{}{}
-	if iresp != nil && iresp.Extensions != nil {
-		extensions = iresp.Extensions
-	}
-	params, err := template.ResolveParams(rt, finalPayload, header, extensions)
-	if err != nil {
-		log.Error("Failed to resolve parameters", err)
-		return nil, err
-	}
-	log.Infof("ResolvedParams : %+v", params)
+		},
+		finalPayload, header, extensions)
 
-	resources := template.ResolveResources(rt.TriggerTemplate, params)
+	if err != nil {
+		log.Error("Failed to resolve trigger resources", err)
+		return nil, err
+	}
 
 	return resources, nil
 }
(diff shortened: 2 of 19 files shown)


jmcshane:
Created a gist if anyone wants to install this in their cluster. Ensure that a functional pipelines/triggers installation with the latest version is already present before running `./hack.sh`:

https://gist.github.com/jmcshane/1473c39b61137fa14a508ecb9643ce6e

jmcshane:
This PR has an issue right now in that two triggers with the same name that are invoked will end up modifying the same pointer. I have it fixed in another branch, will push a fix here in the morning.

jmcshane:
Yeah, this definitely needs some more testing. If the maintainers are interested in merging this, I can get tests and docs updated tomorrow.

docs/triggers.md
```diff
@@ -51,6 +51,28 @@ spec:
...
+metadata:
+  name: trigger
+spec:
+  interceptors:
+    - cel:
+        overlays:
+        - key: extensions.truncated_sha
+          expression: "body.pull_request.head.sha.truncate(7)"
+  bindings:
+  - ref: pipeline-$(body.repository.owner.login)-binding
```
dibyom, line 71: what happens if the expression does not resolve? (Or it resolves to some other binding that was unintended.

docs/triggers.md
```diff
@@ -51,6 +51,28 @@ spec:
...
+  template:
+    ref: pipeline-$(header.X-GitHub-Event)-trigger
```
dibyom, line 73: So someone will still have to create a bunch of templates and bindings with valid values of `X-Github-Event` and `repository.owner.login` ?

Assuming we have a solution the reusable interceptor processing problem, one potential alternative could be having all the interceptors in a chain and then each trigger has a simple CEL filter plus a static binding and template value. 

It is a bit more verbose but IMO makes things the trigger config a bit easier to parse. What do you think?

docs/triggers.md
jmcshane: I'm not sure what `unintended` means in this context. 

Failure scenarios operate in much the same way as other tekton expressions within a Trigger:

1. If the expression is not found, processing is stopped with an error message saying that the key could not be found in the event
2. If the triggerbinding/triggertemplate is not found, processing is stopped as it retrieves a 404 from the getTB/getCTB/getTT function in the same was as if a non-existent triggertemplate was referenced.

docs/triggers.md
jmcshane: I think that this formatting still makes sense when you want flexibility based off extension values. Remember one of the key things we are trying to enable here is workspace configuration that differs based off metadata on the input. I think that static binding and template still have the issue specifically around static triggertemplates.

Bindings have more flexibility for population of a range of values since triggerbindings can have more parameters than the template itself. What I have found using this capability is that I have to worry far less about defaults because I specify exactly the set of extensions I am looking for in a triggerbinding. Now, a goal here is not performance improvement, but rather ease the operation and management of a tekton platform.

The `resolve` operations still operate in the same manner. Triggerbindings and templates resolve within the namespace of the Trigger that invoke them.

I can see a concern around the platform management use case: "What will this trigger do when I invoke it" is an important question. I think improving the story about logging and debugging (as @bigkevmcd has in #954) is the biggest gain here. Personally, I find these configurations still readable, as they parse similarly to TriggerBinding values when I read through them.

jmcshane:
@dibyom I was thinking about a name for a possible new API field, it could be either `dynamicRef` or `bindingRef` to outline that the operation is similar to ref, but it operates based on binding/dynamic values.

docs/triggers.md
gabemontero: would explicitly specifying a default(s) as additional field(s) in our new API if 1) and/or 2) occur provide any value?

apologies, but I don't remember from the conversation I was a part of at the community meeting a couple of weeks ago got into any order of precedence / defaulting wrt the existing API in this space

jmcshane: one way defaulting could work is by setting both a `ref` and `dynamicRef` fields. During processing, the dynamic reference could take first priority, with the `ref` being a fallback if the dynamic processing fails.

That might be challenging from a readability perspective, so maybe an additional field would be better for specifying this default value when dynamic resolution fails.

gabemontero: Yeah might need some order of precedence defined for the `ref` and `dynamicRef` regardless, unless we updated the client side validation and/or add a webhook to prevent both being set.  But minimally some updates to the markdown docs for these api explaining what it would be.

But I tend to side on your thought on the additional field being better.  Plus IIRC there is some precedence in other portion(s) of the Trigger API for default values when resolving stuff of this ilk.

jmcshane: @gabemontero kind of dropped off this conversation thread. I think `dynamicRef` takes precedence over `ref` if present. I'll update this PR to use the additional field tomorrow and you can let me know what you think about this PR after the changes

gabemontero: sounds good @jmcshane

docs/triggers.md
```diff
@@ -51,6 +51,28 @@ spec:
     ref: pipeline-template
 ```
 
+Triggers can reference templates and bindings based on data in the event body, headers, or processed extensions. This allows contextual information to be pulled in from the event to determine which binding or template should be populated. The `ref` field for `bindings` and `templates` can be populated in the same manner as triggertemplate params.
+
+This can be useful when different pipelines require different resources or workspaces to run. `TriggerTemplate` objects by themselves must have a static configuration structurually, even if individual strings are able to be populated.
```
mpetason, line 56:
```suggestion
This can be useful when different pipelines require different resources or workspaces to run. `TriggerTemplate` objects by themselves must have a static configuration structurally, even if individual strings are able to be populated.
```
Maybe just structure?

jmcshane: updated to `structure` in the latest version

jmcshane:
@gabemontero @dibyom I have updated this PR using the `dynamicRef` field. The process (as documented in docs/triggers.md) is that it will try to resolve the `dynamicRef` for both the binding and template and if the resolution fails, it will fallback to a `ref` if that is set. This does allow for nice default behavior.

One situation that is not handled right now would be the following: `dynamicRef` resolves to a value, but the Template/Binding is not found. In the current code, it would not then try the `ref` because the processing of `dynamicRef` with the event is done earlier.

Reading through my own comment, I think I need to support falling back to `ref` in that situation. I'll get this code updated

jmcshane:
The logic is now updated to fall back to the `Ref` as a default in all scenarios. the documentation should reflect this as well.

As demoed in today's working group, this really assists in both fan-in and fan-out of nested triggers. by adding extension fields from external webhooks, most of our examples of using this capability come from using `extensions` fields as dynamic ref values.

I think it may make sense to add a trigger like this to the end to end tests, but I'd love to get some feedback before proceeding here.

marcusmaday:
This makes good sense to me and I'm absolutely in favor of having this implemented as dynamicRef with a failover to ref (if ref exists).  This would solve a huge problem for us as it reduces a lot of the yaml sprawl that we get when managing our Templates/Bindings, especially when combined with Trigger nesting.

gabemontero:
my previous input has been addressed @jmcshane 

given what I've got going on currently, I won't be able to make a more detailed pass through the code

I'll defer to @dibyom and others who are more engaged currently ... I will do the official prow

/assign @dibyom 

for you :-) ... he can unassign or assign to others as needed

dibyom:
Sorry for the delay @jmcshane I'm going to go over this today.

PR #960, opened 2021-02-17
Add ClusterInterceptor CRD for registering interceptors

dibyom:
# Changes

This commit adds a new CRD type called ClusterInterceptor. In TEP-0026 this
type was called InterceptorConfig 

The `spec` currently only contains a clientConfig field to locate where
the interceptoris running. Other fields will be added as they are
implemented in follow ups.

This commit also adds a simple reconciler for this type that resolves
the clientConfig to a URL and adds it to the `status.address.url` field.

Part of #868

# Submitter Checklist

review them:

# Release Notes

```release-note
A new CRD called ClusterInterceptor has been added that allows for users to register new Interceptor types.
```


dibyom:
/cc @savitaashture @wlynch @MarcelMue @bigkevmcd  

I broke off https://github.com/tektoncd/triggers/pull/921 into multiple PRs. One change I did was to call the CRD InterceptorType instead of just Interceptor to indicate that it defines a new type or kind of interceptor.

docs/interceptortypes.md
```diff
@@ -0,0 +1,56 @@
+<!--
+---
+linkTitle: "InterceptorType"
+weight: 9
+---
+-->
+# InterceptorType
+
+A `InterceptorType` is cluster scoped resource that registers a new Interceptor that
```
wlynch, line 9: I'm not 100% sure we want this to be cluster scoped. In particular, I'm thinking about how we can expose custom interceptors that can only be used within a single tenant namespace.

pkg/apis/triggers/v1alpha1/interceptor_type_types.go
```diff
@@ -0,0 +1,103 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec InterceptorTypeSpec `json:"spec"`
+	// +optional
+	Status InterceptorTypeStatus `json:"status"`
+}
+
+// InterceptorTypeSpec describes the Spec for an InterceptorType
+type InterceptorTypeSpec struct {
+	ClientConfig ClientConfig `json:"clientConfig"`
```
wlynch, line 35: Should we reuse [WebhookClientConfig](https://github.com/kubernetes/apiextensions-apiserver/blob/b343cbeb89f1b1f00514360d1e67397207b2e7be/pkg/apis/apiextensions/types.go#L107)? 

At the very least we can likely reuse the ServiceReference if we want to use the apis.URL for URL. The cert data is also useful for HTTPS.

docs/interceptortypes.md
```diff
@@ -0,0 +1,56 @@
...
+
+**NOTE**: This doc is a WIP. Please also see the [Interceptors section](./eventlisteners.md#interceptors) in the EventListener doc.
+
+- [Interceptors](#interceptors)
+  - [Syntax](#syntax)
+    - [clientConfig](#clientConfig)
+
+## Syntax
+
+To define a configuration file for an `Interceptor` resource, you can specify
```
wlynch, line 23: bikeshed: I kind of like InterceptorConfiguration to be consistent with [ValidatingWebhookConfiguration](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/).

Other good ideas to borrow from admission webhooks we should look into later:
- Version info (e.g. have different routes based on interceptor version)
- Timeout

docs/interceptortypes.md
```diff
@@ -0,0 +1,56 @@
...
+EventListeners will send forward requests to this service or URL.
+
+```yaml
+spec:
+  clientConfig:
+    url: "http://interceptor-svc.default.svc/"
+---
+spec:
+  clientConfig:
+    name: "my-interceptor-svc"
```
wlynch, line 51: This should be nested in `service`. We should also make note of the optional fields.

pkg/apis/triggers/v1alpha1/interceptor_type_defaults.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+)
+
+// SetDefaults sets the defaults on the object.
+func (it *InterceptorType) SetDefaults(ctx context.Context) {
+	if IsUpgradeViaDefaulting(ctx) {
+		if svc := it.Spec.ClientConfig.Service; svc != nil {
+			if svc.Port == 0 {
+				svc.Port = 80 // TODO: Should we use a pointer type instead of 0?
```
wlynch, line 28: 🤷  0 isn't a valid port, so it probably doesn't matter.

pkg/apis/triggers/v1alpha1/interceptor_type_types.go
MarcelMue: I think using a WebhookClientConfig is a good idea - maybe just reimplementing the type here if we are concerned about dependencies.

dibyom: yeah, I am basically reimplementing the type here....two differences:
1. using apis.URL instead of string for the URL type for better validation/json marshalling
2. we do not have a `caCert` field type (yet!) for HTTPS


Pulling in a new dependency for just the two structs  seems much so I decided to copy the structs. It also allows us to make changes (e.g. making the port a *int instead of int as suggested below)

docs/interceptortypes.md
dibyom: 🤦

pkg/apis/triggers/v1alpha1/interceptor_type_defaults.go
dibyom: I'll switch to a pointer type to make this clearer.

docs/interceptortypes.md
dibyom: If its not cluster scoped, each new namespace would have to install their own version of Interceptors....could we not use RBAC to restrict if certain interceptors need to be limited?

docs/interceptortypes.md
dibyom: hmmm yeah...this CRD is definitely based on the WebhookConfiguration CRDs....I don't have a strong preference on calling this InterceptorConfguration --> I thought it was a bit confusing in the sense its not clear if it is some interceptor configuration that can be reused within a trigger vs configuration for a new kind of interceptor.

docs/interceptortypes.md
dibyom: chatted with @wlynch about this. We will start off with a cluster scoped CRD but will implement a Namespace scoped version of this in the future as well.

docs/interceptortypes.md
wlynch: This still needs to be fixed 😅

docs/interceptortypes.md
wlynch: Thanks! Let's update the docs to reflect the change.

wlynch (commented):
/lgtm Looks good!

Approval pending doc changes / any other feedback from others.

docs/interceptortypes.md
MarcelMue: I personally don't love the `Configuration` suffix. It makes sense if we have a different `Interceptor` object but a `Configuration` suffix without an `Interceptor` CRD seems weird to me.

No strong opinions though.

MarcelMue (commented):
Docs are out of date, other than that LGTM.

dibyom:
Ok, I think I'm going to update the names to ClusterInterceptor (and Interceptor for the namespaced variant in the future).

docs/clusterinterceptors.md
```diff
@@ -0,0 +1,57 @@
...
+- Required:
+  - [`apiVersion`][kubernetes-overview] - Specifies the API version, for example
+    `triggers.tekton.dev/v1alpha1`.
+  - [`kind`][kubernetes-overview] - Specifies the `Trigger` resource
+    object.
+  - [`metadata`][kubernetes-overview] - Specifies data to uniquely identify the
+    `Interceptor` resource object, for example a `name`.
+  - [`spec`][kubernetes-overview] - Specifies the configuration information for
+    your Interceptor resource object. The spec include:
+    - [`clientConfig`] -  Specifies how a client (e.g. an EventListener) can communicate with the Interceptor.
```
wlynch, line 35: Few more Interceptor -> ClusterInterceptor replacements need to be made in here.

wlynch (commented):
/lgtm w/ minor doc changes.

Since this is a new type, would appreciate a second pair of eyes as an approver!

pkg/apis/triggers/v1alpha1/cluster_interceptor_defaults.go
```diff
@@ -0,0 +1,34 @@
...
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"context"
+)
+
+var defaultPort = int32(80)
```
savitaashture, line 23: Just a thought that 80 and 443 are kind of standard port 
So can we use some other port?

pkg/apis/triggers/v1alpha1/cluster_interceptor_defaults.go
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright 2019 The Tekton Authors
```
savitaashture, line 2:
```suggestion
Copyright 2021 The Tekton Authors
```

pkg/apis/triggers/v1alpha1/cluster_interceptor_defaults_test.go
```diff
@@ -0,0 +1,59 @@
...
+						Namespace: "default",
+						Name:      "github-svc",
+						Port:      ptr.Int32(80),
+					},
+				},
+			},
+		},
+	}}
+
+	for _, tc := range tests {
```
savitaashture, line 50: we have only one case so shall we write without using for loop ? i mean direct calling function

pkg/apis/triggers/v1alpha1/cluster_interceptor_defaults_test.go
```diff
@@ -0,0 +1,59 @@
+package v1alpha1_test
```
savitaashture, line 1: missing License

pkg/apis/triggers/v1alpha1/cluster_interceptor_types.go
```diff
@@ -0,0 +1,107 @@
+package v1alpha1
```
savitaashture, line 1: License

pkg/apis/triggers/v1alpha1/cluster_interceptor_types.go
```diff
@@ -0,0 +1,107 @@
...
+	// URL is a fully formed URL pointing to the interceptor
+	// Mutually exclusive with Service
+	URL *apis.URL `json:"url,omitempty"`
+
+	// Service is a reference to a Service object where the interceptor is running
+	// Mutually exclusive with URL
+	Service *ServiceReference `json:"service,omitempty"`
+}
+
+// ServServiceReference is a reference to a Service object
```
savitaashture, line 59:
```suggestion
// ServiceReference is a reference to a Service object
```

pkg/apis/triggers/v1alpha1/cluster_interceptor_types.go
```diff
@@ -0,0 +1,107 @@
...
+type ServiceReference struct {
+	// Name is the name of the service
+	Name string `json:"name"`
+
+	// Namespace is the namespace of the service
+	Namespace string `json:"namespace"`
+
+	// Path is an optional URL path
+	// +optional
+	Path string `json:"path"`
```
savitaashture, line 70: can we add `omitempty` as its optional field

pkg/apis/triggers/v1alpha1/cluster_interceptor_types.go
```diff
@@ -0,0 +1,107 @@
...
+// ResolveAddress returns the URL where the interceptor is running using its clientConfig
+func (it *ClusterInterceptor) ResolveAddress() (*apis.URL, error) {
+	if url := it.Spec.ClientConfig.URL; url != nil {
+		return url, nil
+	}
+	svc := it.Spec.ClientConfig.Service
+	if svc == nil {
+		return nil, ErrNilURL
+	}
+	port := int32(80)
```
savitaashture, line 97: can we use `defaultPort` from default.go file so that we can avoid hard coding here

pkg/apis/triggers/v1alpha1/cluster_interceptor_types_test.go
```diff
@@ -0,0 +1,86 @@
+package v1alpha1_test
```
savitaashture, line 1: shall we include license here ?

pkg/apis/triggers/v1alpha1/cluster_interceptor_validation.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright 2019 The Tekton Authors
```
savitaashture, line 2:
```suggestion
Copyright 2021 The Tekton Authors
```

pkg/apis/triggers/v1alpha1/cluster_interceptor_types.go
dibyom: done

pkg/apis/triggers/v1alpha1/cluster_interceptor_types.go
dibyom: done!

pkg/apis/triggers/v1alpha1/cluster_interceptor_defaults.go
dibyom: users can always specify a different port. I think 80 makes a good default since otherwise every interceptor author will have to specify the "custom" default port that we added. 

(When we add TLS support, we could try 443 as default perhaps -- that's what the WebhookConfiguration CRDs do.)

savitaashture (commented):
/lgtm

savitaashture (approved):

PR #966, opened 2021-02-23
Rewrite the Tekton Triggers installation instructions for clarity and flow

sergetron:
# Changes

Rewrites the Tekton Triggers installation instructions for clarity and flow.

# Submitter Checklist

review them:

- [X] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#docs) (if user facing)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -1,93 +1,67 @@
 <!--
 ---
-linkTitle: "Installation"
+linkTitle: "Setting Up Tekton Triggers"
 weight: 2
 ---
 -->
-# Installing Tekton Triggers
+# Setting Up Tekton Triggers
 
-Use this page to add the component to an existing Kubernetes cluster.
+Follow the steps below to set up an official release of Tekton Triggers on your Kubernetes cluster. 
+If you want to test Triggers from `HEAD`, see the
+[Tekton Developer Guide](https://github.com/tektoncd/triggers/blob/master/DEVELOPMENT.md#install-triggers).
 
-- [Pre-requisites](#pre-requisites)
-- [Versions](#versions)
-- [Installing Tekton Triggers](#installing-tekton-triggers-1)
+## Prerequisites
 
-## Pre-requisites
+You must satisfy the following prerequisties to set up Tekton Triggers:
 
-1.  A Kubernetes cluster version 1.18 or later
+* You must have a Kubernetes Cluster running Kubernetes 1.18 or above. 
 
-    If you don't already have a cluster, you can create one for testing with
-    `kind`.  [Install
-    `kind`](https://kind.sigs.k8s.io/docs/user/quick-start/#installation) and
-    create a cluster by running [`kind create
-    cluster`](https://kind.sigs.k8s.io/docs/user/quick-start/#creating-a-cluster).
-    This will create a cluster running locally, with RBAC enabled.
+  You can use [`kind`](https://kind.sigs.k8s.io/) to quickly create a local cluster with RBAC enabled for testing purposes:
 
-1.  Grant current user `cluster-admin` privileges.
+  * Install `kind` as described in [Installation](https://kind.sigs.k8s.io/docs/user/quick-start/#installation).
 
-    _See
-    [Role-based access control](https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#prerequisites_for_using_role-based_access_control)
-    for more information._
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -1,93 +1,58 @@
...
 
-    _See
-    [Role-based access control](https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#prerequisites_for_using_role-based_access_control)
-    for more information._
+## Installing Tekton Triggers on Your Cluster
 
-1.  Install Tekton Pipelines. You can install the latest version using the
-    command below or follow the
-    [pipeline installation guide](https://github.com/tektoncd/pipeline/blob/master/docs/install.md):
+1. Open a terminal on your cluster and log in as the user with which you installed Tekton Pipelines.
```
MarcelMue, line 26: This seems a bit misleading to me. You don't open a terminal on your cluster technically. You mainly need a `kubeconfig` which is pointing at the cluster with the correct role.

docs/install.md
```diff
@@ -1,93 +1,58 @@
...
-- [Pre-requisites](#pre-requisites)
-- [Versions](#versions)
-- [Installing Tekton Triggers](#installing-tekton-triggers-1)
+## Prerequisites
 
-## Pre-requisites
+You must satisfy the following prerequisties to set up Tekton Triggers:
 
-1.  A Kubernetes cluster version 1.18 or later
+* Your cluster must be running Kubernetes version 1.18 or above.
```
MarcelMue, line 17:
```suggestion
* You must have a Kubernetes cluster running version 1.18 or later.
```

I would also be in favor of keeping the `kind` instructions as they match `pipeline` install guides:
https://github.com/tektoncd/pipeline/blob/master/docs/install.md#before-you-begin

MarcelMue (commented):
I like the general direction.

docs/install.md
sergetron: I want to avoid duplication of instructions wherever possible because they can quickly go out of sync between all their different locations. Plus, they must have Pipelines running before installing Triggers, they can't run Triggers without Pipelines, so either way they must go look at Pipelines installation instructions.

docs/install.md
sergetron: Fair! Reworded.

MarcelMue:
I think a review from @dibyom would be good - he worked on these docs last.

/LGTM

docs/install.md
```diff
@@ -1,93 +1,58 @@
...
+## Prerequisites
 
-## Pre-requisites
+You must satisfy the following prerequisties to set up Tekton Triggers:
 
-1.  A Kubernetes cluster version 1.18 or later
+* You must have Tekton Pipelines installed on your Kubernetes cluster. For instructions, see [Installing Tekton Pipelines](https://github.com/tektoncd/pipeline/blob/master/docs/install.md).
 
-    If you don't already have a cluster, you can create one for testing with
-    `kind`.  [Install
```
iancoffey, line 20: I think it is helpful to still briefly mention kind here in this context. imo, it helps ensure we have a path with low barriers to contribution and get folks on a common path.

docs/install.md
```diff
@@ -1,93 +1,58 @@
...
-    `kind`](https://kind.sigs.k8s.io/docs/user/quick-start/#installation) and
-    create a cluster by running [`kind create
-    cluster`](https://kind.sigs.k8s.io/docs/user/quick-start/#creating-a-cluster).
-    This will create a cluster running locally, with RBAC enabled.
+* Your Kubernetes cluster must be running Kubernetes version 1.18 or above.
```
dibyom, line 19: I'd mention this before the point about installing Tekton Pipelines. And I'd keep the bit about being able to create a cluster easily using Kind. It does make things a bit more verbose but IMO makes the docs more approachable for someone reading this for the first time.

dibyom (commented):
Thanks @sergetron This looks much better! Only one small point about mentioning how to install k8s.

docs/install.md
sergetron: Restored the `kind` instructions as per review comments.

docs/install.md
sergetron: Restored the `kind` instructions as per review comments.

docs/install.md
sergetron: Restored the `kind` instructions as per review comments and moved the Pipelines requirement after the Kubernetes cluster requirement.

sergetron:
Updated! Please review and re-LGTM.

dibyom (approved):
Thanks Serge!
/lgtm

PR #971, opened 2021-02-25
Try to proof that extensions keys are overriden using multiple webhooks

fiunchinho:
# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/sink/sink_test.go
+++ b/pkg/sink/sink_test.go
@@ -1188,6 +1188,13 @@ func TestExecuteInterceptor_ExtensionChaining(t *testing.T) {
 					},
 				},
 			}, {
+				CEL: &triggersv1.CELInterceptor{
+					Overlays: []triggersv1.CELOverlay{{
+						Key:        "something",
+						Expression: "body.pull_request_url",
+					}},
+				},
+			}, {
 				Webhook: &triggersv1.WebhookInterceptor{
 					ObjectRef: &corev1.ObjectReference{
 						APIVersion: "v1",
@@ -1221,6 +1228,7 @@ func TestExecuteInterceptor_ExtensionChaining(t *testing.T) {
 				"pull_request_body": "key",
 				"pull_request_url":  "https://github.com/some/pr",
 			},
+			"something": "https://github.com/some/pr",
 		},
 	}
 	var gotBody map[string]interface{}
@@ -1247,6 +1255,7 @@ func TestExecuteInterceptor_ExtensionChaining(t *testing.T) {
 		"add_pr_body": map[string]interface{}{
 			"pull_request_url": "https://github.com/some/pr",
 		},
+		"something": "https://github.com/some/pr",
 	}
 
 	if diff := cmp.Diff(iresp.Extensions, wantExtensions); diff != "" {
--- a/pkg/sink/sink_test.go
+++ b/pkg/sink/sink_test.go
@@ -1129,7 +1129,8 @@ func TestExecuteInterceptor_NotContinue(t *testing.T) {
 
 // echoInterceptor stores and returns the body back
 type echoInterceptor struct {
-	body map[string]interface{}
+	body    map[string]interface{}
+	counter int
 }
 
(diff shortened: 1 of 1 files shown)


pkg/sink/sink_test.go
```diff
@@ -1140,7 +1141,13 @@ func (f *echoInterceptor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	defer r.Body.Close()
+	if f.counter == 0 {
```
fiunchinho, line 1144: We are using the same interceptor twice as if there were two different `Webhooks`.

The first webhook adds a new key to the extensions key called `truncated_sha` that was added by first `cel` interceptor.

The second webhook adds a new key to the root extensions key.

pkg/sink/sink_test.go
```diff
@@ -1201,7 +1216,11 @@ func TestExecuteInterceptor_ExtensionChaining(t *testing.T) {
 	wantBody := map[string]interface{}{
 		"sha": sha,
 		"extensions": map[string]interface{}{
-			"truncated_sha": "abcde",
+			"completely": "different",
+			"truncated_sha": map[string]interface{}{
+				"gone": "key",
```
fiunchinho, line 1221: When we run this test it fails because the `gone` key is, well, gone. When the second `Webhook` is executed, the `truncated_sha` key added by first `cel` interceptor in extensions gets re-applied, so *the keys added to `truncated_sha` are lost*.

pkg/sink/sink_test.go
```diff
@@ -1140,7 +1141,13 @@ func (f *echoInterceptor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	defer r.Body.Close()
+	if f.counter == 0 {
+		data["extensions"].(map[string]interface{})["truncated_sha"].(map[string]interface{})["gone"] = "key"
```
MarcelMue, line 1145: Can you change the `truncated_sha` key to something more realistic? E.g. `my_interceptor`

pkg/sink/sink_test.go
```diff
@@ -1192,16 +1207,20 @@ func TestExecuteInterceptor_ExtensionChaining(t *testing.T) {
...
 	}
 
 	wantBody := map[string]interface{}{
-		"sha": sha,
+		"pull_request_url": pull_request_url,
 		"extensions": map[string]interface{}{
-			"truncated_sha": "abcde",
+			"add_team_members": "something",
+			"add_pr_body": map[string]interface{}{
+				"pull_request_body": "key",
```
fiunchinho, line 1221: When we run this test it fails because the `pull_request_body` key is gone. When the second `Webhook` is executed, **the `add_pr_body` key added by first cel interceptor in extensions gets re-applied so the keys added to `add_pr_body` on the first `Webhook` are lost when it reaches the second `Webhook`.**

MarcelMue:
@dibyom Can you also take a look here? I can confirm this problem from my local testing but no idea why this is happening.

pkg/sink/sink_test.go
dibyom: yeah I think I get it....this happens if we are using nested json keys i.e. if both keys were added at the top-level instead of nested under `add_pr_body` I think this behavior would not happen.


I think the lines to blame is here: https://github.com/tektoncd/triggers/blob/master/pkg/sink/sink.go#L351

We are simply accounting for top level keys (like  add_pr_body) and resetting its value to the one from request.Extensions. Instead we need to do some sort of a merge. 

One way might be to marshal the body into a map[string]interface{}, then do a json merge, and then turn it back into a []byte.

jmcshane:
@fiunchinho this should be resolved in the v0.13 release using custom interceptors as the target of an interceptor chain. Hopefully with #1035 the documentation will be thorough enough to help you refactor your webhooks into custom interceptors returning extensions.

fiunchinho:
ok, in that case I'll close this one. Thank you very much.

PR #977, opened 2021-03-02
Forcing eventlistener sink to resolve eventlistener

jmcshane:
# Changes

This change forces the eventlistenersink process to be able to resolve
the EventListener before starting the HTTP server.

This means that the sink HTTP process (and readiness probe)
will not be listen until the EventListenerLister is able to resolve
the EventListener from the API server.

This is especially useful in startup cases, but can also assist
if the pod is started without permission to read the EventListener
object. In this situation, given this change, the eventlistener
pod will restart with a logged error message about lack of access to
that specific API resource.

# Submitter Checklist

review them:

# Release Notes

```release-note
Force eventlistener pod to resolve EventListener resource before startup
```

### DIFF
--- a/cmd/eventlistenersink/main.go
+++ b/cmd/eventlistenersink/main.go
@@ -124,7 +124,7 @@ func main() {
 	}
 	err = r.WaitForEventListener(eventListenerBackoff)
 	if err != nil {
-		panic(err)
+		logger.Fatal(err)
 	}
 
 	// Listen and serve
--- a/cmd/eventlistenersink/main.go
+++ b/cmd/eventlistenersink/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/tektoncd/triggers/pkg/client/informers/externalversions"
 	triggerLogging "github.com/tektoncd/triggers/pkg/logging"
 	"github.com/tektoncd/triggers/pkg/sink"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -113,6 +114,17 @@ func main() {
 		TriggerBindingLister:        factory.Triggers().V1alpha1().TriggerBindings().Lister(),
 		ClusterTriggerBindingLister: factory.Triggers().V1alpha1().ClusterTriggerBindings().Lister(),
 		TriggerTemplateLister:       factory.Triggers().V1alpha1().TriggerTemplates().Lister(),
+	}
+	eventListenerBackoff := wait.Backoff{
+		Duration: 100 * time.Millisecond,
+		Factor:   2.5,
+		Jitter:   0.3,
+		Steps:    10,
+		Cap:      5 * time.Second,
+	}
+	err = r.WaitForEventListener(eventListenerBackoff)
+	if err != nil {
+		panic(err)
 	}
 
 	// Listen and serve
--- a/pkg/sink/initialization.go
+++ b/pkg/sink/initialization.go
@@ -18,10 +18,13 @@ package sink
 
 import (
 	"flag"
+	"fmt"
 	"time"
 
 	triggersclientset "github.com/tektoncd/triggers/pkg/client/clientset/versioned"
 	"golang.org/x/xerrors"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	discoveryclient "k8s.io/client-go/discovery"
 	kubeclientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -132,4 +135,20 @@ func ConfigureClients(clusterConfig *rest.Config) (Clients, error) {
 		RESTClient:      kubeClient.RESTClient(),
 		TriggersClient:  triggersClient,
 	}, nil
+}
+
+// WaitForEventListener waits for the Kubernetes eventlistener configuration
+func (r *Sink) WaitForEventListener(backoff wait.Backoff) error {
+
+	if err := wait.ExponentialBackoff(backoff, func() (bool, error) {
+		if _, err := r.EventListenerLister.EventListeners(r.EventListenerNamespace).Get(r.EventListenerName); errors.IsNotFound(err) {
+			return false, nil
+		} else if err != nil {
+			return false, err
+		}
+		return true, nil
+	}); err != nil {
+		return fmt.Errorf("Unable to retrieve EventListener %s in Namespace %s: %s", r.EventListenerName, r.EventListenerNamespace, err)
+	}
+	return nil
 }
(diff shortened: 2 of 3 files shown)


pkg/sink/initialization.go
```diff
@@ -133,3 +135,28 @@ func ConfigureClients(clusterConfig *rest.Config) (Clients, error) {
 		TriggersClient:  triggersClient,
 	}, nil
 }
+
+var elBackoff = wait.Backoff{
+	Duration: 50 * time.Millisecond,
+	Factor:   2.0,
+	Jitter:   0.3,
+	Steps:    10,
```
dibyom, line 143: If I'm understanding this correctly, it will try a max of 10 times right? (Not try for a max of 5seconds?)

jmcshane: so, its a little of both. the steps is the max number of times, if it ever hits the cap it will set the wait time to that amount and then bail out.

in this case, the first wait time is 50 milliseconds with a scale factor of 2 and jitter of 0.3. The jitter means the scale factor will be somewhere between 1.7 and 2.3. This means that after 10 steps, it waits somewhere around 10 seconds max before it fails

pkg/sink/initialization.go
```diff
@@ -133,3 +135,28 @@ func ConfigureClients(clusterConfig *rest.Config) (Clients, error) {
...
+		if errors.IsNotFound(err) {
+			return false, nil
+		} else if err != nil {
+			return false, err
+		}
+		return true, nil
+	})
+	if err != nil {
+		r.Logger.Fatalw("Unable to retrieve EventListener %s in Namespace %s: %s", r.EventListenerName, r.EventListenerNamespace, err)
+	}
```
savitaashture, line 161:
```suggestion
	if err := wait.ExponentialBackoff(elBackoff, func() (bool, error) {
		if _, err := r.EventListenerLister.EventListeners(r.EventListenerNamespace).Get(r.EventListenerName); errors.IsNotFound(err) {
			return false, nil
		} else if err != nil {
			return false, err
		}
		return true, nil
	}); err != nil {
		r.Logger.Fatalw("Unable to retrieve EventListener %s in Namespace %s: %s", r.EventListenerName, r.EventListenerNamespace, err)
	}
```

pkg/sink/initialization_test.go
```diff
@@ -101,3 +111,54 @@ func Test_GetArgs_error(t *testing.T) {
...
+		EventListenerName:      eventListenerName,
+		EventListenerNamespace: namespace,
+		EventListenerLister:    eventlistenerinformer.Get(ctx).Lister(),
+	}
+
+	r.WaitForEventListenerOrDie()
+}
+
+func TestWaitForEventlistener_Fatal(t *testing.T) {
+	if os.Getenv("EL_FATAL_CRASH") == "1" {
```
savitaashture, line 137: i executed locally this testcase and i see the code under if block never runs until we set env `EL_FATAL_CRASH` 
If thats the case should we keep this? or am i missing something here :thinking:

jmcshane: so, the reason that i do this indirection is because the fatal execution will kill the `go test` process. The way we get around this is by creating a subprocess that then fails and adds environment variable that forces the crash.

If the tests were run with EL_FATAL_CRASH in the root process, the `Fatalw` invocation in the backoff would cause the tests to be marked as failed in their entirety

pkg/sink/initialization.go
dibyom: makes sense...do you think 10 seconds is quick enough? I was thinking 30s might be a better default?

pkg/sink/initialization.go
```diff
@@ -133,3 +135,26 @@ func ConfigureClients(clusterConfig *rest.Config) (Clients, error) {
...
+var elBackoff = wait.Backoff{
+	Duration: 50 * time.Millisecond,
+	Factor:   2.0,
+	Jitter:   0.3,
+	Steps:    10,
+	Cap:      5 * time.Second,
+}
+
+// WaitForEventListenerOrDie waits for the Kubernetes eventlistener configuration
+func (r *Sink) WaitForEventListenerOrDie() {
```
dibyom, line 148: Can this elBackOff be a param that is passed to this function? That will also make the test a bit more easier to follow since we won't be modifying the variable in this package to test.

pkg/sink/initialization_test.go
```diff
@@ -101,3 +111,54 @@ func Test_GetArgs_error(t *testing.T) {
...
+			Factor:   1.0,
+			Jitter:   0.1,
+			Steps:    1,
+			Cap:      100 * time.Millisecond,
+		}
+		// will fail
+		r.WaitForEventListenerOrDie()
+	}
+	cmd := exec.Command(os.Args[0], "-test.run=TestWaitForEventlistener_Fatal") //nolint:gosec
+	cmd.Env = append(os.Environ(), "EL_FATAL_CRASH=1")
```
dibyom, line 158: this is clever...though we might be able to get away with passing in a test logger and capturing its output and then comparing (like we do here: https://github.com/tektoncd/triggers/blob/1066d18ecb7b1e4ca88472251a2561e5c684046c/pkg/sink/sink_test.go#L926-L943)

What do you think?

Another option would be to have the function return an error and then have main.go call the log.Fatal.

dibyom (commented):
Just two minor comments. Otherwise good to go!

pkg/sink/initialization_test.go
jmcshane: yes, let me take a look at this strategy and try to implement it here

pkg/sink/initialization.go
jmcshane: Let me increase the factor and the num steps. Again, this is basically a question of:

`how long will it take for the k8s API server to commit the eventlistener`

I think closer to the 15-20 seconds is appropriate, especially considering that this is a function that blocks the startup of the HTTP endpoint. Let me modify the factor so it ends up around this value

pkg/sink/initialization_test.go
jmcshane: so, it doesn't look like this is possible because the behavior of fatal can't be overridden even though there is a noop action in the library. basically, its not a supported action on fatal.

i updated this function to return an error and panic instead

cmd/eventlistenersink/main.go
```diff
@@ -114,6 +115,17 @@ func main() {
...
+	eventListenerBackoff := wait.Backoff{
+		Duration: 100 * time.Millisecond,
+		Factor:   2.5,
+		Jitter:   0.3,
+		Steps:    10,
+		Cap:      5 * time.Second,
+	}
+	err = r.WaitForEventListener(eventListenerBackoff)
+	if err != nil {
+		panic(err)
```
dibyom, line 127:
```suggestion
		logger.Fatalf(err)
```

dibyom:
/approve

Just one minor thing -> panic to logger.Fatal in main

Issue #983, opened 2021-03-05
EventListener fails to reconcile when using Knative Service with "validation.webhook.serving.knative.dev" denied the request


Fabian-K:
Hi,

thanks for all the effort to implement Knative Service with EventListeners! Unfortunately I´m struggeling with the following issue:

# Expected Behavior
Reconciliation loop is successful when using Knative Service with EventListeners

# Actual Behavior
Reconciliation loop is stuck with
```
{"level":"info","ts":"2021-03-05T09:01:10.826Z","logger":"controller.event-broadcaster","caller":"record/event.go:282","msg":"Event(v1.ObjectReference{Kind:\"EventListener\", Namespace:\"default\", Name:\"sample\", UID:\"edf55d3f-4278-4615-9016-6fde4c57aa2a\", APIVersion:\"triggers.tekton.dev/v1alpha1\", ResourceVersion:\"60353\", FieldPath:\"\"}): type: 'Normal' reason: 'FinalizerUpdate' Updated \"sample\" finalizers"}
{"level":"info","ts":"2021-03-05T09:01:10.969Z","logger":"controller","caller":"eventlistener/eventlistener.go:501","msg":"Created EventListener Deployment el-sample in Namespace default","knative.dev/controller":"github.com.tektoncd.triggers.pkg.reconciler.v1alpha1.eventlistener.Reconciler","knative.dev/kind":"triggers.tekton.dev.EventListener","knative.dev/traceid":"d3cec124-3d93-455a-a07a-2f956164212c","knative.dev/key":"default/sample"}
{"level":"info","ts":"2021-03-05T09:01:11.035Z","logger":"controller","caller":"controller/controller.go:530","msg":"Reconcile succeeded","knative.dev/controller":"github.com.tektoncd.triggers.pkg.reconciler.v1alpha1.eventlistener.Reconciler","knative.dev/kind":"triggers.tekton.dev.EventListener","knative.dev/traceid":"d3cec124-3d93-455a-a07a-2f956164212c","knative.dev/key":"default/sample","duration":0.219598241}
[...]

PR #995, opened 2021-03-11
Fix reconciler issue while creating knative service

savitaashture:
# Changes

Fix: https://github.com/tektoncd/triggers/issues/983

**Root Cause:**
The return type of `unstructured.NestedFieldCopy` is `map[string]interface{}` in code we were comparing with `map[string]string` and because of that annotations were not matching.
Knative adds strict check on default annotation so because of that validation was failing.

/cc @dibyom @Fabian-K 

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```


savitaashture:
it seems some infra issues

pkg/dynamic/custom_reconcile.go
```diff
@@ -238,6 +238,19 @@ func MergeMaps(m1, m2 map[string]string) map[string]string {
 	return merged
 }
 
+// mergeMaps merges the values in the passed maps into a new map.
+// Values within m2 potentially clobber m1 values.
+func mergeMaps(m1, m2 map[string]interface{}) map[string]interface{} {
```
dibyom, line 243: small nit: we already have a function called MergeMaps, so maybe call this mergeStringInterfaceMaps (or call the other one mergeStringMaps)?

dibyom:
Also: I don't know how hard it is but could we add an unit test to verify the fix?

savitaashture:
@dibyom i have refactored and updated code
So now we don't require extra `MergeMaps` function

dibyom:
We don't have to do this for this PR but we should track adding a full e2e test for the knative EL (perhaps using https://github.com/csantanapr/knative-kind and the e2e kind stuff that Billy set up)

savitaashture:
> We don't have to do this for this PR but we should track adding a full e2e test for the knative EL (perhaps using https://github.com/csantanapr/knative-kind and the e2e kind stuff that Billy set up)

+1 
right now if i add e2e the pre req would be cluster should have Knative running 

But as you said we should do this sometime

dibyom:
> right now if i add e2e the pre req would be cluster should have Knative running

Let's open an issue to track this (Billy has been doing some work on running e2e tests using kind -- we could piggy back off of that)

PR #976, opened 2021-02-26
Migrate core interceptors to use InterceptorType CRD

dibyom:
# Changes

Migrate core interceptors to use InterceptorType CRD

This commit does the following:
1. Add YAML definitions for installing each core interceptor using the
new InterceptorType CRD.
2. Modify the EventListener to use the InterceptorType CRD to find the
Interceptor's URL.

Fixes #868

This is NOT a breaking change since users can still configure their interceptors in the same way was before. Changes to the way a TriggerInterceptor is configured is coming in #869.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#tests) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#docs) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commits)
- [ ] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/triggers/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

Depends on #960

Issue #869, opened 2020-12-16
Interceptors:  Implement new config API aka how to use interceptors in Triggers


dibyom:
The work for this issue is to implement changes to how users writing a Trigger configure Interceptors as described in [TEP-0026](https://github.com/tektoncd/community/blob/master/teps/0026-interceptor-plugins.md#using-interceptors-in-triggers) i.e. how to supply params declared by an Interceptor in a Trigger.

# What?

Discuss options for how we should configure interceptors inside a Trigger.

# Why?

TEP-009: Since interceptors are pluggabble, we can't make them part of the API spec anymore. We need a pluggable mechanism!

# How do we do it today?

```yaml
interceptors:
  - cel:
       filter:  `body.comment == "plz run my test"` # simplest case: string key-val
       overlays:                   # most complex obj that contains array of objs
          - key: "a_new_field"
            value: "2 * 5"
          - key: "another_field"
            value: "bar"
  -  github:
        eventTypes: ["pull_request", "push"]   # val is of type array of strings
        secretRef:                             # complex obj
           name: "github-secret"
           key: "webhook-token"
```

## Design:

1. We need a way to refer to the Interceptor. The simplest, most consistent way would be to use a `ref` as we do with other resources e.g.

```yaml
interceptors:
  - ref:
      apiVersion: interceptors.triggers.tekton.dev/v1alpha1
      kind: cel
```

1. (Based on a previous WG discussion) It might also makes sense to add an optional `name` field to identify what interceptor block we are using. This makes it more descriptive when you have multiple interceptors, and can be useful when using kustomize.

```yaml
interceptors:
  - name: "payload-validation"
    ref:
      apiVersion: interceptors.triggers.tekton.dev/v1alpha1
      kind: github
  - name: "Cel -filter"
    ref:
      apiVersion: interceptors.triggers.tekton.dev/v1alpha1
      kind: cel
```

1. Finally, we need a way to supply parameters. Today, the paramter names are always keys while the values can be strings, array of strings (eventTypes in github), or complex objects (e.g. overlays in CEL, secretRef in github). The following section has some ideas for representing these:

#### 1. Only allow string values

1. This is the simplest option and consistent with how we use params elsewhere.
1. We'd have to figure out a way to represent the current params that use non string types:

<pre>
interceptors:
- name: "github-validation"
  ref:
    kind: GitHub
  params:
  - name: "eventTypes"
    value: "pull_request,"push" <strong># part of same string separated by commas</strong>
[...]

PR #1001, opened 2021-03-16
 Add API to configure Interceptors from a Trigger

dibyom:
# Changes

This commit adds new fields (`ref`, `param`, `name`) to refer to a
pluggable Interceptor from within a Trigger's spec. Previously a user
could either reference one of core interceptors (e.g. `cel`) or use the
`webhook` field to refer to webhook interceptors. The new API is a
replacement for these old options though the old API will continue to
work for now to ensure backwards compatibility.

The new fields are:
1. `ref`: This is required and references a ClusterInterceptor
2. `params`: Any optional params to pass on to the Interceptor. The
values can be any valid JSON.
3. `name`: Optional name to configure a named interceptor config.

Part of #869
Depends on #976 

In a future PR, we will add logic to automatically upgrade old style TriggerInterceptors to the new style.

# Submitter Checklist

review them:

# Release Notes

```release-note
Trigger spec authors can now configure interceptors using a new API that includes  a `ref` field to refer to an Interceptor, and a `params` field to add parameters to pass on to the interceptor for processing.
```


sergetron:
Doc changes look good. I will include them in my rewrites of the EventListener and Interceptor docs next week.

pkg/apis/triggers/v1alpha1/trigger_types.go
```diff
@@ -82,25 +83,67 @@ type Trigger struct {
...
-	Bitbucket *BitbucketInterceptor `json:"bitbucket,omitempty"`
+	// Optional name to identify the current interceptor configuration
+	Name *string `json:"name,omitempty"`
+	// Ref refers to the Interceptor to use
+	Ref InterceptorRef `json:"ref"`
+	// Params are the params to send to the interceptor
+	Params []InterceptorParams `json:"params,omitempty"`
+
+	// Deprecated old fields below
+	Webhook             *WebhookInterceptor   `json:"webhook,omitempty"`
```
wlynch, line 94: Should this also be `DeprecatedWebhook`?

pkg/apis/triggers/v1alpha1/trigger_types.go
```diff
@@ -82,25 +83,67 @@ type Trigger struct {
...
+	DeprecatedCEL       *CELInterceptor       `json:"cel,omitempty"`
+	DeprecatedBitbucket *BitbucketInterceptor `json:"bitbucket,omitempty"`
+}
+
+type InterceptorParams struct {
+	Name  string               `json:"name"`
+	Value apiextensionsv1.JSON `json:"value"`
+}
+
+type InterceptorRef struct {
```
wlynch, line 106: nit: Missing godoc comment

pkg/apis/triggers/v1alpha1/trigger_types.go
```diff
@@ -82,25 +83,67 @@ type Trigger struct {
...
+
+	// Deprecated old fields below
+	Webhook             *WebhookInterceptor   `json:"webhook,omitempty"`
+	DeprecatedGitHub    *GitHubInterceptor    `json:"github,omitempty"`
+	DeprecatedGitLab    *GitLabInterceptor    `json:"gitlab,omitempty"`
+	DeprecatedCEL       *CELInterceptor       `json:"cel,omitempty"`
+	DeprecatedBitbucket *BitbucketInterceptor `json:"bitbucket,omitempty"`
+}
+
+type InterceptorParams struct {
```
wlynch, line 101: nit: Missing godoc comment

wlynch (commented):
LGTM! Few minor changes.

savitaashture:
@dibyom I see update with new interceptor changes for all the examples
shall we keep few examples to use old syntax until we remove ?

pkg/apis/triggers/v1alpha1/trigger_types.go
dibyom: We need to deprecate Webhooks too but that will be a more involved process since we cannot automatically upgrade existing webhooks to the new API. So, I decided to keep Webhook as non deprecated. (though i put the comment in the wrong spot 😓 )

wlynch (approved):
/lgtm

Issue #1014, opened 2021-03-23
Missing resource cleanup when switching an existing EventListener to make use of Knative Services


Fabian-K:
# Expected Behavior

When updating the spec of an existing EventListener to use Knative Service, the now obsolete resources (Deployment, Service) should be deleted.

# Actual Behavior

Deployment and Service are not deleted. This also prevents the Knative Service from rolling out successfully because it does not own the Service `el-*` ("There is an existing placeholder Service "el-sample" that we do not own")

# Steps to Reproduce the Problem

1. Create a "traditional" EventListener and wait until it´s ready
2. Update the EventListener to use Knative Service instead of a Deployment (https://github.com/tektoncd/triggers/blob/main/docs/eventlisteners.md#customresource)

# Additional Info
- Kubernetes version: `v1.19.4`
- Tekton Triggers version: `0.12.1`

Let me know if you need additional info!

Thanks, Fabian

PR #1022, opened 2021-03-29
Cleanup existing k8s resources after updating to customeResources

savitaashture:
# Changes

Fixes : https://github.com/tektoncd/triggers/issues/1014

/cc @dibyom @Fabian-K 

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```


pkg/reconciler/v1alpha1/eventlistener/eventlistener_test.go
```diff
@@ -1878,3 +1878,45 @@ func Test_generateObjectMeta(t *testing.T) {
...
+							NodeSelector: map[string]string{
+								"hi": "hello",
+							},
+						},
+					}},
+			}),
+		}
+	})
+	startResources := test.Resources{
+		EventListeners: []*v1alpha1.EventListener{elWithCustomResourceForNodeSelector},
```
dibyom, line 1905: making sure I'm understanding this test properly:

the initial state is that we have a k8s deployment and an EL with customResource. We should expect the end state to be that we have a knative service present but not the deployment. Right? If so, shouldn't we look at the resources remaining after reconcile and verify that we do not have a deployment/service present?

savitaashture: Yes you are right

I have updated testcase to validate the resources left after reconcile

dibyom (approved):
/lgtm

PR #1029, opened 2021-04-01
Migrate core interceptors to new format

### LINKED ISSUE #1009: Migrate deprecated TriggerInterceptors to new API
### Feature request

In https://github.com/tektoncd/triggers/pull/1001 we added a new API for configuring interceptors in a Trigger and deprecated the old way (e.g. directly referencing `github`, `gitlab` etc.). We should add logic to our defaulting webhook to automatically switch any uses of old style of interceptors to the new API. We also need to call this logic from the reconciler so that already created Triggers also benefit from this.

dibyom:
# Changes

This commit updates the defaulting webhook to automatically set any Triggers
using the old style syntax for core interceptors to the new ref/params based
syntax.

Fixes #1009

# Submitter Checklist

review them:

# Release Notes

```release-note
Any new Triggers created using the old style syntax for core interceptors is now automatically switched to the new refs/params based syntax.
```

### DIFF
--- a/pkg/apis/triggers/v1alpha1/trigger_types.go
+++ b/pkg/apis/triggers/v1alpha1/trigger_types.go
@@ -17,6 +17,9 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"encoding/json"
+	"fmt"
+
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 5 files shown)


dibyom:
(The drop in coverage is due to missing tests for some err cases where the error happens due to malformed JSON. this should never happen in practice since we are converting from concrete structs)

savitaashture:
> (The drop in coverage is due to missing tests for some err cases where the error happens due to malformed JSON. this should never happen in practice since we are converting from concrete structs)

:+1: 

How about to pass invalid json data from testcase to increase the coverage :thinking:

pkg/apis/triggers/v1alpha1/trigger_types.go
```diff
@@ -135,6 +138,74 @@ func (ti *TriggerInterceptor) defaultInterceptorKind() {
...
 
+func (ti *TriggerInterceptor) updateCoreInterceptors() error {
+	if ti.Ref.Name != "" {
+		return nil
+	}
+	ti.Ref.Name = ti.GetName()
+	ti.Params = []InterceptorParams{}
+	switch ti.Ref.Name {
+	case "bitbucket":
+		if err := addToParams(&ti.Params, "secretRef", ti.DeprecatedBitbucket.SecretRef); err != nil {
```
wlynch, line 149: These could nil panic if ti or DeprecatedBitbucket is nil. We should add checks to prevent this (as well as for the other fields)

wlynch (commented):
Mostly LGTM! Just one small concern around pointer derefs.

dibyom:
> How about to pass invalid json data from testcase to increase the coverage 🤔

That's going to be hard since we are not actually passing JSON but a TriggerInterceptor struct which will always serialize to valid JSON.

pkg/apis/triggers/v1alpha1/trigger_types.go
dibyom: Added a check for when ti is nil. 
A bit convoluted but the checks for the other fields being not nil happens in the call ti.GetName() which ensures that if the name is bitbucket, then ti.DeprecatedBitbucket is not nil: https://github.com/tektoncd/triggers/blob/main/pkg/apis/triggers/v1alpha1/trigger_types.go#L145

savitaashture (approved):

wlynch (approved):
/lgtm

Issue #697, opened 2020-08-04
Proposal: Make triggers less generic and more coupled to Pipelines


bobcatfish:
TriggerTemplates are very generic and can (theoretically) be used to create anything. We initially made them very generic on purpose b/c this felt like a better design: less coupling = better!

However, by making them less generic, the interface is pretty verbose and we lose out on validation (i.e. we have no idea what you want to create until you try to create it).

We've also started to talk about co-ordinating releases https://github.com/tektoncd/plumbing/issues/413 and it seems like the releases that it makes the most sense to combine are pipelines + triggers. We've also debated how to display what versions of triggers + pipelines items in the catalog are compatible with, and what the CLI and dashboard will be compatible with.

I'd like to propose: what if we release Triggers + Pipelines together? If we're doing that, what if we couple them together as well? And if want to take this to its extreme, we could even move them into the same repo (this would simplify things such as consuming updates from knative/pkg for our controllers).

We don't have to go that far just yet, but in the near term I'd like to propose an update to TriggerTemplates that is more coupled to Pipelines. If we go this route, we might even want to rename the type, and deprecate TriggerTemplates.

So my proposal is: replace `TriggerTemplates` with something else. I'm not sure what to call it, it's not exactly a template, but it would be used to create instances of PipelineRuns, TaskRuns and Runs - maybe it's a "factory"? ( @ImJasonH you might have some naming ideas here) "recipe" even? `PipelineRecipe` <-- that's my strawman proposal but totally open to anything better.

Here's what it could look like (based loosely on https://github.com/tektoncd/plumbing/blob/0ccc386dc6247a4b6e0669d8ab2b11a3f5a81d51/tekton/ci/plumbing-template.yaml):

```
# There's definitely a question of what api group this is part of: it's so coupled
# to Pipelines that maybe it makes sense to have it in the pipelines api group?
# If we DO want to go as far as combining the repos, we might want to use the same
# api group for both triggers and pipelines
apiVersion: triggers.tekton.dev/v1beta1
kind: PipelineRecipe
metadata:
  name: tekton-plumbing-ci-pipeline
spec:
  params:
  - name: buildUUID
  - name: package
  - name: pullRequestNumber
  - name: gitRepository
  - name: gitRevision
  - name: gitCloneDepth
  - name: gitHubCommand
  pipelineruns:
  -  metadata:
        name: tekton-noop-check-$(uid)
        annotations:
          tekton.dev/gitURL: "$(tt.params.gitRepository)"
[...]

PR #1030, opened 2021-04-02
resources: Fail on non-string template labels.

wlynch:
# Changes

Although ObjectMeta defines labels as a map[string]string, Unstructured
types allow for map[string]interface{}, which means that users can set
non-string values like ints in their configurations, and we will parse it
with no issues.

We discovered that the behavior of
[unstructured.GetLabels](https://github.com/kubernetes/apimachinery/blob/dd12c7a65e7f49c4e00b5bd5eb93de2fcbf7831d/pkg/apis/meta/v1/unstructured/unstructured.go#L395)
will silently ignore errors, returning an empty label set if a
non-string value is included. When used in triggers, this looks like we
are completely ignoring the user provided labels.

This change replaces GetLabels with it's underlying implementation, and bubbles
back the non-string error if it is encountered.

# Submitter Checklist

review them:

- [n/a] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#docs) (if user facing)

# Release Note

```release-note
NONE
```


dibyom:
Nice -- one thing I was thinking was that we could do a similar check in the resourceTemplate validation as well...so that if a user creates a resource with the wrong type, we catch it at creation time vs run time. What do you think?

/approve

wlynch:
> Nice -- one thing I was thinking was that we could do a similar check in the resourceTemplate validation as well...so that if a user creates a resource with the wrong type, we catch it at creation time vs run time. What do you think?

That's probably going to be more work than I want to commit to for this PR. 😅 

TriggerResourceTemplates are currently stored as `runtime.RawExtension`s. We likely want to convert this to an `unstructured.Unstructured`, and I'm not sure what the scope of this change would entail (particularly in ensuring backwards compatibility). Alternatively, we'd could unmarshal the message in the webhook and in the controller, but that seems wasteful.

Worth noting - if we move forward with https://github.com/tektoncd/triggers/issues/697 this validation should happen automatically.

dibyom:
Ahh, missed the runtime.RawExtension vs unstructured.Unstructured difference. I don't think it would be too bad converting them in the controller/webhook but I agree that is outside the scope of this PR.
/lgtm

PR #1031, opened 2021-04-02
Add a new ClusterInterceptor to saves the incoming HTTP requests to Results

yaoxiaoqi:
# Changes

This PR adds a ClusterInterceptor called `results`. It has 2 optional parameters:
APIAddr and Token. Both parameters are used to set up Results client. `results` capture
and save incoming requests to Results database if added to Eventlistener.

/kind feature

# Submitter Checklist

review them:

# Release Notes

```release-note
Add a new ClusterInterceptor to saves the incoming HTTP requests to Results
```


yaoxiaoqi:
I think this PR should /hold because the versions of client-go that triggers relies on and results relies on are different.
results is using 0.17.6 while triggers is using 0.19.7
Due to the difference, I had some trouble when setting up a fake results client in the test file because the create functions from these 2 versions have different parameters.
I think some change should be made in results before merging this pr, like updating the client-go version to 0.19.7.

pkg/interceptors/results/results.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	p := triggersv1.ResultsInterceptor{}
+	if err := interceptors.UnmarshalParams(r.InterceptorParams, &p); err != nil {
+		return interceptors.Failf(codes.InvalidArgument, "failed to parse interceptor params: %v", err)
+	}
+	if p.APIAddr == "" {
+		p.APIAddr = localAPIAddr
+	}
+	if w.ResultClient == nil {
```
jmcshane, line 84: the way that this is implemented, once the ResultClient is set on the interceptor service, it won't ever reconfigure the API server address or token. Given you have these as a parameter for triggers to set, i don't think the resultclient should be a property of the interceptor itself.

PR #974, opened 2021-02-26
Begin to address metrics in namespaced eventlistener

jmcshane:
# Changes

Adding metrics endpoint for eventlistener deployment

Closes #736 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#tests) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#docs) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commits)
- [x] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/triggers/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

```release-note
Add metrics for eventlistener deployment
```

PR #1042, opened 2021-04-10
Interceptor SDK

### LINKED ISSUE #1033: Interceptors: Library to make it easier to author interceptors
### Feature request

We should make it easy to do common things for Interceptor authors in Go so that they can just focus on writing business logic (similar to [knative/pkg/webhook](https://github.com/knative/pkg/tree/main/webhook))

Some functionality that this library can have 
* Setting up web server with sensible defaults (users only implement a handler)
* Helper functions to convert errors to InterceptorResponse.Statuses (e.g. https://pkg.go.dev/google.golang.org/grpc/status#FromError)
* (in the future) auto TLS support using something like Knative pkg's [certificates controller](https://github.com/knative/pkg/tree/main/webhook/certificates)
* (in the future) openAPI validation for Interceptor params or config

### LINKED ISSUE #1041: Add readiness check for core-interceptors deployment
### Feature request

Interceptors deployment should have readiness probe so that restarts on that pod do not cause issues with resolving the HTTP request.

### Use case

Expect all triggers core components to have probe endpoints that can be used to validate pod health

jmcshane:
# Changes

Creating interceptor sdk in tree and using it for core interceptors

Closes #1033 
Closes #1041 

# Submitter Checklist

review them:

# Release Notes

```release-note
Creating interceptor SDK to simplify the process of creating new interceptors
```

### DIFF
--- a/docs/clusterinterceptors.md
+++ b/docs/clusterinterceptors.md
@@ -58,3 +58,46 @@ The Kubernetes object running the custom business logic for your `ClusterInterce
 - Returns an HTTP 200 OK response that contains an [`InterceptorResponse`](https://pkg.go.dev/github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1#InterceptorResponse) 
   as a JSON body
 - Returns a response other than HTTP 200 OK only if payload processing halts due to a catastrophic failure
+
+## Interceptor SDK
+
+The interceptor sdk allows you to easily create new interceptors for use with Tekton. The main SDK
+function can be found in [pkg/interceptors/sharedmain.go](pkg/interceptors/sharedmain.go). Invoke the
+function `sdk.InterceptorMainWithConfig` passing in a map with a key of the interceptor HTTP request
+path and the value as an object that implements the `InterceptorInterface`.
+
+The `Process` function of the `InterceptorInterface` is where you should implement the logic around
+your interceptor HTTP request execution. The second function `InitializeContext` passes a context
+that can be used to retrieve cancellation events from the top level HTTP server as well
+as retrieving the `*zap.SugaredLogger` and `Kubernetes.Interface` objects. To retrieve these from the
+context, use the following methods:
+
+```
+import (
+  "knative.dev/pkg/logging"
+  "github.com/tektoncd/triggers/pkg/interceptors"
+)
+
+func (w *Interceptor) InitializeContext(ctx context.Context) {
+	w.Logger = logging.FromContext(ctx)
+	w.KubeClientSet = interceptors.GetKubeClient(ctx)
+}
+```
+
+If any of these functions are added to the resource type, the interceptor SDK will pass
+the values from the main interceptor initialization to the downstream resource. To see
+how these values are configured, see [`pkg/interceptors/sharedmain.go`](pkg/interceptors/sharedmain.go).
+
+To confirm that your interceptor appropriate implements the `InterceptorInterface`, at the top of your
+interceptor file, you can add the following:
+
+```
+import (
+  	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
+)
+
+var _ triggersv1.InterceptorInterface = (*Interceptor)(nil)
+```
+
+This is done at the top of each of the core interceptors as an example.
+
--- a/pkg/apis/triggers/v1alpha1/interceptor_types.go
+++ b/pkg/apis/triggers/v1alpha1/interceptor_types.go
@@ -14,6 +14,8 @@ type InterceptorInterface interface {
 	// see if processing should continue and InterceptorResponse.Status.Code to distinguish between the kinds of errors
 	// (i.e user errors vs system errors)
 	Process(ctx context.Context, r *InterceptorRequest) *InterceptorResponse
+
+	InitializeContext(ctx context.Context)
 }
 
 // Do not generate DeepCopy(). See #827
--- a/cmd/interceptors/main.go
+++ b/cmd/interceptors/main.go
@@ -17,75 +17,26 @@ limitations under the License.
 package main
 
 import (
-	"context"
-	"fmt"
-	"log"
-	"net"
-	"net/http"
-	"time"
-
-	"github.com/tektoncd/triggers/pkg/interceptors/server"
-	"go.uber.org/zap"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
-	"knative.dev/pkg/logging"
+	"github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
+	"github.com/tektoncd/triggers/pkg/interceptors"
+	"github.com/tektoncd/triggers/pkg/interceptors/bitbucket"
+	"github.com/tektoncd/triggers/pkg/interceptors/cel"
+	"github.com/tektoncd/triggers/pkg/interceptors/github"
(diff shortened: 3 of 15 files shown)


jmcshane:
The `interceptor/sdk/sharedmain.go` file is templated based off the knative pkg sharedmain: https://github.com/knative/pkg/blob/main/injection/sharedmain/main.go. I do wonder if some of this code could be removed in favor of referencing that package directly

jmcshane:
> The interceptor/sdk/sharedmain.go file is templated based off the knative pkg sharedmain: https://github.com/knative/pkg/blob/main/injection/sharedmain/main.go. I do wonder if some of this code could be removed in favor of referencing that package directly

This main does not return until the webhook and controller context closes. We'll have to stick with the current method of picking up subsections of the sharedmain `MainWithConfig` into the SDK main.

afrittoli (commented):
I have not reviewed the code in details, but I think this is a great idea, having a common SDK no only helps with development of new interceptors, it also helps with consistency in configuration, logging and operations - which is great for Tekton operators!

jmcshane:
An idea that's been bouncing around in my head is the concept of extending what can be "done" with interceptors. This is in line with what we've discussed with nested triggers, but there may be some easy things we can add to the interceptor API to get some of the same benefits. Currently, triggers do two things:

* Determine if processing should continue
* Add data to the body

Some ideas here are:

* Trigger processing could iterate a result map or list key from an extension. For example, a single `push` event that should trigger multiple pipelines on a repository
* Have triggers that terminate without template/bindings - the use case here is some gitops activities that just rely on the source system APIs. I'd love to be able to implement something like Tekton's `/lgtm` as a set of interceptors.

In order to change the behavior of interceptors on the triggers side, we would need to make changes in the `eventlistenersink` processing, rather than the interceptors themselves. I can write up more about this separately (the SDK is just about building the interceptor side of this processing) to discuss in a larger group.

jmcshane:
One challenge with this implementation is that you have to return an interface from the New function. Generally, in go, the desire is to return objects and accept interfaces.

I'll have to take a look at refactoring this a bit and managing the types a little more cleanly

dibyom:
Overall this looks great! Couple of minor things I can think of:

1. Do we want to call the package sdk? Or can we just use the interceptors as the package name?
2. The InterceptorMainWithConfig function takes in a `map[string]func(kubernetes.Interface, *zap.SugaredLogger) v1alpha1.InterceptorInterface` This might be too tied to the current core interceptors implementation? What if we need to pass in more arguments to initialize the interceptor. And like you already mentioned, it would be nice to return an actual type vs the interface.

jmcshane:
1. There was an existing `interceptors` package, so I moved this down into sdk. I can move this stuff up into the top level package.
2. I did think about this quite a bit, here's where I'm at on this:
   * if there are other resources that a specific interceptor needs, they can invoke them in the NewInterceptor function. This is already done by the webook interceptor here: https://github.com/tektoncd/triggers/blob/main/pkg/interceptors/webhook/webhook.go#L55
   * Since some configs could be provided in main, we could modify the InterceptorInterface to have two functions: The current `Process` and then an `Initialize` that takes the kube client and the logger. It makes sense for the SDK to initialize these two resources (since the kube client gets loaded with the context and the logger sets up its watch to the logger config map), but the downstream interceptor should consume them for its execution (or ignore them).
   * What this two function interface would allow us to do is say that the interceptor Main function accepts an implementation object that implements the new InterceptorInterface, then calls Initialize before sending any Process InterceptorRequests.

jmcshane:
One challenge that this Initialize method presents is it does move away from the golang standard "New" function. The `NewController` function in each of the current interceptors allows each interceptor to decide what is necessary to initialize the controller. For a more generic approach, we could approach it via reflection like the waypoint project did with their plugin-sdk: https://www.waypointproject.io/docs/extending-waypoint/plugin-interfaces/default-parameters. 

With this approach, the interceptor SDK could invoke a "New" function via reflection and provide a number of parameters that could be used by custom interceptors. The challenge here is I haven't seen an interceptor that would use anything _besides_ the kube client and a logger. A metrics resource may make sense (reworking #974) is on my personal backlog, so that custom interceptors can expose their own metrics. As we move towards beta, we will need interface stability, so adding new resources to an Initialize function would be a challenge. There is significant complexity in how a reflection based implementation would work, so we'd have to decide whether that complexity is worth the flexibility that reflection would provide.

jmcshane:
The sharedmain has a single function that blocks return on a context it creates itself. I modeled it off the webhook/controller sharedmain in knative which does not have any unit tests either

jmcshane:
@dibyom I incorporated your requested changes as well as changing the type of the objects that you pass in. Check out cmd/interceptors/main.go to see how this is implemented now, I think its fairly clean.

The way I did "customization" of the interceptor object was by adding a few additional interface types in the SDK package. If the interceptor implements the target interface, it will invoke the defined function with the object that it has.

In this way, the SDK can provide any of the values that it initializes to the interceptor object, which can then create any additional Kube clients or other objects before initialization. An obvious extension of the current interfaces will be when we have a metrics recorder to pass to the interceptor type so it can record metrics about the behavior of the interceptor plugin. Again, all of these things are optional, so it is easy to add new ones without breaking old functionality.

jmcshane:
@afrittoli I believe this latest update should align with your goals as well, because we will provide additional operational capabilities via these optional interfaces. As more are added, each interceptor can optionally pick up the change (without breaking previous iterations).

jmcshane:
The reduction in test "coverage" is just due to the fact that cross package test coverage is not recorded. Those lines are all invoked in the interceptor_test package in pkg/sink.

docs/clusterinterceptors.md
```diff
@@ -15,6 +15,7 @@ top-level `extensions` field to other `Interceptors` and `ClusterInterceptors` c
 
 ## Structure of a `ClusterInterceptor`
 
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
jmcshane: Removed and rebased code

jmcshane:
Updating to pass arguments via ctx

pkg/apis/triggers/v1alpha1/interceptor_types.go
```diff
@@ -14,6 +14,8 @@ type InterceptorInterface interface {
 	// see if processing should continue and InterceptorResponse.Status.Code to distinguish between the kinds of errors
 	// (i.e user errors vs system errors)
 	Process(ctx context.Context, r *InterceptorRequest) *InterceptorResponse
+
+	InitializeContext(ctx context.Context)
```
dibyom, line 18: What do you think of changing each interceptor so that in the process func itself we do something like this: `logger := logging.FromContext(ctx)`


The idea is that the context might contain request specific info (e.g. eventID/requestID).

jmcshane: that makes sense, ill update the interceptors and the documentation around how this should work.

do we do the same thing with the kube client? then we can just remove this function altogether from the interface

dibyom: yup! exactly! 

https://github.com/tektoncd/triggers/blob/4939226120cdf4646e1f5941fb1b10b8f866abac/vendor/knative.dev/pkg/client/injection/kube/client/client.go

afrittoli:
@jmcshane @dibyom is this still going to happen?

dibyom:
> @jmcshane @dibyom is this still going to happen?

Yes :)

PR #1043, opened 2021-04-11
Creating rego interceptor implementation

### LINKED ISSUE #484: Add Rego interceptor
# Expected Behavior

In addition to supporting CEL for filtering and mutation of payloads, we can support other options out of the box, such as Rego.

# Actual Behavior

We only support CEL, tho folks can bring their own interceptors of course!

# Additional Info

To actually make this happen we should propose a design since there are a few ways we could express filtering + overlays with Rego.

[This is a POC of a Rego interceptor](https://github.com/tektoncd/triggers/compare/master...bobcatfish:rego_interceptor). It uses a `filter` query and an `overlay` query which provides the same functionality as [this example CEL interceptor](https://github.com/tektoncd/triggers/blob/0ea7560954be72b176b4ce199083404da4b9a0bb/examples/eventlisteners/cel-eventlistener-interceptor.yaml#L9-L14).

Preview of the rego syntax:

```yaml
  triggers:
    - name: rego-trig-with-matches
      interceptors:
        - github:
            secretRef:
              secretName: webhook-secret
              secretKey: secret
        - rego:
            source: |
              package main
              default filter = false
              filter {
                input.header[k][i] == "pull_request"
                lower(k) == "x-github-event"
              }
              
              default overlay = {}
              overlay = {
                "extensions": {
                   "truncated_sha": substring(input.body.pull_request.head.sha, 0, 7),
                }
              }
      bindings:
      - name: rego-binding
      template:
        name: rego-trigger-template
```

[This yaml](https://github.com/bobcatfish/triggers/blob/8c74714382dc2f3610f60ea3e96a507882104a18/examples/eventlisteners/rego-eventlistener-interceptor.yaml) is an example that uses [the pullrequest task](https://github.com/tektoncd/catalog/tree/v1beta1-wip/pullrequest) to comment on a PR when it is opened.

jmcshane:
# Changes

Closes #484 

Demo implementation of rego interceptor using some dynamic parameter capabilities

# Submitter Checklist

review them:

# Release Notes

```release-note
Create rego interceptor
```


jmcshane:
Docs would be done if this implementation was determined appropriate to move forward

jmcshane:
@afrittoli not sure what to do here because the license files for these two dependencies are named `LICENSE-APACHE-2.0.txt`. Could use some help from plumbing to resolve

jmcshane:
This really needs some design consideration before moving the implementation in tree. If there was an `interceptors` repository, we wouldn't have to do all the interceptors in tree and we could document "custom" interceptors separately from the main triggers project.

Here's some things that I've mapped out:
* We need some standard way to query the policy to find out if the interceptor should "continue" processing. You can see some ways to determine "truth" here: https://www.openpolicyagent.org/docs/latest/policy-testing/. The question is whether there should be some magic query hidden in the interceptor implementation, or just allow consumers to specify the module and query together.
* The query statement for filter must return a boolean value for us to be able to determine whether the interceptor should continue
* By itself, a filter query using rego could be a great feature (it is a policy language after all). Still, it makes sense that there would be a way to specify extensions. In my mind, this can be done using additional queries that are paired with extension "keys". You can see how this is implemented within the PR (see `pkg/interceptors/rego`). I am not a huge fan of the current design, but you can take a look at the rego go-api to see how these additional resources could be queried: https://pkg.go.dev/github.com/open-policy-agent/opa/rego#example-Rego.Eval-MultipleBindings

jmcshane:
Due to license issues, this is going to be challenging for awhile. I think we just assume this will be stuck in experimental for some time.

Issue #385, opened 2020-01-24
Nested TriggerBinding expressions $($())


dibyom:
Following the discussion in https://github.com/tektoncd/triggers/pull/376#discussion_r370209602
let's discuss what the behavior should be for nested Tekton expressions in TriggerBindings

## Current documented behavior

- Nested expressions are allowed. However, only the text in innermost `$()` is evaluated as JSONPath, the rest is treated as regular strings. From the [docs](https://github.com/tektoncd/triggers/blob/master/docs/triggerbindings.md#event-variable-interpolation):

If the `$()` is embedded inside another `$()` we will use the contents of the innermost
`$()` as the JSONPath expression

```shell script
$($(body.b)) -> $(body.b)
$($($(body.b))) -> $(body.b) 
```

## Questions
- Is there an actual use case for this?
- Should we just return an error instead of the behavior above?

vtereso:
In this format, it seems like we are allowing for the inner value to be resolved recursively rather than just taking the text. I think an error is probably better?

dibyom:
This is another "unintentional" feature/bug that we should fix

dibyom:
Still a bug that we need to fix :)

jmcshane:
@dibyom what is the "correct" behavior here? has that ever been decided?

dibyom:
Leaning towards just making this an error.

PR #1049, opened 2021-04-14
Provide error for invalid bindings

jmcshane:
# Changes

TriggerBinding values do not yet have any validation on their values, but we have situations, as
discussed in #385, where the binding value is somewhat nonsensical. In templates, it is possible that
a nesting situation may take place within a TaskRun script, but for a binding parameter value it makes
sense to avoid these types of situations.

This PR currently only avoids the situations where expressions are directly nested, meaning
dollar-parens immediately followed by dollar-parens. This would not detect a situation where they are
separated by some characters

# Submitter Checklist

review them:

# Release Notes

```release-note
TriggerBinding error on invalid Tekton template values
```

### DIFF
--- a/pkg/apis/triggers/v1alpha1/trigger_binding_validation.go
+++ b/pkg/apis/triggers/v1alpha1/trigger_binding_validation.go
@@ -18,6 +18,8 @@ package v1alpha1
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/apis"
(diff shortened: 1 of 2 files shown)


dibyom:
Looks like build test is failing with: 

```
pkg/template/jsonpath.go:187:11: `if` block ends with a `return` statement, so drop this `else` and outdent its block (golint)
			} else {
			       ^
```

jmcshane:
So, this happens at processing time, but I wonder if we want to change this to a validation error? It might be an issue in existing installs, but an error at triggerbinding creation time might make more sense than during the HTTP request.

dibyom:
> So, this happens at processing time, but I wonder if we want to change this to a validation error? It might be an issue in existing installs, but an error at triggerbinding creation time might make more sense than during the HTTP request.

Agreed. That makes more sense.

jmcshane:
Updated tests and Validation function (as well as validation error messages). The HTTP request processing remains completely the same, we will just reject binding values that have nested tekton expressions within them.

pkg/apis/triggers/v1alpha1/trigger_binding_validation.go
```diff
@@ -18,29 +18,71 @@ package v1alpha1
...
+func validateParamValue(in string) *apis.FieldError {
+	if !strings.Contains(in, "$(") {
+		return nil
+	}
+	// Splits string on $( to find potential Tekton expressions
+	maybeExpressions := strings.Split(in, "$(")
+	terminated := true
+	for _, e := range maybeExpressions[1:] { // Split always returns at least one element
+		// Iterate until we find the first unbalanced )
+		numOpenBrackets := 0
```
dibyom, line 63: question: is it possible to reuse this bit from the jsonpath.go? Maybe this bit could be extracted to its own function? If that's too much for this PR, I'm happy to merge this as is.

dibyom (approved):

Issue #931, opened 2021-01-27
HTTP Response StatusCodes from EventListener


dibyom:
# Background

Today the EventListener waits for at least one resource to be created before returning a HTTP response.  We return 201 if a resource is created and 202 otherwise.  If we fail to create a resource due to an auth error, we return a 401 response instead.

Note that the 201 status code, the user still may not know exactly what resource was created. They'd still have to look into the EL logs (using the eventID from the HTTP response) to find out which resource was created.

# Problem

1. If an EL is handling many triggers (or a single trigger had many interceptor calls), the EL will take a long time to respond and may even time out. The webhook originator (e.g. GitHub) don't generally care about the response codes (unless its 5xx)

1. The 201 status code gives the impression that everything succeeded while that may not be the case always e.g. one trigger might succeed while another failed (see #656, #758)

# Proposal

I'd propose we return a response back as soon as the EL has read the body. We'd return a 2xx if we can fully read the body as a valid JSON document, and 4xx otherwise. We'd still return an eventID (and the name and namespace of the EventListener) so that the user can track the progress of the trigger.

dibyom:
We are going to publicize the status change a bit and then put it in a release.

PR #1052, opened 2021-04-14
Initial implementation for trigger groups

jmcshane:
# Changes

Initial demo of using trigger groups to process eventlistener requests

# Submitter Checklist

review them:

# Release Notes

```release-note
Implement triggerGroups and interceptor chains
```


jmcshane:
@dibyom @afrittoli @wlynch this implementation is now ready in line with our discussion for [TEP-0053](tektoncd/community#386). I reduced the changes that are a part of this PR to the absolute simplest possible, including maintaining the response code validation. The key reason I kept the response codes in here for now is because of the huge test surface that depends on this response code and expects the HTTP function to block until all resources are created.

I think it makes sense to open a separate issue to track discussion on how we manage the HTTP status code. For now, this PR is ready from a code perspective. I think what we need to finish it are:

- [ ] Documentation about triggerGroups in the eventlistener document
- [ ] Example eventlistener configuration including triggergroups

Please let me know if there is anything else you would like to see in this PR.

dibyom:
> The key reason I kept the response codes in here for now is because of the huge test surface that depends on this response code and expects the HTTP function to block until all resources are created.

Sounds good. We can tackle that separately as part of #931

pkg/sink/sink.go
```diff
@@ -95,49 +95,11 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
-		if err != nil {
-			r.Logger.Errorf("Error getting Triggers: %s", err)
-			response.WriteHeader(http.StatusInternalServerError)
-			return
-		}
-		trItems = append(trItems, trList...)
+	trItems, err := r.selectTriggers(el.Spec.NamespaceSelector, el.Spec.LabelSelector)
+	if err != nil {
+		r.Logger.Errorf("Unable to select configured triggers: %s", err)
+		response.WriteHeader(http.StatusInternalServerError)
```
dibyom, line 101: Should we return some info in the HTTP response as well?

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
```diff
@@ -699,6 +722,36 @@ func TestEventListenerValidate_error(t *testing.T) {
...
+						Params: []v1alpha1.InterceptorParams{{
+							Name:  "filter",
+							Value: test.ToV1JSON(t, "has(body.repository)"),
+						}},
+					}},
+				}))),
+	}, {
+		name: "triggerGroup requires interceptor",
+		el: bldr.EventListener("name", "namespace",
+			bldr.EventListenerSpec(
```
dibyom, line 744: One request: Could we use regular structs instead of the builders? We have been trying to get reduce/remove test builder usage for a while -- See https://github.com/tektoncd/pipeline/issues/3178 for some discussion

dibyom (commented):
Haven't gone through all of the code yet. But yes, docs and examples would be the two remaining things.
Also, we should update the release notes section of the PR to something a bit more descriptive -- (Adds a way to group common interceptors across multiple triggers ...)

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
jmcshane: Aha, I wasn't aware that this was the goal. I will refactor

jmcshane: Aha, I wasn't aware that this was the goal. I will refactor

pkg/sink/sink.go
jmcshane: What info are you considering here to return besides the downstream error message?

savitaashture:
PR handled as part of this https://github.com/tektoncd/triggers/pull/1232

PR #1075, opened 2021-05-06
Update Getting Started README.md

EarthlingDavey:
# Changes

Add a step to the getting started guide so that admin-role.yaml is all within the getting-started namespace.

Without this step, the developer will get the assign a role to a ServiceAccount in the default namespace. Without the change, further in the guide, the event listener this throw errors:

`User "system:serviceaccount:getting-started:tekton-triggers-example-sa" cannot list resource "clustertriggerbindings"`

`User "system:serviceaccount:getting-started:tekton-triggers-example-sa" cannot list resource "clusterinterceptors"`

# Submitter Checklist


jmcshane:
@EarthlingDavey should we just update the role to have the correct namespace in it? Is the getting-started used for anything else besides this example?

EarthlingDavey:
@jmcshane , I'm not sure. I'm new to this project.

As it stands now, the file `./docs/getting-started/rbac/admin-role.yaml` is a yaml shortcut to `examples/rbac.yaml`.

I assumed that `examples/rbac.yaml` may be subject to frequent(ish) changes.
Having a copy at `./docs/getting-started/rbac/admin-role.yaml` with just one line changed might be a pain to maintain.

If you prefer a copy/paste with the corrected namespace then, I can do it and make a PR.
Please advise.

dibyom:
(Also, could you please sign the CLA?)

EarthlingDavey:
> What if we just add a new file just for the clusterrolebinding bit? 

This sounds fine to me.

I think the choices are one of the following.

- split `examples/admin-role.yaml`. Users apply each one, except for a `getting-started/rbac/cluster-role-binding.yaml`
- users could apply the complete file `examples/admin-role.yaml` as it is. Then apply a patch `getting-started/rbac/cluster-role-binding.yaml` in a following step.

I've signed the CLA btw.

EarthlingDavey:
Cool, I'm not too familiar with this project's processes... should I wait for a while for further consideration?

Let me know, and I'll either make the change or hold off.

dibyom:
> Cool, I'm not too familiar with this project's processes... should I wait for a while for further consideration?

I think its fine to go ahead!

EarthlingDavey:
I've made the suggested changes. But integration tests are failing with an error and warning related to my change.

I'm no go dev. and earning about the project's testing is too much for me to take on 😅 Perhaps the test side of things could be handled by someone more familiar with the project.

PR #1077, opened 2021-05-07
Update status code handling

### LINKED ISSUE #931: HTTP Response StatusCodes from EventListener

### LINKED ISSUE #932: Status Conditions for EventListener
# Background

Currently, the EventListener's `status` field contains following 4 condition types - `Available`, `Deployment`, `Progressing`, `Service`.  
The `Deployment` and `Service` condition indicate whether the deployment or service exists.  `Available` and `Progressing` are conditions propagated from the underlying Deployment's conditions.

# Problems
1. 4 conditions are a lot and can be confusing -- which conditions is most relevant? Usually the Available one since that will tell you if the deployment is actually available or not. But in that case, why do we need the other 3?

2. Two of them are of type "Deployment",  and "Service". DeploymentExists and ServiceExists are probably more explicit.

3. With TEP-008, we will support Knative services as an alternate deployment model for the EventListeners. With that the existing 4 conditions do not make as much sense.

# Proposal
1. Add a new condition of type `Ready`. It will be true if both service/deployment are up and available. We can use the `reason` and `message` fields to add details on why the EL is not ready e.g. ServiceNotAvailable, DeploymentProgressing etc.

2. Deprecate the 4 older conditions and remove them when we move to the v1beta1 APIversion.

# References:
1. [Knative Serving API spec](https://knative.dev/docs/serving/spec/knative-api-specification-1.0/#error-signalling) says: `Each resource MUST have either a Ready condition (for ongoing systems) `

1. K8s API [conventions on conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)

jmcshane:
# Changes

Closes #931

Removes status code handling in HTTP requests for triggers. Replacing with waitgroup that counts current number of triggers that are being processed by the Sink.

# Submitter Checklist

review them:

# Release Notes

```release-note
Remove status code processing for EventListener HTTP request
```

### DIFF
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sync"
 
 	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
 	triggersclientset "github.com/tektoncd/triggers/pkg/client/clientset/versioned"
@@ -34,7 +35,6 @@ import (
 	"github.com/tektoncd/triggers/pkg/template"
 	"github.com/tidwall/sjson"
 	"go.uber.org/zap"
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	discoveryclient "k8s.io/client-go/discovery"
@@ -55,6 +55,7 @@ type Sink struct {
 	Logger                 *zap.SugaredLogger
 	Recorder               *Recorder
 	Auth                   AuthOverride
+	WaitGroup              *sync.WaitGroup
 
 	// listers index properties about resources
 	EventListenerLister         listers.EventListenerLister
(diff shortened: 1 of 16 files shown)


dibyom:
Since this is a breaking change, should we announce it and give users more of a heads up before releasing it?

jmcshane:
> Since this is a breaking change, should we announce it and give users more of a heads up before releasing it?

Yeah, im not sure how those announcements are normally done. We can discuss in WG next week, or potentially API working group to talk about how these types of behavior updates should be managed between the maintainers and the end user community.

We saw a similar thing happen with the install for interceptors for v0.13 and there were a few people that popped in the Slack to discuss the change. I think the best course of action will be to have a note about the change in behavior in the release notes and potentially have some more thorough documentation in the eventlistener doc about status code behavior.

pkg/sink/sink_test.go
```diff
@@ -920,173 +916,6 @@ func TestHandleEvent(t *testing.T) {
...
-const userWithPermissions = "user-with-permissions"
-const userWithoutPermissions = "user-with-no-permissions"
-const userWithForbiddenAccess = "user-forbidden"
-
-var triggerAuthWG sync.WaitGroup
-
-type fakeAuth struct {
-}
-
-func (r fakeAuth) OverrideAuthentication(sa string, _ string, _ *zap.SugaredLogger, defaultDiscoverClient discoveryclient.ServerResourcesInterface,
```
dibyom, line 933: FYI @gabemontero @khrm With this change, we are no longer returning 4xx errors for unauthorized resource creation. Does this impact anything you are working on?

gabemontero: No impacts on my end.

pkg/sink/sink.go
```diff
@@ -147,45 +148,17 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
 		return
 	}
-	result := make(chan int, 10)
 	// Execute each Trigger
+	r.WaitGroup.Add(len(triggers))
 	for _, t := range triggers {
 		go func(t triggersv1.Trigger) {
+			defer r.WaitGroup.Done()
 			localRequest := request.Clone(request.Context())
-			if err := r.processTrigger(t, localRequest, event, eventID, eventLog); err != nil {
```
gabemontero, line 155: So yeah it still fails.  The user presumably learns of it from some other means.  Full disclosure, I did not research *how* that might now be or not be achieved.

If there is an easy enough reader's digest explanation @jmcshane @dibyom please elaborate.  Or if a pointer to what I have to read to understand makes more sense, that works too.

thanks

gabemontero (commented):
Sorry for the delay was out of the office for a few days

pkg/sink/sink.go
dibyom: I think the diff for the docs/eventlisteners.md page shows the change: https://github.com/tektoncd/triggers/pull/1077/files#diff-7da54b4f5484ec784351290e106d6aebea551a17046b6116cc8e65964ebb693f

i.e instead of waiting for a resource to be created and then responding with a 201 status code, Triggers will now return a 202 response earlier (i.e. when it has selected the triggers to process -- it will no longer wait for the trigger processing to complete)

gabemontero: OK so yeah that markdown, assuming basic HTTP rc knowledge, makes it clear that the processing here is now asynchronous vs. synchronous.

With all the scalability advantages, etc. obtained with that shift.

In theory then the poster of the event then needs to know what was triggered by the event and then query those artifacts directly for status.

Quick code scan / memory refresh and I see the `/triggers-eventid` label set on the objects created (which we have had for a while) facilitates that query.

Would a cross reference to https://github.com/tektoncd/triggers/blob/4b9aa3157f3989255f350c8c60c2b6aecab38783/docs/triggertemplates.md#structure-of-a-triggertemplate somewhere and its explanation of the label near https://github.com/tektoncd/triggers/pull/1077/files#diff-7da54b4f5484ec784351290e106d6aebea551a17046b6116cc8e65964ebb693f be a useful change in your opinion @dibyom ?

Other than that potential tweak, I'm good.  Thanks.

dibyom: Ok yeah, I think adding a couple of lines or (maybe link to a different section) with the information on querying makes sense.

dibyom:
Closing -- This commit has been migrated in #1132

PR #1080, opened 2021-05-07
Migrate from test builders to structs

dibyom:
# Changes

This is helpful as we migrate to the v1beta1 types since we won't have
to maintain separate test builders for v1beta1 types.

Ref #1067, #735

/kind cleanup
/release-note-none

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#tests) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#docs) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commits)
- [ ] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/triggers/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

PR #1082, opened 2021-05-10
Add a Ready StatusCondition for EventListener

### LINKED ISSUE #932: Status Conditions for EventListener

dibyom:
# Changes

The Ready Status is true when the other 4 status conditions - (Service,
Deployment, Progessing, and Available) are all true. The idea is that in the
future, Ready would be the one Condition users could  use to see if an
EventListener is ready to serve traffic or not.

Fixes #932

# Submitter Checklist

review them:

# Release Notes

```release-note
The EventListener Status now has a new Condition called Ready. Ready is set to True when the other status conditions are also true (i.e. the EventListener is ready to serve traffic). It is false if any of the other statuses are false.

DEPRECATION NOTICE: In the future, we plan to deprecate the other status conditions in favor of the Ready status condition.
```

### DIFF
--- a/pkg/apis/triggers/v1alpha1/event_listener_types.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_types.go
@@ -177,7 +177,10 @@ const (
 	ClusterTriggerBindingKind TriggerBindingKind = "ClusterTriggerBinding"
 )
 
-var eventListenerCondSet = apis.NewLivingConditionSet(ServiceExists, DeploymentExists)
+var eventListenerCondSet = apis.NewLivingConditionSet(
+	ServiceExists,
+	DeploymentExists,
+)
 
 // GetCondition returns the Condition matching the given type.
 func (els *EventListenerStatus) GetCondition(t apis.ConditionType) *apis.Condition {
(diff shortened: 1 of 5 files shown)


savitaashture:
@dibyom as per the issue we will be deprecating Deployment status and keeping only `READY` state
So could you just update the Release notes to point this info

pkg/apis/triggers/v1alpha1/cluster_trigger_binding_validation_test.go
```diff
@@ -30,23 +30,49 @@ func Test_ClusterTriggerBindingValidate(t *testing.T) {
...
+					Value: "$(body.input1)",
+				}, {
+					Name:  "param2",
+					Value: "$(body.input2)",
+				}, {
+					Name:  "param3",
+					Value: "$(body.input3)",
+				}},
+			},
+		},
```
savitaashture, line 56: I see this PR mainly focuses on status related changes
So can we move these changes to [Migrate from test builders to structs](https://github.com/tektoncd/triggers/pull/1080) ?

dibyom: yeah this PR builds on the other PR. so once that is merged, I'll rebase. Otherwise, there were a bunch of conflicts.

dibyom:
> Also do we need to do anything for these https://github.com/tektoncd/triggers/blob/main/config/300-eventlistener.yaml#L59-L64 in this PR 🤔

That's a good catch. I was thinking of keeping it the same way it was until we finish the deprecation. But I can change it if you feel stongly.

dibyom:
The ingress e2e tests passes locally 🤔

savitaashture:
> > Also do we need to do anything for these https://github.com/tektoncd/triggers/blob/main/config/300-eventlistener.yaml#L59-L64 in this PR thinking
> 
> That's a good catch. I was thinking of keeping it the same way it was until we finish the deprecation. But I can change it if you feel stongly.

I think lets keep it for a release as deprecation so that user won't see any surprises if they are dependent on status

savitaashture (approved):

PR #1086, opened 2021-05-11
Fix metrics port expose issue for triggers eventlistener

savitaashture:
# Changes

Metrics related changes introduced as part of https://github.com/tektoncd/triggers/pull/1061 and because of those changes we are facing below issues when we create Knative Service using customeResources

1. when we apply [custom-resource](https://github.com/tektoncd/triggers/tree/main/examples/custom-resource) example  EL fail with below error
```
Warning  InternalError    15s (x14 over 56s)  EventListener  admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: More than one container port is set: spec.template.spec.containers[0].ports
Only a single port is allowed
```
2. Knative Serving adds `queue-proxy` container along with EL container and Knative itself handles metrics on `9090` port for queue-proxy and because of that we face already bind error
```
{"severity":"ERROR","timestamp":"2021-05-11T09:28:00.588588485Z","logger":"queueproxy","caller":"queue/main.go:228","message":"Failed to bring up queue-proxy, shutting down.","commit":"813aa65","knative.dev/key":"dtest/el-github-knative-listener-00001","knative.dev/pod":"el-github-knative-listener-00001-deployment-fc8fb5cbc-xbqqv","error":"metrics server failed to listen: listen tcp :9090: bind: address already in use","stacktrace":"main.main\n\tknative.dev/serving/cmd/queue/main.go:228\nruntime.main\n\truntime/proc.go:204"}
```

The PR solves above issues and exposes metrics on `9000` port

Fixes https://github.com/tektoncd/triggers/issues/1090

/cc @jmcshane @dibyom 

# Submitter Checklist

review them:

# Release Notes

```release-note
Triggers exposes metrics on 9000 port 
```

### DIFF
--- a/cmd/eventlistenersink/main.go
+++ b/cmd/eventlistenersink/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"strings"
 	"time"
 
 	dynamicClientset "github.com/tektoncd/triggers/pkg/client/dynamic/clientset"
@@ -30,6 +31,8 @@ import (
 	"github.com/tektoncd/triggers/pkg/sink"
 	"github.com/tektoncd/triggers/pkg/system"
 	"go.uber.org/zap"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -83,7 +86,24 @@ func main() {
 	profilingServer := profiling.NewServer(profilingHandler)
 	metrics.MemStatsOrDie(ctx)
 
-	sharedmain.WatchObservabilityConfigOrDie(ctx, configMapWatcher, profilingHandler, logger, EventListenerLogKey)
+	exp := metrics.ExporterOptions{
+		Component: strings.ReplaceAll(EventListenerLogKey, "-", "_"),
+		ConfigMap: nil,
+		Secrets:   sharedmain.SecretFetcher(ctx),
+	}
+	cmData, err := metrics.UpdateExporterFromConfigMapWithOpts(ctx, exp, logger)
+	if err != nil {
+		logger.Fatalw("Error in update exporter from ConfigMap ", zap.Error(err))
+	}
+	if _, err := kubeClient.CoreV1().ConfigMaps(system.GetNamespace()).Get(ctx, metrics.ConfigMapName(),
+		metav1.GetOptions{}); err == nil {
+		configMapWatcher.Watch(metrics.ConfigMapName(),
+			cmData,
+			profilingHandler.UpdateFromConfigMap)
+	} else if !apierrors.IsNotFound(err) {
+		logger.Fatalw("Error reading ConfigMap "+metrics.ConfigMapName(), zap.Error(err))
+	}
+
 	logger.Info("Starting configuration manager...")
 	if err := configMapWatcher.Start(ctx.Done()); err != nil {
 		logger.Fatalw("Failed to start configuration manager", zap.Error(err))
--- a/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
+++ b/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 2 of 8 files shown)


cmd/eventlistenersink/main.go
```diff
@@ -83,7 +86,26 @@ func main() {
 	profilingServer := profiling.NewServer(profilingHandler)
 	metrics.MemStatsOrDie(ctx)
 
-	sharedmain.WatchObservabilityConfigOrDie(ctx, configMapWatcher, profilingHandler, logger, EventListenerLogKey)
+	exp := metrics.ExporterOptions{
+		Domain:         "",
+		Component:      strings.ReplaceAll(EventListenerLogKey, "-", "_"),
+		PrometheusPort: 9000,
```
dibyom, line 92: Would this only work if the metrics backend is Prometheus (and not one of the other supported types?)

cmd/eventlistenersink/main.go
```diff
@@ -83,7 +86,26 @@ func main() {
 	profilingServer := profiling.NewServer(profilingHandler)
 	metrics.MemStatsOrDie(ctx)
 
-	sharedmain.WatchObservabilityConfigOrDie(ctx, configMapWatcher, profilingHandler, logger, EventListenerLogKey)
+	exp := metrics.ExporterOptions{
+		Domain:         "",
```
dibyom, line 90: I don't know a ton about metrics but from reading the docs for ExporterOptions, it says that this field "Must be present". Is "" as valid value?

dibyom:
I tried to quickly dig down into where we were actually setting up the metrics server port to see if there was an easier way to do this but failed.  One idea: Would it be possible to just set the port on the [defaultObservabilityConfig()](https://github.com/tektoncd/triggers/blob/ba08d13c21ed7fef9c10bc973b08ec9d1959aebc/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go#L989)

cmd/eventlistenersink/main.go
dibyom: Ok Ignore this. I did not know what I was talking about.

dibyom:
How about something like this:

```
	updateFunc, err := metrics.UpdateExporterFromConfigMapWithOpts(ctx, exp, logger)
	if err != nil {
		logger.Fatal("Failed to create metrics exporter update function", zap.Error(err))
	}
	configMapWatcher.Watch(metrics.ConfigMapName(), updateFunc)
```

cmd/eventlistenersink/main.go
savitaashture: You are right 
if its empty by default it takes value from here https://github.com/tektoncd/triggers/blob/main/vendor/knative.dev/pkg/metrics/exporter.go#L123-L125 which internally takes value from env `METRICS_DOMAIN`

savitaashture:
> I tried to quickly dig down into where we were actually setting up the metrics server port to see if there was an easier way to do this but failed. One idea: Would it be possible to just set the port on the [defaultObservabilityConfig()](https://github.com/tektoncd/triggers/blob/ba08d13c21ed7fef9c10bc973b08ec9d1959aebc/pkg/reconciler/v1alpha1/eventlistener/eventlistener.go#L989)

Used [METRICS_PROMETHEUS_PORT](https://github.com/tektoncd/triggers/blob/main/vendor/knative.dev/pkg/metrics/config.go#L63)  env to set the port as part of controller.yaml

pkg/reconciler/v1alpha1/eventlistener/eventlistener.go
```diff
@@ -681,28 +703,6 @@ func getDeployment(el *v1alpha1.EventListener, c Config) *appsv1.Deployment {
...
-	container.Env = append(container.Env, corev1.EnvVar{
-		Name:  "CONFIG_OBSERVABILITY_NAME",
-		Value: metrics.ConfigMapName(),
-	})
-	container.Env = append(container.Env, corev1.EnvVar{
-		Name:  "METRICS_DOMAIN",
-		Value: triggersMetricsDomain,
-	})
-
-	container = addCertsForSecureConnection(container, c)
```
savitaashture, line 705: This change is not part of issue
But its a duplicate code so just removed

jmcshane:
outside of the gocritic issue, this looks good to me

/lgtm

cmd/eventlistenersink/main.go
```diff
@@ -83,7 +88,29 @@ func main() {
 	profilingServer := profiling.NewServer(profilingHandler)
 	metrics.MemStatsOrDie(ctx)
 
-	sharedmain.WatchObservabilityConfigOrDie(ctx, configMapWatcher, profilingHandler, logger, EventListenerLogKey)
+	metricsPort, err := strconv.ParseInt(os.Getenv("METRICS_PROMETHEUS_PORT"), 10, 64)
+	if err != nil {
```
dibyom, line 92: Do we still need all of this code if we set the METRICS_PROMETHEUS_PORT? My understanding was that the metrics code looks for this env var and if it is set, it will automatically use that port (just like we do with METRICS_DOMAIN). 

See: https://github.com/tektoncd/triggers/blob/4939226120cdf4646e1f5941fb1b10b8f866abac/vendor/knative.dev/pkg/metrics/config.go#L308-L325

savitaashture: Yes you are right 
Forgot to remove the code after setting env

Updated code now

dibyom (approved):

Issue #1070, opened 2021-04-26
Eventlistener expose information about cluster


chmouel:
# Expected Behavior

Not showing the namespace where the eventlistenner is running

# Actual Behavior

```json
{"eventListener":"openshift-pipelines-ascode-interceptor","namespace":"pipelines","eventID":"773414d6-f8e7-4b4e-a3fb-f99d7758cfa6"}
```

# Additional Info

It's usually bad security practice to expose cluster internal information to the public, since the eventlistenner are usually exposed to the internet for GitHUB or other VCS webhooks it may be better to hide this information.

afrittoli:
Nice catch. The `eventListener` part too might be considered sensitive, as it implies the service name. 
We could perhaps make that information available as opt-in via the service configuration?

dibyom:
> It's usually bad security practice to expose cluster internal information to the public, since the eventlistenner are usually exposed to the internet for GitHUB or other VCS webhooks it may be better to hide this information.

Agreed. Is there a particular issue that exposing the namespace/el name exposes us to? 
The trade-off here is around ease of debugging. Having the EL name and the namespace along with the eventID gives one all the required information to look up the logs for a particular event for debugging. Without those two, the user would have to manually figure out which event listener/namespace the webhook is connected.

PR #1087, opened 2021-05-11
Add EventListener UID to sink response, mark name/namespace as deprecated.

wlynch:
# Changes

Add EventListener UID to sink response, mark name/namespace as deprecated.

The goal here is to provide a more opaque identifier to name/namespace,
which might contain PII in response. We will continue to support
name/namespace for the time being, but mark the fields as deprecated to
be removed in a later release.

This change also includes minor clean up to log annotations for common identifiers.

Part of #1070 

# Submitter Checklist

review them:

# Release Notes

```release-note
- `eventListener` and `namespace` fields in EventListener response are now deprecated. Use `eventListenerUID` instead.
```

### DIFF
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -375,12 +375,24 @@ Otherwise, it responds with a `202 ACCEPTED` HTTP response.
 
 After detecting an event, the `EventListener` responds with the following message:
 
-```JSON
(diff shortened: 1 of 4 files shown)


wlynch:
> We need to fix the e2e tests with the updated sink response. Otherwise, LGTM!

Done.

dibyom (approved):
/lgtm

PR #1088, opened 2021-05-12
Bump CEL version to latest version.

bigkevmcd:
# Changes

This bumps CEL to v0.7.3, fixes the code to work with the new CEL interfaces, and adds in the CEL Base64 encoding functionality.

# Submitter Checklist

review them:

- [X] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commits)

# Release Notes

```release-note
The CEL interceptor is updated to the latest version of CEL, this brings in the CEL base64 decoding functions.

These are documented here https://github.com/google/cel-go/tree/master/ext#base64decode

Note: the CEL base64 decoding function decodes to a CEL byte array, which is not a string, it uses the b'hello' syntax to represent a string of bytes "hello".

The triggers `decodeb64` CEL function is now deprecated in favour of the upstream encoders package, you can replace `body.value.decodeb64` with `string(base64.decode(body.value))`
```


dibyom:
> Note: the CEL base64 decoding function decodes to a CEL byte array, which is not a string, it uses the b'hello' syntax to represent a string of bytes "hello".

Interesting, is there an easy way to transform that into a string?

bigkevmcd:
> > Note: the CEL base64 decoding function decodes to a CEL byte array, which is not a string, it uses the b'hello' syntax to represent a string of bytes "hello".
> 
> Interesting, is there an easy way to transform that into a string?

Good question, so I added a bit of documentation, and some additional tests to show how to handle this :-)

bigkevmcd:
@dibyom Ideally we'd deprecate the one in the triggers extension.

dibyom:
> @dibyom Ideally we'd deprecate the one in the triggers extension.

That makes sense. We haven't deprecated a CEL function before. What I'm thinking is we add a DEPRECATION NOTICE in the release notes here and update the docs for the triggers extension doc to mention that this function is now deprecated and users should switch to the new CEL function?

Issue #1067, opened 2021-04-23
V1Beta1 types for Triggers


dibyom:
Tracking issue for adding a v1beta1 apiVersion for Trigger APIs.

- [x] Get a POC for v1beta1 types: https://github.com/tektoncd/triggers/pull/1079
- [x] Refactor: Remove Builders: #1080 
- [x] Refactor: Move functions and types out of versioned v1alpha1 packages to reduce copy pasting
- [x] Merge v1beta1 types in
- [x] Switch Reconciler to use v1beta1 types
- [x] Switch EventListener sink to use v1beta1 types
- [ ] Add v1beta1 examples --> use selectors + trigger CRDs where possible
- [x] Validate and update docs

dibyom:
WIP v1beta1 branch: https://github.com/dibyom/triggers/tree/v1beta1

PR #1103, opened 2021-05-20
Add v1beta1 Go API types

dibyom:
# Changes

This commit adds v1beta1 go types and the corresponding generated code. These types cannot be used yet until we enable them in the CRDs.

Part of #1067

# Submitter Checklist

review them:

# Release Notes


pkg/apis/triggers/v1beta1/event_listener_defaults.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+		}
+		el.Spec.updatePodTemplate()
+		// To be removed in a later release #1020
+		el.Spec.updateReplicas()
+	}
+}
+
+func (spec *EventListenerSpec) updatePodTemplate() {
+	if spec.DeprecatedPodTemplate != nil {
```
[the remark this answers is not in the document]
dibyom: That particular field is going away with https://github.com/tektoncd/triggers/issues/1105 though generally it depends - we'd have to keep supporting fields in v1beta1 that are a part of v1alpha1 due to how kube api machinery works (v1beta1 is the stored version so all fields must be supported). We'll use the validation webhook to return errors for fields that should not be supported in v1beta1 version.

dibyom:
Integration test failed due to an install error this time.


/test pull-tekton-triggers-integration-tests

pkg/apis/triggers/v1beta1/event_listener_defaults.go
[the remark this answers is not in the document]
wlynch: There's also other deprecated fields for interceptors, etc. My worry is if we introduce these fields, if we're strictly following the deprecation policy this effectively resets the clock for removal since we're "introducing a new field" even though it's deprecated from the start.

If I'm understanding https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility correctly, we don't need to have 1:1 parity with every field in the stored version - we just need to make sure users can go from v1alpha1 to v1beta1 and back without data loss.
The same way [we already mutate `DeprecatedPodTemplate` -> `spec.Resources.KubernetesResource`](https://github.com/tektoncd/triggers/blob/7c72387dd1144c648fecae0746174fab970c6f4a/pkg/apis/triggers/v1alpha1/event_listener_defaults.go#L55-L71), I don't think we need to support the deprecated fields in v1beta1, since we can safely convert back to v1alpha1 by using `KubernetesResource`.  This is analogous to the k8s example where v6 holds (Param, Params), where as the stored v7 version only stores Params and does not preserve the param field.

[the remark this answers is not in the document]
dibyom: I think that's a valid concern. I think there were some issues with JSON marshalling without having the fields set but let me verify. I think given the conversions we are doing in the defaulting webhook, we should not need an extra conversion webhook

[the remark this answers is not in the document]
dibyom: Ok, I think this will actually work!

dibyom:
Here is the path I verified with the storage type set to `v1beta1`:
1. Create a v1alpha1 EL with old deprecated Interceptor fields - the defaulting webhook fixes the old fields

2. Create a v1beta1 EL with no deprecated fields - this works fine.

3. Create a v1beta1 EL with deprecated Interceptor fields - this fails but that is expected since the old fields are not supported in beta.

One question is around what we do with the old style `webhook` interceptors since we can't auto-upgrade them to new fields. Maybe just a validation error is sufficient?

dibyom:
Logs say:

==== UNIT TESTS PASSED ==== 

🙃

pkg/apis/triggers/v1beta1/event_listener_defaults.go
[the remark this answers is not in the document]
savitaashture: I assume it works because when we upgrade from old version to new version there won't be API update so if user want they will create new resources with `v1beta1` am i right ?

savitaashture:
> Here is the path I verified with the storage type set to `v1beta1`:
> 
> 1. Create a v1alpha1 EL with old deprecated Interceptor fields - the defaulting webhook fixes the old fields
> 2. Create a v1beta1 EL with no deprecated fields - this works fine.
> 3. Create a v1beta1 EL with deprecated Interceptor fields - this fails but that is expected since the old fields are not supported in beta.
> 
> Also, checked the coverage reports to verify we did not lose coverage when removing old fields.
> 
> One question is around what we do with the old style `webhook` interceptors since we can't auto-upgrade them to new fields. Maybe just a validation error is sufficient?

I think validation error would be great to give clear message

pkg/apis/triggers/v1beta1/event_listener_defaults.go
```diff
@@ -0,0 +1,42 @@
+/*
+Copyright 2019 The Tekton Authors
```
savitaashture, line 2:
```suggestion
Copyright 2021 The Tekton Authors
```

same applies to all other place

dibyom: Ahh yes, will make that change.

pkg/apis/triggers/v1beta1/trigger_types.go
```diff
@@ -0,0 +1,193 @@
...
+	EventTypes []string   `json:"eventTypes,omitempty"`
+}
+
+// CELInterceptor provides a webhook to intercept and pre-process events
+type CELInterceptor struct {
+	Filter   string       `json:"filter,omitempty"`
+	Overlays []CELOverlay `json:"overlays,omitempty"`
+}
+
+// CELOverlay provides a way to modify the request body using DeprecatedCEL expressions
```
wlynch, line 178: nit: probably needs to be reworded to not include "DeprecatedCEL"

wlynch (commented):
/lgtm

cmd/webhook/main.go
```diff
@@ -42,6 +44,12 @@ var types = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
 	v1alpha1.SchemeGroupVersion.WithKind("TriggerBinding"):        &v1alpha1.TriggerBinding{},
 	v1alpha1.SchemeGroupVersion.WithKind("TriggerTemplate"):       &v1alpha1.TriggerTemplate{},
 	v1alpha1.SchemeGroupVersion.WithKind("Trigger"):               &v1alpha1.Trigger{},
+
+	v1beta1.SchemeGroupVersion.WithKind("ClusterTriggerBinding"): &v1beta1.ClusterTriggerBinding{},
+	v1beta1.SchemeGroupVersion.WithKind("EventListener"):         &v1beta1.EventListener{},
+	v1beta1.SchemeGroupVersion.WithKind("TriggerBinding"):        &v1beta1.TriggerBinding{},
+	v1beta1.SchemeGroupVersion.WithKind("TriggerTemplate"):       &v1beta1.TriggerTemplate{},
+	v1beta1.SchemeGroupVersion.WithKind("Trigger"):               &v1beta1.Trigger{},
```
savitaashture, line 52: Based on the above changes i assume we are not adding `ClusterInterceptor` to beta right ?

pkg/apis/triggers/v1beta1/trigger_types.go
```diff
@@ -0,0 +1,193 @@
+/*
+Copyright 2020 The Tekton Authors
```
savitaashture, line 2: I still see few file have old copyrights

cmd/webhook/main.go
dibyom: Yes, they will be in the v1alpha1 type but can be used from v1beta1 triggers

pkg/apis/triggers/v1beta1/trigger_types.go
dibyom: ahh will fix.

savitaashture (approved):
/lgtm

PR #1125, opened 2021-06-11
Switch reconciler & sink to using v1beta1 types

dibyom:
The reconciler and sink code now use v1beta1 Triggers types instead of v1alpha1
ones. The v1beta1 type has also been enabled as the storage type. This should
be a backwards compatible change. Users should still be able to use the
v1alpha1 types as they have before.

Part of #1067

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/interceptors/interceptors.go
+++ b/pkg/interceptors/interceptors.go
@@ -28,7 +28,8 @@ import (
 	"google.golang.org/grpc/codes"
 	"knative.dev/pkg/apis"
 
-	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
+	triggersv1alpha1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1alpha1"
+	triggersv1beta1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1beta1"
 	corev1lister "k8s.io/client-go/listers/core/v1"
 )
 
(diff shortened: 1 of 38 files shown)


pkg/interceptors/interceptors.go
```diff
@@ -94,35 +95,6 @@ func GetInterceptorParams(i *triggersv1.EventInterceptor) map[string]interface{}
...
-		if i.DeprecatedCEL.Overlays != nil {
-			ip["overlays"] = i.DeprecatedCEL.Overlays
-		}
-	case i.DeprecatedBitbucket != nil:
-		if i.DeprecatedBitbucket.EventTypes != nil {
-			ip["eventTypes"] = i.DeprecatedBitbucket.EventTypes
-		}
-		if i.DeprecatedBitbucket.SecretRef != nil {
-			ip["secretRef"] = i.DeprecatedBitbucket.SecretRef
-		}
```
wlynch, line 125: Is this a breaking change? is it possible to break this out into its own PR for visibility, or does it need to be coupled with the alpha->beta change?

wlynch (commented):
LGTM! One small question.

pkg/interceptors/interceptors.go
dibyom: It should backwards compatible. Users can still create ELs with the old Deprecated interceptor syntax in the v1alpha1 API group. The defaulting webhook should convert it to the new syntax and the sink code can still work -- this defaulting logic has been around for a couple of releases so as long as the EL was created after v0.13 or after this change should be backwards compatible.

PR #1132, opened 2021-06-17
Send EL response before Triggers finish processing 

dibyom:
# Changes

**Note**: This is @jmcshane 's PR #1077 that I rebased.

In the current version of the product, the HTTP request to the EventListener
endpoint waits for all triggers to process to determine if a resource was
created. This is not in line with the suggestion by the source control
systems for webhook endpoints. This change updates the endpoint to respond
as soon as the triggers have been selected for processing.

# Submitter Checklist

review them:

# Release Notes

```release-note
BREAKING CHANGE: With this change the EventListener will stop responding with `201 Created` status code when it creates resources. Instead it will always respond with a `202 Accepted` response code.
```

### DIFF
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sync"
 
 	"github.com/tektoncd/triggers/pkg/apis/triggers"
 	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1beta1"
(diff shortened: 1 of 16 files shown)


savitaashture (approved):

pkg/sink/sink.go
```diff
@@ -57,6 +57,7 @@ type Sink struct {
 	Logger                 *zap.SugaredLogger
 	Recorder               *Recorder
 	Auth                   AuthOverride
+	WaitGroup              *sync.WaitGroup
```
wlynch, line 60: We should probably add a comment here to describe what this wait group is used for.

wlynch (commented):
We should update the commit + PR description to be more git blame friendly.

Otherwise LGTM!.

pkg/sink/sink.go
dibyom: done!

wlynch (approved):
/lgtm

PR #1140, opened 2021-06-28
Update examples for consistency

dibyom:
# Changes

There are a few different updates to the examples folder in this PR.

1. `examples/README.md` described a single example while there were also sub-folders containing other Triggers examples. I moved the example described in `examples/README.md` to its own folder to be consistent with the other examples (embedded-triggers)

2.  I modified `examples/README.md` with a more generic Readme that now contains instructions on invoking triggers locally as well as how to look for the resources that Triggers creates (and logs in the EL when things go wrong).

3. I also fixed the ingress e2e test that was being flaky.

Part of #1067

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#tests) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#docs) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commits)
- [ ] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/triggers/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

PR #1141, opened 2021-06-29
Switch branch name from master to main

jferna57:
# Changes

master branch doesn't exist in https://github.com/tektoncd/triggers repo. Main is the new name.

### DIFF
--- a/examples/cron/triggerbinding.yaml
+++ b/examples/cron/triggerbinding.yaml
@@ -5,6 +5,6 @@ metadata:
 spec:
   params:
   - name: gitrevision
-    value: master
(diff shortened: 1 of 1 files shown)


dibyom:
Thanks @jferna57 
Just two things:
1. Could you sign the CLA? 
2. Could you update the commit message a bit - something like - `switch branch name from master to main` or something like that

dibyom:
(The test failure is unrelated and should be fixed by https://github.com/tektoncd/triggers/pull/1140/commits/f02d92ca29195b544a3235a6b15dfa7d0a481451)

Issue #192, opened 2019-10-28
Meaningful (e2e) tests for examples


ncskier:
# Expected Behavior
We should know when code changes break the examples we are hosting in the Triggers repository.

# Actual Behavior
We do not currently run automated tests on all of the examples included in the Triggers repository. So, a code change can break the examples without us knowing about it. This was made apparent recently when a change broke the cron example (https://github.com/tektoncd/triggers/pull/190).

I think that we should try to have meaningful automated tests for all of the examples, and we should have examples for the most common use-cases of Triggers. This will help stop us from making changes that have unintended consequences, and will make it easy for people to start using Triggers.

I also hope that having meaningful tests for each example will force us to keep the number of examples to a small and manageable size. If we don't want to create tests for an example, then we probably shouldn't have the example in the Triggers repository.

# Additional Info
I think that at the moment the examples that lack meaningful tests are:
- [cron](https://github.com/tektoncd/triggers/tree/master/examples/cron)
- [getting started guide](https://github.com/tektoncd/triggers/tree/master/docs/getting-started)
- documentation [examples](https://github.com/tektoncd/triggers/tree/master/examples)

The documentation examples seem to overlap with the cron example and getting started guide, so it might be beneficial to merge the documentation examples into one of the other examples?

dibyom:
priority/important-soon
/maybe-next-milestone

dibyom:
Have a branch with a simple bash test runner: https://github.com/tektoncd/triggers/compare/master...dibyom:examples_test?expand=1

dibyom:
There are 3 parts to this:
- [x] Validate the Trigger YAMLs - This is simple. We can apply the YAMLs from each example and ensure they are created.
- [ ] Invoke the Trigger - This would involve connecting to the EventListener, and then sending a HTTP request with some predefined Headers and event body that is different for each example. For my PoC, I created a convention that each example folder would have a separate file for headers and body. The test script will look for these files and use them for crafting the curl request to send to the EventListener. (there are other ways to do this e.g. we could have an convention where each example folder contains a test.sh file with the curl command).
[...]

PR #1142, opened 2021-07-01
Adds E2E tests for Examples

sm43:
The e2e tests will apply the yaml files from directory and and validate
eventlistener to be in running state. later, we can improve by triggering the el by
a curl command and check if works fine.

Implements part of #192 

# Submitter Checklist

review them:

# Release Notes


test/e2e-tests-examples.sh
```diff
@@ -0,0 +1,90 @@
...
+      exit 1
+    }
+  done
+}
+
+check_eventlistener() {
+  info "Waiting for EventListener to be available"
+
+  elName=${current_example}-listener
+  kubectl wait --for=condition=Available --timeout=20s eventlisteners/${elName} || {
```
dibyom, line 53: I think we should use condition=Ready since that is available for both k8s Els as well as Knative based ELs

test/e2e-tests-examples.sh
```diff
@@ -0,0 +1,90 @@
...
+  done
+}
+
+# Assumptions:
+# Name of example would be name of directory
+# Name of eventlistener must be (exampleName)-listener
+
+main() {
+  header "Starting Examples E2E Test"
+  folders="bitbucket github gitlab cron v1alpha1-task"
```
dibyom, line 77: I think we should be able to use most examples here. Two which probably won't work are the tls-connection one and the custom resource (knative) one.

dibyom (approved):
/lgtm

thanks @sm43

PR #1155, opened 2021-07-16
Adds curl test in e2e tests for examples

sm43:
In the present e2e test for examples this adds port forwading the
event listener and then doing a curl to test if a taskrun/pipelinerun
is created.

Ignored examples: `( cron trigger-ref v1alpha1-task )`

Part of #192 

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/test/e2e-tests-examples.sh
+++ b/test/e2e-tests-examples.sh
@@ -23,20 +23,21 @@ source $(dirname $0)/e2e-common.sh
 set -u -e -o pipefail -x
 
 current_example=""
+current_example_version=""
 yaml_files=""
+port_forward_pid=""
 
 info() {
-  echo; echo "*** Example ${current_example}: $@ ***"; echo;
+  echo; echo "*** Example ${current_example_version}/${current_example}: $@ ***"; echo;
 }
 
 err() {
-  echo; echo "ERROR: Example ${current_example}: $@"; echo;
+  echo; echo "ERROR: Example ${current_example_version}/${current_example}: $@"; echo;
 }
 
 apply_files() {
   info "Applying Resources"
-  local version=$1
-  folder=${REPO_ROOT_DIR}/examples/${version}/${current_example}
+  folder=${REPO_ROOT_DIR}/examples/${current_example_version}/${current_example}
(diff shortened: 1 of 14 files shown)


test/e2e-tests-examples.sh
```diff
@@ -60,10 +61,65 @@ check_eventlistener() {
...
+    info 'ignoring for port forwarding'
+    return
+  fi
+
+  info "Port forwarding to execute curl command"
+
+  kubectl port-forward service/el-${current_example}-listener 8080:8080 &
+  port_forward_pid=$!
+
+  # Wait a few seconds for forwarding to happen
```
dibyom, line 78: hmm I wonder if there is a way we can wait for port-forward to succeed instead of a sleep...but if there is nothing obvious, this is ok for now

dibyom (commented):
Generally looks good...We just moved the examples between v1alpha1 and v1beta1, so you'd have to copy the curl.sh files to the specific examples in v1alpha1 and v1beta1.

sm43:
@dibyom it seems `kubectl wait` cmd waits for a field to get assigned a value and not till it becomes true
so sometimes `el` ready is `false` kubectl cmd passes and executes port forward and fails.
later the ready status changes to true.
so it seems we need something which will wait till ready is true.
I was thinking to write a for loop which would wait for a particular time and check status ready is true or not.
Do you have any other alternative in mind for waiting for status?

dibyom:
Hmmm...that's weird. Maybe we can try explicitly setting wait condition to be ready=true i.e. `kubectl wait --for=condition=Ready=true`?

sm43:
> Hmmm...that's weird. Maybe we can try explicitly setting wait condition to be ready=true i.e. `kubectl wait --for=condition=Ready=true`?
> 
> If that doesn't work, then yeah we could retry the port-forward a few times before giving up.

I looked into it, well the script was exiting due to some error 😅 due to wait command was failing and 
also by default in kubectl wait command `Ready` is `true` 
`--for=condition=Ready=true`

updated the script for the examples. 
I am not sure why `v1alpha1-task` is failing in script, will figure out separately.

dibyom (approved):

PR #1158, opened 2021-07-21
Packages ClusterRoles with Triggers deployment for eventlistener 

sm43:
This packages 2 clusterroles with triggers deployment for el
that can be used by users for their deployments.
User will have to creates following resources:
- a serviceaccount which would be used with eventlistener
- a rolebinding with above sa and `tekton-triggers-eventlistener-roles` clusterrole
- a clusterrolebinding with above sa and `tekton-triggers-eventlistener-clusterroles`
  clusterrole

Signed-off-by: Shivam Mukhade smukhade@redhat.com

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/config/200-clusterrole.yaml
+++ b/config/200-clusterrole.yaml
@@ -54,3 +54,40 @@ rules:
   - apiGroups: [""]
     resources: ["secrets"]
     verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: tekton-triggers-eventlistener-roles
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-triggers
+rules:
+  - apiGroups: ["triggers.tekton.dev"]
+    resources: ["eventlisteners", "triggerbindings", "triggertemplates", "triggers"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["tekton.dev"]
+    resources: ["pipelineruns", "pipelineresources", "taskruns"]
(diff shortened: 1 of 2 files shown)


config/200-clusterrole.yaml
```diff
@@ -53,4 +53,41 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["secrets"]
+    verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: tekton-triggers-eventlistener
```
dibyom, line 61: what do you think of calling this simply `eventlistener-roles` and the other one `eventlistener-clusterroles`

dibyom (commented):
Looks good...my only comment is to see if we can make the names a bit smaller but that's minor

config/200-clusterrole.yaml
sm43: end users might have many things installed on the cluster so if we keep `tekton-triggers-` as prefix so wouldn't it be easier to understand where the clusterrole came from.

sm43: and also it would be uniform with other resources we package

dibyom: fair enough!

sm43:
let me rename it to `-roles` and `-clusterroles` sounds better :)

dibyom:
@sm43 One thing I forgot -- could you add a release notes section to this PR?

/kind feature

sm43:
> @sm43 One thing I forgot -- could you add a release notes section to this PR?
> 
> /kind feature

@dibyom done. would it be better to add this in docs somewhere?  could you point where I can add this in docs

dibyom:
Yup!
https://github.com/tektoncd/triggers/blob/main/docs/eventlisteners.md#specifying-the-kubernetes-service-account seems like the right place

PR #1162, opened 2021-07-25
Declares named metrics ports

solsson:
# Changes

Adds `metrics` port to controller deployment and fixes the port number for the same name for webhook.

Named ports are useful for prometheus-operator [podMetricsEndpoints](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#podmetricsendpoint).

The port number for webhook wasn't the actual port number for the released image.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -77,6 +77,9 @@ spec:
           value: tekton.dev/triggers
         - name: METRICS_PROMETHEUS_PORT
           value: "9000"
+        ports:
+        - name: metrics
+          containerPort: 9000
         securityContext:
           allowPrivilegeEscalation: false
           # User 65532 is the distroless nonroot user ID
--- a/config/webhook.yaml
+++ b/config/webhook.yaml
@@ -69,7 +69,7 @@ spec:
           value: tekton.dev/triggers
         ports:
         - name: metrics
-          containerPort: 9000
+          containerPort: 9090
         - name: profiling
           containerPort: 8008
         - name: https-webhook


dibyom:
> The port number for webhook wasn't the actual port number for the released image.

Yikes good catch! @savitaashture do we have different metrics port numbers for webhook and controller metrics?

khrm:
@dibyom By default, metrics port is 9090 but in case of controller, we are defining [port] as 9000 in environment. (https://github.com/tektoncd/triggers/blob/f7b6ed0f33045f9ebd19aeb83f0545e37bfcca80/config/controller.yaml#L79)

config/controller.yaml
```diff
@@ -77,6 +77,9 @@ spec:
           value: tekton.dev/triggers
         - name: METRICS_PROMETHEUS_PORT
           value: "9000"
```
savitaashture, line 79: @dibyom we use this env while creating Knative service and Knative itself have metrics exposed on `9090` in order to avoid conflict i have changed to `9000` as part of this PR https://github.com/tektoncd/triggers/pull/1086 (given details over the PR description)

But i think forgot to change for  webhook

config/webhook.yaml
```diff
@@ -69,7 +69,7 @@ spec:
           value: tekton.dev/triggers
         ports:
         - name: metrics
-          containerPort: 9000
+          containerPort: 9090
```
savitaashture, line 72: as controller is exposed on `9000` lets change this also to `9000` itself
/cc @dibyom @khrm

dibyom:
ok that sounds good. Looks like we do not expose the metrics ports via the respective services. Should we do that too?

solsson:
I'm quite confused now what changed in v0.15, if anything. Controller now has metrics on :9000 while webhook has metrics on :9090. Is that what I should expect?

PR #1166, opened 2021-07-28
Remove Redundant Status Tag from EL Metrics

khrm:
Removing the status tag from event_count and http_duration_seconds because it is always 200.

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
Remove Redundant Status Tag from EventListener Metrics. 
Breaking Change:
Status label has been removed from metric eventlistener_http_duration_seconds_*. Please remove querying base on status label.
```

### DIFF
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -441,10 +441,10 @@ kubectl get pods --selector eventlistener=my-eventlistener
 The following pipeline metrics are available on the `eventlistener` Service on port `9000`.
 
 |  Name | Type | Labels/Tags | Status |
-| ---------- | ----------- | ----------- | ----------- |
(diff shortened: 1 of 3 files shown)


dibyom:
@khrm This won't be a breaking change for any current users right?

khrm:
@dibyom Good catch. This might be a breaking change if a user is using status while querying in metering system. I updated this in release section. Status field is needlessly increasing the cardinality without providing any benefit.

dibyom:
@khrm Yeah, so I do not think the metrics is part of the beta API guarantees so we should be ok making this change even though it might be backwards incompatible? What do you think

dibyom (approved):

khrm:
Yes, it is written in docs that it is experimental. So this is fine.

Issue #1089, opened 2021-05-12
Create integration test run for knative eventlistener service


jmcshane:
### Feature request

A regression was introduced for the knative service type. We can avoid this in the future by creating a good integration test that deploys an eventlistener in this fashion.

### Use case

Avoid future regressions against the knative Eventlistener deployment

PR #1170, opened 2021-08-02
Fix status reconcilation for Knative service

savitaashture:
### Issue:
When we create custom object the `READY` status of EL always shows `false`
### RootCause:
Polling of Knative service are not done properly and because of that only initial status are set to EL status [here](https://github.com/tektoncd/triggers/blob/main/pkg/reconciler/eventlistener/eventlistener.go#L669)

### Changes
Added a condition check to set the status to EL when Knative service `READY` state is `True`

/cc @dibyom
 
### Effected versions:
Triggers `v0.15.0` and `v0.15.0-rc1`

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/eventlistener/eventlistener.go
+++ b/pkg/reconciler/eventlistener/eventlistener.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	stdError "errors"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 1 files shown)


dibyom:
Oh interesting.
Code looks good. Two questions:
1. Do we need a patch release for this?
2. Could we add a test for this behavior?

savitaashture:
> Oh interesting.
> Code looks good. Two questions:
> 
> 1. Do we need a patch release for this?

Yes

> 2. Could we add a test for this behavior?

For Knative service we are using [withPod](https://github.com/tektoncd/triggers/blob/main/pkg/reconciler/eventlistener/eventlistener_test.go#L417) ducktype and that doesn't have status field to set some dummy value

Lets cover e2e test as part of this issue https://github.com/tektoncd/triggers/issues/1089 so that in future these kind of issues won't miss

Issue #1172, opened 2021-08-03
Installed Tekton pipelines, triggers etc. into a new OKD 4.7 cluster.


johnlongo:
Installed the latest Tekton pipelines, triggers etc. into a new OKD 4.7 (4.7.0-0.okd-2021-07-03-190901) cluster using the following steps:
oc adm policy add-scc-to-user anyuid -z tekton-pipelines-controller
oc adm policy add-scc-to-user anyuid -z tekton-pipelines-webhook
oc apply --filename https://storage.googleapis.com/tekton-releases/pipeline/latest/release.notags.yaml
oc apply --filename https://storage.googleapis.com/tekton-releases/triggers/latest/release.notags.yaml
oc apply --filename https://storage.googleapis.com/tekton-releases/triggers/latest/interceptors.notags.yaml
oc apply --filename https://storage.googleapis.com/tekton-releases/dashboard/latest/tekton-dashboard-release.yaml

![image](https://user-images.githubusercontent.com/23619505/127953603-d650499d-1e35-4707-bc09-5ed6646f3ba7.png)

After that I defined a pipeline for one of my deployments and an event listener (see attached).
[trigger-openshift-alertmanager-listener.txt](https://github.com/tektoncd/triggers/files/6920838/trigger-openshift-alertmanager-listener.txt)

When the listener starts up if goes into a crash loop (see attached)
![image](https://user-images.githubusercontent.com/23619505/127953819-915c7ec5-42da-41ff-9864-3e7fda0b8992.png)

Please let me know how to fix this issue because I really would like to use Tekton but if I cant use a listener it's going to be a manual process to start the pipeline and I would like to avoid that if possible.

Thank you in advance for any help.

khrm:
I think this is happening because we are setting security context for EventListener  which isn't allowed in OpenShift (and probably in OKD also). We apply a patch in downstream to fix this. https://github.com/openshift/tektoncd-triggers/blob/master/openshift/patches/0001-Change-eventlistener-flag-default-value-to-false.patch

You can use this release.yaml to fix this issue: https://github.com/openshift/tektoncd-triggers/tree/release-v0.14.2/openshift/release

Or change value of `el-default-security-context` in controller spec. This value was missing and will be added in next release.

johnlongo:
Installed tekton again in a freshly restored cluster with the following, but still seeing the same issue:

oc new-project tekton-pipelines

oc adm policy add-scc-to-user anyuid -z tekton-pipelines-controller
[...]

PR #1175, opened 2021-08-04
Add SetSecuritycontext for EventListener in Controller.yaml

khrm:
Put it in the controller.yaml so that a user can refer release.yaml to configure this value instead of digging in the code.
Potentially enable us to run EL in OKD: #1172
# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
SetSecuritycontext for EventListener in controller deployment can be configured now.
```

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -57,6 +57,7 @@ spec:
           "-stderrthreshold", "INFO",
           "-el-image", "ko://github.com/tektoncd/triggers/cmd/eventlistenersink",
           "-el-port", "8080",
+          "-el-security-context=true",
           "-el-readtimeout", "5",
           "-el-writetimeout", "40",
           "-el-idletimeout", "120",


config/controller.yaml
```diff
@@ -57,6 +57,7 @@ spec:
           "-stderrthreshold", "INFO",
           "-el-image", "ko://github.com/tektoncd/triggers/cmd/eventlistenersink",
           "-el-port", "8080",
+          "-el-security-context", "true",
```
savitaashture, line 60: @khrm i see default value for `el-security-context` is `true` only 
https://github.com/tektoncd/triggers/blob/main/pkg/reconciler/eventlistener/config.go#L25

khrm: Yes, that's why I put it as true here. I added it here so that if a user want to configure, he can refer here release.yaml instead of code.

savitaashture:
@khrm even if i set `-el-security-context` to `false` in controller.yaml file its alwasy taking value as `true` only

khrm:
@savitaashture I updated controller.yaml file. The way to pass bool is either using =false/true or not passing in case of false.

savitaashture:
> @savitaashture I updated controller.yaml file. The way to pass bool is either using =false/true or not passing in case of false.

Yup it works now as expected :+1:

savitaashture (approved):
/lgtm

Issue #1178, opened 2021-08-06
HA support for Tekton Triggers controllers


dibyom:
### Feature request

We should investigate what it would take to support HA for the Tekton Triggers controller + webhook.
Docs for Pipelines: https://tekton.dev/docs/pipelines/enabling-ha/#controller-ha

hokadiri:
/remove-lifecycle stale
Still needed please!

PR #1179, opened 2021-08-06
Mark webhook and controller as safe-to-evict

dibyom:
# Changes

This is a port of https://github.com/tektoncd/pipeline/pull/4124

The safe-to-evict annotation tells the cluster autoscaler whether the
pod can be evicted to allow the node it's on to scale down.

Unfortunately, blocking node eviction means the node that the pod(s) get
scheduled to can't be scaled down. Furthermore, the nodes can't be fully
drained when updating the cluster. This can leave a cluster in a
mid-upgrade state that can make issues difficult to diagnose and reason
about.

With this change, a cluster scale-down event might cause temporary
service unreliability with the default single-replica configuration

One difference from the pipelines PR is that Triggers currently does not document enabling HA.
I opened #1178 for it. Should we implement that before merging this? (@vdemeester @khrm )

# Submitter Checklist

review them:

# Release Notes

```release-note
By default, controller components are now marked as safe-to-evict by the cluster autoscaler. 
```

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -35,8 +35,6 @@ spec:
       app.kubernetes.io/part-of: tekton-triggers
   template:
     metadata:
-      annotations:
-        cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
       labels:
         app.kubernetes.io/name: controller
         app.kubernetes.io/component: controller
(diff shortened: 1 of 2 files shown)


khrm (commented):
/lgtm

savitaashture (approved):

PR #1189, opened 2021-08-17
Adds annotation to disable payload validation in eventlistener

### LINKED ISSUE #1152: 0.12: PR #969 breaks custom interceptors that convert non-json to JSON
# Expected Behavior

in triggers < v0.12, my interceptor that convert non-JSON bodies to JSON, work fine. For example I have Slack POSTing a body to my event listener endpoint where BODY = `payload=[urlencodedvalue]`

My interceptor converts the url encoded payload value into JSON and returns it back to the caller (event listener) then the rest of the interceptor continues and things move forward

https://github.com/bitsofinfo/cicdstatemgr/blob/master/examples/tekton/pipelines/triggers.event-listener.yaml#L404 & https://github.com/bitsofinfo/slack-payload-handler

In > 0.12, the interceptors are no longer called, instead the event listener immediately fails fast on these POSTs with:

```
{"level":"error","ts":"2021-07-14T17:13:04.159Z","logger":"eventlistener","caller":"sink/validate_payload.go:39","msg":"Invalid event body format format: invalid character 'p' looking for beginning of value","knative.dev/controller":"eventlistener","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.IsValidPayload.func1\n\tgithub.com/tektoncd/triggers/pkg/sink/validate_payload.go:39\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2069\ngithub.com/tektoncd/triggers/pkg/sink.(*Sink).NewMetricsRecorderInterceptor.func1\n\tgithub.com/tektoncd/triggers/pkg/sink/metrics.go:81\ngithub.com/tektoncd/triggers/pkg/sink.(*MetricsHandler).Intercept.func1\n\tgithub.com/tektoncd/triggers/pkg/sink/metrics.go:153\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2069\nnet/http.(*ServeMux).ServeHTTP\n\tnet/http/server.go:2448\nnet/http.(*timeoutHandler).ServeHTTP.func1\n\tnet/http/server.go:3316"}
```

# Actual Behavior

See above:

https://github.com/bitsofinfo/cicdstatemgr/tree/master/examples/tekton

Change: https://github.com/bitsofinfo/cicdstatemgr/blob/master/examples/tekton/core/install.sh#L36 to v12+ prior to going through the example.

# Additional Info

- Tekton Pipeline version: v0.20.1

see #969 

@boarder981

sm43:
This adds an annotation support `tekton.dev/payload-validation`
for eventlistener. If it is added to el with value as `false` then
the payload from events will not be validated and will be directly
passed to interceptors.
By default the payload validation is enabled. Only if annotation
is defined and its value is `false` then it will be disabled
for that particular el.

Closes #1152 

# Submitter Checklist

review them:

# Release Notes

User can define an annotation on eventlistener `tekton.dev/payload-validation: "false"` to disable the payload validation in that event listener. By default payload validation is enabled.

### DIFF
--- a/pkg/reconciler/eventlistener/eventlistener.go
+++ b/pkg/reconciler/eventlistener/eventlistener.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/tektoncd/triggers/pkg/apis/triggers"
 	"github.com/tektoncd/triggers/pkg/apis/triggers/contexts"
 	"github.com/tektoncd/triggers/pkg/apis/triggers/v1beta1"
 	triggersclientset "github.com/tektoncd/triggers/pkg/client/clientset/versioned"
@@ -883,6 +884,13 @@ func getContainer(el *v1beta1.EventListener, c Config, pod *duckv1.WithPod) core
 		isMultiNS = true
 	}
 
+	payloadValidation := true
+	if value, ok := el.GetAnnotations()[triggers.PayloadValidationAnnotation]; ok {
+		if value == "false" {
+			payloadValidation = false
+		}
+	}
+
 	return corev1.Container{
 		Name:  "event-listener",
 		Image: *c.Image,
@@ -900,6 +908,7 @@ func getContainer(el *v1beta1.EventListener, c Config, pod *duckv1.WithPod) core
 			"--idletimeout=" + strconv.FormatInt(*c.IdleTimeOut, 10),
 			"--timeouthandler=" + strconv.FormatInt(*c.TimeOutHandler, 10),
 			"--is-multi-ns=" + strconv.FormatBool(isMultiNS),
+			"--payload-validation=" + strconv.FormatBool(payloadValidation),
 		},
 		Env: env,
 	}
--- a/cmd/eventlistenersink/main.go
+++ b/cmd/eventlistenersink/main.go
@@ -149,6 +149,7 @@ func main() {
 		HTTPClient:             http.DefaultClient,
 		EventListenerName:      sinkArgs.ElName,
 		EventListenerNamespace: sinkArgs.ElNamespace,
+		PayloadValidation:      sinkArgs.PayloadValidation,
 		Logger:                 logger,
 		Recorder:               recorder,
 		Auth:                   sink.DefaultAuthOverride{},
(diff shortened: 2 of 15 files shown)


dibyom (commented):
code looks fine. 
quick bike shed on the actual flag name -- `disable-payload-validation` vs `enable-payload-validation` (true by default). Or even `validate-payload`

I think enable-payload-validation: true sounds clearer than disable-payload-validation: false

pkg/reconciler/eventlistener/eventlistener.go
```diff
@@ -76,6 +76,8 @@ const (
 	defaultConfig = `{"level": "info","development": false,"sampling": {"initial": 100,"thereafter": 100},"outputPaths": ["stdout"],"errorOutputPaths": ["stderr"],"encoding": "json","encoderConfig": {"timeKey": "ts","levelKey": "level","nameKey": "logger","callerKey": "caller","messageKey": "msg","stacktraceKey": "stacktrace","lineEnding": "","levelEncoder": "","timeEncoder": "iso8601","durationEncoder": "","callerEncoder": ""}}`
 
 	triggersMetricsDomain = "tekton.dev/triggers"
+
+	disablePayloadValidationAnnotation = "tekton.dev/disable-payload-validation"
```
savitaashture, line 80: instead of hardcoding here again can we make use from validation.go file ?
because i see `disablePayloadValidationAnnotation` added in multiple place

savitaashture:
> code looks fine.
> quick bike shed on the actual flag name -- `disable-payload-validation` vs `enable-payload-validation` (true by default). Or even `validate-payload`
> 
> I think enable-payload-validation: true sounds clearer than disable-payload-validation: false

yup keeping default value as true `enable-payload-validation: true` and if user want to disable it they can pass it as false as part of controller.yaml similar to `-el-security-context`

sm43:
> > code looks fine.
> > quick bike shed on the actual flag name -- `disable-payload-validation` vs `enable-payload-validation` (true by default). Or even `validate-payload`
> > I think enable-payload-validation: true sounds clearer than disable-payload-validation: false
> 
> yup keeping default value as true `enable-payload-validation: true` and if user want to disable it they can pass it as false as part of controller.yaml similar to `-el-security-context`

yeah so do you mean to have field in config-default to change the default value at controller level?
and annotation would be at eventlistener level?

pkg/reconciler/eventlistener/eventlistener.go
sm43: addressed.

dibyom (approved):

savitaashture (commented):
/lgtm
Thank you :+1:

PR #1213, opened 2021-08-31
eldeployment: remove securityContext.runAsUser

zhouhaibing089:
The setting on `securityContext.runAsUser` is redundant. The container
runtime determines which user to run based on image config and thus uses
65532 by default.

Furthermore, some kubernetes distributions randomizes runAsUser id
according to their security policies(for e.g, openshift enables per
application uid range), and when that happens, the uid can be set to
something else.

```console
$ crane config gcr.io/distroless/static:nonroot | jq .
{
  "architecture": "amd64",
  "author": "Bazel",
  "created": "1970-01-01T00:00:00Z",
  "history": [
    {
      "author": "Bazel",
      "created": "1970-01-01T00:00:00Z",
      "created_by": "bazel build ..."
    }
  ],
  "os": "linux",
  "rootfs": {
    "type": "layers",
    "diff_ids": [
      "sha256:07363fa8421000ad294c2881d17b0535aabdd17ced2a874eb354a9d8514d3d59"
    ]
  },
  "config": {
    "Env": [
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
      "SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt"
    ],
    "User": "65532",
    "WorkingDir": "/home/nonroot"
  }
}
```

See `config.User` on its default configuration.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/eventlistener/eventlistener_test.go
+++ b/pkg/reconciler/eventlistener/eventlistener_test.go
@@ -260,7 +260,6 @@ func makeDeployment(ops ...func(d *appsv1.Deployment)) *appsv1.Deployment {
 					}},
 					SecurityContext: &corev1.PodSecurityContext{
 						RunAsNonRoot: ptr.Bool(true),
-						RunAsUser:    ptr.Int64(65532),
 					},
 				},
 			},
--- a/pkg/reconciler/eventlistener/resources/deployment.go
+++ b/pkg/reconciler/eventlistener/resources/deployment.go
@@ -50,7 +50,6 @@ var (
 	}
 	strongerSecurityPolicy = corev1.PodSecurityContext{
 		RunAsNonRoot: ptr.Bool(true),
-		RunAsUser:    ptr.Int64(65532),
 	}
 )
 


vdemeester:
> Furthermore, some kubernetes distributions randomizes runAsUser id
> according to their security policies(for e.g, openshift enables per
> application uid range), and when that happens, the uid can be set to
> something else.

Indeed. Because of this, we do patch this on release on OpenShift. Removing this would remove one of our patch 🙃 . I think it predates `distroless` images using that user. 
(Note that we should do the same on `tektoncd/pipeline` and maybe `tektoncd/operator`)

gabemontero:
> > Furthermore, some kubernetes distributions randomizes runAsUser id
> > according to their security policies(for e.g, openshift enables per
> > application uid range), and when that happens, the uid can be set to
> > something else.
> 
> Indeed. Because of this, we do patch this on release on OpenShift. Removing this would remove one of our patch upside_down_face . I think it predates `distroless` images using that user.
> (Note that we should do the same on `tektoncd/pipeline` and maybe `tektoncd/operator`)

Yep I mentioned consistency with the other repos in the slack conversation around this that occurred (last week I believe)

savitaashture (approved):

khrm (commented):
/lgtm

Issue #649, opened 2020-06-30
Emit Events when a Trigger is processed successfully


dibyom:
# Expected Behavior

When a Trigger successfully creates a resources, we should emit a k8s event.

# Actual Behavior

Triggers does not emit any k8s events today.

# Additional Info

- We could also consider emitting events on Trigger failures -- but likely there will be many of them so this should be optional.
- Events in Pipelines: https://github.com/tektoncd/pipeline/blob/c7a60b8a77c05d8ddbc0d5a931e3f66486d512ef/docs/events.md

/kind feature

siamaksade:
Events in case of failure are also very valuable given that the alternative is for the user to dig into the eventlistener pod logs. The use-case is that user commits a change to a GitHub repo and GitHub reports successfully invocation of the webhook but Tekton does not start the pipeline. Events would be a good starting point for finding what has gone wrong.

dibyom:
Agreed! Though for failure events we'd also need to resolve #336 first!

savitaashture:
@dibyom planning to move to next release

Issue #1219, opened 2021-09-03
Eventlistener using Knative Service: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update


Fabian-K:
Hi,

I´m facing an issue when creating a EventListener using Knative Service. When applying the EventListener

```yaml
apiVersion: triggers.tekton.dev/v1beta1
kind: EventListener
metadata:
  name: default-listener
spec:
  resources:
    customResource:
      apiVersion: serving.knative.dev/v1
      kind: Service
      spec:
        template:
          spec:
            serviceAccountName: default
```

I can see the Knative Service being created
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  annotations:
    serving.knative.dev/creator: system:serviceaccount:tekton-pipelines:tekton-triggers-controller
    serving.knative.dev/lastModifier: system:serviceaccount:tekton-pipelines:tekton-triggers-controller
  creationTimestamp: "2021-09-03T16:31:48Z"
  generation: 1
  labels:
    app.kubernetes.io/managed-by: EventListener
    app.kubernetes.io/part-of: Triggers
    eventlistener: default-listener
  name: el-default-listener
  namespace: default
  ownerReferences:
  - apiVersion: triggers.tekton.dev/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: EventListener
    name: default-listener
    uid: 653d854e-e9f4-42da-bfa2-a959b3eca48c
  resourceVersion: "44431"
  uid: 89acc44d-5d96-475a-a6bb-1f49b1a3e0f9
spec:
  template:
    metadata:
      creationTimestamp: null
    spec:
      containerConcurrency: 0
      containers:
      - args:
        - --el-name=default-listener
        - --el-namespace=default
        - --port=8080
        - --readtimeout=5
        - --writetimeout=40
        - --idletimeout=120
        - --timeouthandler=30
        - --is-multi-ns=false
        - --payload-validation=true
        env:
        - name: SYSTEM_NAMESPACE
          value: default
        - name: CONFIG_OBSERVABILITY_NAME
          value: config-observability-triggers
        - name: METRICS_DOMAIN
          value: tekton.dev/triggers
        - name: METRICS_PROMETHEUS_PORT
          value: "9000"
[...]

PR #1222, opened 2021-09-07
Add kubernetes based events for triggers

savitaashture:
# Changes

This PR adds changes to display k8s events for EventListener Pod when http request reaches to EL

```
kubectl get events -w | grep eventlistener

0s          Normal    dev.tekton.event.triggers.started.v1   eventlistener/github-listener              
0s          Normal    dev.tekton.event.triggers.successful.v1   eventlistener/github-listener              

```
/cc @dibyom @khrm

Addresses https://github.com/tektoncd/triggers/issues/649

# Submitter Checklist

review them:

# Release Notes

```release-note
Triggers now support emitting Kubernetes events from EventListener Pod
```

### DIFF
--- a/config/200-clusterrole.yaml
+++ b/config/200-clusterrole.yaml
@@ -85,6 +85,9 @@ rules:
     resources: ["podsecuritypolicies"]
     resourceNames: ["tekton-triggers"]
     verbs: ["use"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["create"]
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
--- a/pkg/apis/triggers/v1alpha1/event_listener_types.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_types.go
@@ -261,6 +261,12 @@ func (els *EventListenerStatus) SetConditionsForDynamicObjects(conditions v1beta
 			Message: cond.Message,
 		})
 	}
+
+	els.SetCondition(&apis.Condition{
+		Type:    apis.ConditionReady,
+		Status:  corev1.ConditionTrue,
+		Message: "EventListener is ready",
+	})
 }
 
 // SetExistsCondition simplifies setting the exists conditions on the
--- a/pkg/reconciler/eventlistener/eventlistener.go
+++ b/pkg/reconciler/eventlistener/eventlistener.go
@@ -19,7 +19,6 @@ package eventlistener
 import (
 	"context"
 	"encoding/json"
-	stdError "errors"
 	"fmt"
 	"strings"
 	"sync"
@@ -342,8 +341,8 @@ func (r *Reconciler) reconcileCustomObject(ctx context.Context, el *v1beta1.Even
 		for _, cond := range customConditions {
 			if cond.Type == apis.ConditionReady {
 				if cond.Status != corev1.ConditionTrue {
-					logging.FromContext(ctx).Warn("custom object is not yet ready")
-					return stdError.New("custom object is not yet ready")
+					logging.FromContext(ctx).Warnf("custom object is not yet ready because %s", cond.Message)
+					return fmt.Errorf("custom object is not yet ready because %s", cond.Message)
 				}
 			}
 		}
--- a/pkg/reconciler/events/event.go
+++ b/pkg/reconciler/events/event.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2021 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package events
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+)
+
+const (
+	// Emit events if user provides enable-eventlistener-events: "true"
+	EnableEventListenerEvents = "enable-eventlistener-events"
+	// TriggerProcessingStartedV1 is sent for Sink Triggers when a trigger is started
+	TriggerProcessingStartedV1 = "dev.tekton.event.triggers.started.v1"
+	// TriggerProcessingSuccessfulV1 is sent for Sink Triggers when a trigger is successful
+	TriggerProcessingSuccessfulV1 = "dev.tekton.event.triggers.successful.v1"
+	// TriggerProcessingFailedEventV1 is sent for Sink Triggers when we fail to process trigger
+	TriggerProcessingFailedV1 = "dev.tekton.event.triggers.failed.v1"
+)
+
+// Emit emits events for object
+// Supported events are k8s events.
+func Emit(recorder record.EventRecorder, eventType string, object runtime.Object, err error) {
+	sendKubernetesEvents(recorder, eventType, object, err)
+}
+
+func sendKubernetesEvents(c record.EventRecorder, eventType string, object runtime.Object, err error) {
+	switch err {
+	case nil:
+		if eventType == TriggerProcessingFailedV1 {
+			c.Event(object, corev1.EventTypeWarning, eventType, "")
+		} else {
+			c.Event(object, corev1.EventTypeNormal, eventType, "")
+		}
+	default:
+		c.Event(object, corev1.EventTypeWarning, eventType, err.Error())
+	}
+}
--- a/pkg/reconciler/events/event_test.go
+++ b/pkg/reconciler/events/event_test.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2021 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package events
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"testing"
+	"time"
+
+	"github.com/tektoncd/triggers/pkg/apis/triggers/v1beta1"
+	"k8s.io/client-go/tools/record"
+)
+
+func TestEmit(t *testing.T) {
+	testcases := []struct {
+		name      string
+		eventType string
+		error     error
+		reason    string
+		wantEvent string
+	}{{
+		name:      "Trigger processing is success",
+		eventType: TriggerProcessingSuccessfulV1,
+		error:     nil,
+		wantEvent: "Normal dev.tekton.event.triggers.successful.v1",
+	}, {
+		name:      "Trigger processing is started",
+		eventType: TriggerProcessingStartedV1,
+		error:     nil,
+		wantEvent: "Normal dev.tekton.event.triggers.started.v1",
+	}, {
+		name:      "EventListener is Ready but some operation failed while serving HTTP req",
+		eventType: TriggerProcessingFailedV1,
+		error:     errors.New("failed to read request body"),
+		wantEvent: "Warning dev.tekton.event.triggers.failed.v1 failed to read request body",
+	}}
+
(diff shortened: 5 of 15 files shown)


vdemeester (approved):

vaikas:
FWIW, I'd caution against sending events for successful reconciliation and even for things that are non-terminal failures. This particular example emits 15 events in 78sec.

Creating events can be a problem in couple of scenarios:
 * Large number of EventListener objects get created in a short amount of time
 * Global resync, for example, when a controller restarts, or if HA is enabled, a leader switches.
 
 In both cases there might be problems with the controller getting ratelimited due to large number of events and hilarity will ensue. We ran into this in Knative and just sharing our experience. In the end we ended up dialing back the number of events we emit because it ended up creating problems.

khrm:
@savitaashture Based on #649, we should emit events when we trigger a resource successfully. This code should go inside EventListener's sink, not reconciliation.

savitaashture:
@khrm @dibyom PR is ready for Review PTAL

config/200-clusterrole.yaml
```diff
@@ -85,6 +85,9 @@ rules:
     resources: ["podsecuritypolicies"]
     resourceNames: ["tekton-triggers"]
     verbs: ["use"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
```
vaikas, line 90: Curious if here and below we need all these permissions? Seems like create would suffice because we only emit events?

savitaashture: Oh yes only create is required
Thanks for the heads up :+1:

config/200-clusterrole.yaml
```diff
@@ -96,4 +99,7 @@ metadata:
 rules:
   - apiGroups: ["triggers.tekton.dev"]
     resources: ["clustertriggerbindings", "clusterinterceptors"]
-    verbs: ["get", "list", "watch"]
\ No newline at end of file
+    verbs: ["get", "list", "watch"]
```
dibyom, line 102: Do we need it for both `eventlistener-roles` and `eventlistener-clusterroles`?

pkg/apis/triggers/v1alpha1/event_listener_types.go
```diff
@@ -261,6 +261,12 @@ func (els *EventListenerStatus) SetConditionsForDynamicObjects(conditions v1beta
 			Message: cond.Message,
 		})
 	}
+
+	els.SetCondition(&apis.Condition{
```
dibyom, line 265: question: why do we need this in this PR?

pkg/reconciler/eventlistener/eventlistener.go
```diff
@@ -342,8 +341,8 @@ func (r *Reconciler) reconcileCustomObject(ctx context.Context, el *v1beta1.Even
 		for _, cond := range customConditions {
 			if cond.Type == apis.ConditionReady {
 				if cond.Status != corev1.ConditionTrue {
-					logging.FromContext(ctx).Warn("custom object is not yet ready")
-					return stdError.New("custom object is not yet ready")
+					logging.FromContext(ctx).Warn(fmt.Sprintf("custom object is not yet ready because %s", cond.Message))
```
dibyom, line 344: Can use Warnf instead of Warn + fmt.Sprintf

pkg/reconciler/events/event.go
```diff
@@ -0,0 +1,55 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"knative.dev/pkg/apis"
+	"knative.dev/pkg/controller"
+)
+
+const (
+	// EventReasonSucceded is the reason set for events about successful completion of EventListener
```
dibyom, line 29: Do we need to make sure the event reasons/types here are the same as the one in #1259?

pkg/sink/sink.go
```diff
@@ -91,6 +94,14 @@ type Response struct {
 	ErrorMessage string `json:"errorMessage,omitempty"`
 }
 
+func (r Sink) emitEvents(ctx context.Context, el *triggersv1.EventListener, err error) {
+	var afterCondition *apis.Condition
+	if el != nil {
+		afterCondition = el.Status.GetCondition(apis.ConditionReady)
```
dibyom, line 100: I don't think we should send the `apis.ConditionReady` as the message for the event. Maybe we can try to be consistent with the cloudevents PR here.

pkg/sink/sink.go
```diff
@@ -111,6 +123,7 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 		log.Errorf("Error reading event body: %s", err)
 		r.recordCountMetrics(failTag)
 		response.WriteHeader(http.StatusInternalServerError)
+		r.emitEvents(r.ContextForEvents, el, err)
```
dibyom, line 126: I think this should be co-ordinated with #1259 as well - can we have a single `r.EmitEvents` that sends both K8sEvents and CloudEvents (if enabled).

pkg/sink/sink.go
```diff
@@ -65,6 +67,7 @@ type Sink struct {
 	// WGProcessTriggers keeps track of triggers or triggerGroups currently being processed
 	// Currently only used in tests to wait for all triggers to finish processing
 	WGProcessTriggers *sync.WaitGroup
+	ContextForEvents  context.Context
```
dibyom, line 70: Minor: It seems like we are creating an eventRecorder, associating it to a context and then storing it in a field in Sink? Wouldn't it be simpler store the eventRecorder in the Sink struct and pass it around? It seems like the only use for the contextForEvents is to extract the recorder from it in the events.Emit function - could we just directly pass around the EventRecorder?

dibyom (commented):
Thanks @savitaashture 

My main comment is around this PR and #1259 - I think we should be consistent in where and how we are emitting the event. As well as be consistent on the type of events that we are emitting (though the exact contents might be different).

config/200-clusterrole.yaml
savitaashture: Verified that adding `events` to `eventlistener-roles` would work

pkg/sink/sink.go
savitaashture: with `el.Status.GetCondition(apis.ConditionReady)` not sending message to event instead we are getting el Status

pkg/sink/sink.go
savitaashture: Yup i agree but i think we can do that once PRs merge

pkg/sink/sink.go
savitaashture: Modified :+1:

pkg/reconciler/events/event.go
savitaashture: Yup for now added similar constant fro #1259 PR

pkg/apis/triggers/v1alpha1/event_listener_types.go
savitaashture: Added as part of this PR because EL for Knative when succeed was not displaying any messages so added this change as part of this PR itself

pkg/sink/sink.go
```diff
@@ -91,17 +95,40 @@ type Response struct {
 	ErrorMessage string `json:"errorMessage,omitempty"`
 }
 
+func (r Sink) emitEvents(recorder record.EventRecorder, el *triggersv1.EventListener, err error) {
+	afterCondition := el.Status.GetCondition(apis.ConditionReady)
+	if el.Annotations[events.EnableEventListenerEvents] == "true" {
```
savitaashture, line 100: @dibyom emitting events only when user sets `enable-eventlistener-events: "true"`  annotation
LMK your opinion on this.
If it looks okay i will update same info in Doc as well
/cc @khrm

dibyom: Yeah that sounds good. This should be an option in the config-features.

pkg/sink/sink.go
```diff
@@ -91,17 +95,40 @@ type Response struct {
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      r.EventListenerName,
+			Namespace: r.EventListenerNamespace,
+			Annotations: map[string]string{
+				// enabled by default for temporary EL
+				events.EnableEventListenerEvents: "true",
+			}}}
+
+	r.emitEvents(r.EventRecorder, &elTemp, nil)
```
dibyom, line 124: Doesn't this mean that we will always emit an event due to `elTemp` always setting the annotation to true? i.e. the actual EL could have it set of false but due to `elTemp` defaulting to true, we are still emitting one event?

pkg/reconciler/events/event.go
```diff
@@ -0,0 +1,59 @@
...
+	"k8s.io/client-go/tools/record"
+	"knative.dev/pkg/apis"
+)
+
+const (
+	// Emit events if user provides enable-eventlistener-events: "true"
+	EnableEventListenerEvents = "enable-eventlistener-events"
+	// TriggerProcessingStarted is sent for Sink triggers received which
+	// are being processed
+	TriggerProcessingStartedV1 = "triggers.tekton.eventlistener.triggerprocessing.started.v1"
```
dibyom, line 30: I'm glad that this is consistent with cloudevents but I had a comment that those names should be similar to the ones for Pipelines - see https://github.com/tektoncd/triggers/pull/1259#discussion_r759553630

pkg/reconciler/events/event.go
```diff
@@ -0,0 +1,59 @@
...
+	TriggerProcessingFailedEventV1 = "triggers.tekton.eventlistener.triggerprocessing.failed.v1"
+)
+
+// Emit emits events for object
+// Supported events are k8s events.
+func Emit(recorder record.EventRecorder, afterCondition *apis.Condition, object runtime.Object, err error) {
+	sendKubernetesEvents(recorder, afterCondition, object, err)
+}
+
+func sendKubernetesEvents(c record.EventRecorder, afterCondition *apis.Condition, object runtime.Object, err error) {
```
dibyom, line 43: I'm still not sure if we need to rely on the status of the EventListener for emitting events in Triggers - if the EL is not ready, we are probably not going to receive the event at all i.e. the server is not up or something. 

IMO it will be clearer emitEvents just sends the eventType (Successful, Failed etc.) as a param instead of the afterCondition.

savitaashture: We had a scenario with `custom-resource` example as part of `0.16.0` Triggers release we had an issue  https://github.com/tektoncd/triggers/issues/1219 related to `metadata.resourceVersion` as part of Knative and because of that EL status was not **True** but pods were up and running so that is the reason i considered this way of using EL status in order to cover such scenarios also

pkg/sink/sink.go
savitaashture: @dibyom do you mean we no need to take from eventlistener annotation instead we should depend on `config-features` cm ?

pkg/sink/sink.go
savitaashture: Yes because from this operation
`el, err := r.EventListenerLister.EventListeners(r.EventListenerNamespace).Get(r.EventListenerName)` there could be chances of error while doing Get 
So added elTemp so that get operation failure captured

pkg/sink/sink.go
dibyom: my bad. I missed that this was an annotation. Just an annotation is fine for now as long as we document it.

pkg/reconciler/events/event.go
savitaashture: Updated :+1:

pkg/reconciler/events/event.go
savitaashture: But agree with you that the current PR focuses on events for Trigger process so we no need to consider eventlistener status instead we can send the type as you suggested

I have updated PR PTAL

pkg/reconciler/events/event_test.go
```diff
@@ -0,0 +1,90 @@
...
+		eventType: TriggerProcessingSuccessfulV1,
+		error:     nil,
+		wantEvent: "Normal dev.tekton.event.triggers.successful.v1",
+	}, {
+		name:      "Trigger processing is started",
+		eventType: TriggerProcessingStartedV1,
+		error:     nil,
+		wantEvent: "Normal dev.tekton.event.triggers.started.v1",
+	}, {
+		name:      "EventListener is Ready but some operatiorn failed while serving HTTP req",
```
dibyom, line 48:
```suggestion
		name:      "EventListener is Ready but some operation failed while serving HTTP req",
```

dibyom (approved):

dibyom:
Few small nits, but otherwise good to go!

examples/v1alpha1/github/github-eventlistener-interceptor.yaml
```diff
@@ -3,6 +3,8 @@ apiVersion: triggers.tekton.dev/v1alpha1
 kind: EventListener
 metadata:
   name: github-listener
+  annotations:
```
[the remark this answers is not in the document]
savitaashture: Its not needed but just as an example for emitting event added annotation

Reverted back

savitaashture:
@dibyom I have addressed all review comments PTAL

PR #1232, opened 2021-10-01
Feature: TriggerGroups

dibyom:
# Changes

This is a port of https://github.com/tektoncd/triggers/pull/1052

```
This feature allows an operator to specify a set of interceptors that will be executed
before a group of triggers are selected and executed. This allows common data
to be passed from interceptor execution down to multiple triggers to solve
a set of common use cases across multiple Triggers.

This feature is enabled for now inline in the EventListener spec, but in the future
may be enabled only in alpha once the feature gates proposal is implemented within
this project.
```

So far I've done the following
- Rebased #1052 with latest main branch
- Removed some old code that returned that waited for triggers to finish processing before returning
- Move the types to v1beta1 (still need to delete it from v1alpha1)
- Fix the correct body/header to triggers within a trigger group
- Ensure we wait for triggers in triggerGroups to process to fix sink tests

**TODO**:

- [ ] Port validation/defaulting to v1beta1
- [ ] Move behind an alpha feature gate
- [ ] Move examples to behind v1beta1 - would have to figure out how to enable feature gate for the example tests

# Submitter Checklist

review them:

# Release Notes


dibyom:
>  Move behind an alpha feature gate

Added validation in the webhook. Should probably add another check in the reconciler too

>  Move examples to behind v1beta1 - would have to figure out how to enable feature gate for the example tests

Due to the allowlist in  https://github.com/tektoncd/triggers/blob/main/test/e2e-tests-examples.sh#L138, the triggerGroup example would not be run. I'll work on something like https://github.com/tektoncd/pipeline/tree/main/docs/developers#example-yamls in a separate PR to enable the YAML test for this example.

dibyom:
I'll add some extra validation in the reconciler so that the alpha fields cannot be used and rebase the commits but in the meantime, this should be ready for a review
/assign @wlynch 
/assign @savitaashture

Issue #1270, opened 2021-11-24
feature: add spec to specify EventListener service port :door: 


waveywaves:
### Feature request

Update EventListener port to something other than port 8080. The EventListener Service has port config similar to the following. This feature request entails allowing the user to change the "port" to something other than 8080 via a spec change.
```
ports:
    - protocol: TCP
      port: 8080 <----- this
      targetPort: ??
```
Currently, ServiceType is something the user can provide to change the type of the service from the default "ClusterIP" to the alternatives (LoadBalancer, NodePort). Similarly, we should also be able to change the port. It can be named ServicePort initially without involving any breaking changes.

### Use case

User wants to expose their Webhooks to a port apart from 8080, cuz they need spice :hot_pepper: in lyf.

dibyom:
> cuz they need spice 🌶️ in lyf.

🤣

PR #1272, opened 2021-11-25
feature: add service port spec for eventlistener

waveywaves:
# Changes

add the ability to specify the service port for a eventlistener
service. the TLS port will default to 443 instead of 8443 in
the case where the service port is set.
this has been done to circumvent the logic of 8443 only being set
when the default port is set with TLS enabled else the servicePort
is used.

# Submitter Checklist

review them:

# Release Notes

```release-note
- API changes
  Added ServicePort specification to kubernetesResource to allow users to modify on which port their eventListener service is exposed on (defaults to 8080).
```


waveywaves:
https://github.com/tektoncd/triggers/issues/1270

pkg/reconciler/eventlistener/resources/service.go
```diff
@@ -99,6 +106,10 @@ func ServicePort(el *v1beta1.EventListener, c Config) corev1.ServicePort {
 			// We return port 8443 if TLS is enabled and the default HTTP port is set.
 			// This effectively makes 8443 the default HTTPS port unless a user explicitly sets a different port.
 			servicePortPort = 8443
+		} else {
+			// Return port 443 if TLS is enabled and the HTTP port is not the default one.
+			// This will allow traffic to be directed to the default HTTPS port.
+			servicePortPort = 443
 		}
```
savitaashture, line 113: @waveywaves How about to remove 
```
if *c.Port == DefaultPort {
```
and just override `servicePortPort` to `8443` instead of confusing with `8443` and `443` 

WDYT?

docs/eventlisteners.md
```diff
@@ -24,6 +24,7 @@ or more [`Interceptors`](./interceptors.md).
 - [Specifying `TriggerGroups`](#specifying-triggergroups)
 - [Specifying `Resources`](#specifying-resources)
   - [Specifying a `kubernetesResource` object](#specifying-a-kubernetesresource-object)
+    - [Specifying `Service` configuration](#specifying-service-configuration)
```
dibyom, line 27: Thanks for adding the docs!

go.sum
```diff
@@ -1036,15 +1036,14 @@ github.com/tektoncd/pipeline v0.27.1-0.20210830154614-c8c729131d4a h1:JoBF2MT3hg
...
-github.com/tidwall/pretty v1.1.0/go.mod h1:XNkn88O1ChpSDQmQeStsy+sBenx6DDtFZJxhVysOjyk=
+github.com/tidwall/gjson v1.10.2 h1:APbLGOM0rrEkd8WBw9C24nllro4ajFuJu0Sc9hRz8Bo=
+github.com/tidwall/gjson v1.10.2/go.mod h1:/wbyibRr2FHMks5tjHJ5F8dMZh3AcwJEMf5vlfC0lxk=
+github.com/tidwall/match v1.1.1 h1:+Ho715JplO36QYgwN9PGYNhgZvoUSc9X2c80KVTi+GA=
+github.com/tidwall/match v1.1.1/go.mod h1:eRSPERbgtNPcGhD8UCthc6PmLEQXEWd3PRB5JTxsfmM=
 github.com/tidwall/pretty v1.2.0 h1:RWIZEg2iJ8/g6fDDYzMpobmaoGh5OLl4AXtGUGPcqCs=
 github.com/tidwall/pretty v1.2.0/go.mod h1:ITEVvHYasfjBbM0u2Pg8T2nJnzm8xPwvNhhsoaGGjNU=
-github.com/tidwall/sjson v1.2.1 h1:r0D/mPikA5YxxFluOftF9DBnwTv9LzY9J4UteHTdh3A=
-github.com/tidwall/sjson v1.2.1/go.mod h1:3nkMFbUMK4z5nlDu1y6g7O+zvjJ6hbyoJUZa4WzepBE=
+github.com/tidwall/sjson v1.2.3 h1:5+deguEhHSEjmuICXZ21uSSsXotWMA0orU783+Z7Cp8=
```
dibyom, line 1045: Was this intentional?

pkg/apis/triggers/v1beta1/event_listener_types.go
```diff
@@ -78,6 +78,7 @@ type CustomResource struct {
 type KubernetesResource struct {
 	Replicas           *int32             `json:"replicas,omitempty"`
 	ServiceType        corev1.ServiceType `json:"serviceType,omitempty"`
+	ServicePort        int32              `json:"servicePort,omitempty"`
```
dibyom, line 81: Minor/Consistency: Use *int32. Generally we use pointer types for nullable/optional fields so that it is easier to distinguish between explicit zero and not specified

pkg/reconciler/eventlistener/resources/service.go
dibyom: Could you explain why we need 443 as the default TLS port?

waveywaves:
> Could you explain why we need 443 as the default TLS port?
After giving it some thought I realized that this may not be a great solution to the TLS port issue. Allowing port override should allow the user to configure the port as 443 if they want to. Having it as a default port is not a great idea fosho.

go.sum
waveywaves: no it wasn't. I will revert this.

pkg/reconciler/eventlistener/resources/service.go
waveywaves: After giving it some thought I realized that this may not be a great solution to the TLS port issue. Allowing port override should allow the user to configure the port as 443 if they want to. Having it as a default port is not a great idea fosho.

pkg/apis/triggers/v1beta1/event_listener_types.go
waveywaves: done

dibyom:
@waveywaves as I mentioned in the WG today, we might want to look into passing the servicePort into the[ EventListenerStatus.SetAddress func](https://github.com/tektoncd/triggers/blob/d9b14bf6441108027f4f4a26314dc906ac21855c/pkg/apis/triggers/v1beta1/event_listener_types.go#L334) so that the `status.address.URL` that is set on the EL contains the right port.

waveywaves:
> @waveywaves as I mentioned in the WG today, we might want to look into passing the servicePort into the[ EventListenerStatus.SetAddress func](https://github.com/tektoncd/triggers/blob/d9b14bf6441108027f4f4a26314dc906ac21855c/pkg/apis/triggers/v1beta1/event_listener_types.go#L334) so that the `status.address.URL` that is set on the EL contains the right port.
> 
> UPDATE: It looks like it might already do the right thing - calls to SetAddress uses resources.ListenerHostname which relies on the ServicePort function that contains the changes.

@dibyom In that case, are there any other changes you think I should make ?

dibyom (approved):

savitaashture (commented):
/lgtm :tada:

waveywaves:
@dibyom

Issue #1268, opened 2021-11-24
Too much memory used with large number of namespaces


guillaumerose:
# Expected Behavior

Same memory usage as the pipeline controller

# Actual Behavior

The memory taken by the interceptor and trigger controller is around 1GB.

This is not the case of the pipeline controller which stay around 50MB.

<img width="1468" alt="Screenshot 2021-11-24 at 15 30 56" src="https://user-images.githubusercontent.com/172624/143257378-c21e36ae-6f86-42a5-9e69-000f8c09c6b2.png">

(memory usage of pods in Tekton namespace during the load test - highest are tekton-triggers-controller and tekton-triggers-core-interceptors pods)

# Steps to Reproduce the Problem

1. Create 2000 namespaces. In the first 600, create a PipelineRun like this:
```
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  name: pipeline1
spec:
  pipelineSpec:
    tasks:
      - name: step1
        taskSpec:
          steps:
          - name: step1
            image: busybox
            imagePullPolicy: IfNotPresent
            script: ls /
```
2. Observe the memory taken by the 2 controllers

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

 ```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.2", GitCommit:"8b5a19147530eaac9476b0ab82980b4088bbc1b2", GitTreeState:"clean", BuildDate:"2021-09-15T21:31:32Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1+d8c4430", GitCommit:"2ade4d46986c94405aa0542f8d6ff7691bddb153", GitTreeState:"clean", BuildDate:"2021-11-02T02:02:53Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"linux/amd64"}
 ```

- Tekton Pipeline version:

pipeline: v0.28.2
triggers: v0.16.1 

- Other additional information 

It is stable: if I kill the pod, the controller is recreated and takes almost the same amount of memory.

The cluster has 681 deployments and 14180 configMaps.

I turned on profiling and saw there is something happening in the informers.

[...]

PR #1274, opened 2021-11-26
Avoid watching all secrets in the cluster

guillaumerose:
# Changes

Replace the secret informer by a secret getter.
It implies that for each webhook an API call will be issued against the
k8s API to get the secret.
Previously, all secrets of the cluster were in the interceptor memory.

Before this change:
crictl stats reports 79.18MB with around 5k secrets of 2.5kB with kind.

After this change:
crictl stats reports 8.221MB

Proposition related to https://github.com/tektoncd/triggers/issues/1268

The other way to solve this would be to add a specific label on all secrets watched by the interceptor.
But it would introduce a breaking change between 2 releases.

# Submitter Checklist

review them:

# Release Notes

```release-note
Reduce memory usage of the core-interceptors container
```

### DIFF
--- a/cmd/cel-eval/cmd/root.go
+++ b/cmd/cel-eval/cmd/root.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bufio"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -14,10 +15,8 @@ import (
 	"github.com/google/cel-go/checker/decls"
 	celext "github.com/google/cel-go/ext"
 	"github.com/spf13/cobra"
+	triggersv1beta1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1beta1"
 	triggerscel "github.com/tektoncd/triggers/pkg/interceptors/cel"
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	corev1lister "k8s.io/client-go/listers/core/v1"
 )
 
 var (
@@ -48,17 +47,9 @@ func rootRun(cmd *cobra.Command, args []string) {
 	}
 }
 
-type secretLister struct{}
+type secretGetter struct{}
 
-func (sl secretLister) List(selector labels.Selector) (ret []*v1.Secret, err error) {
-	return nil, nil
-}
-
-func (sl secretLister) Secrets(namespace string) corev1lister.SecretNamespaceLister {
-	return secretLister{}
-}
-
-func (sl secretLister) Get(name string) (*v1.Secret, error) {
+func (sg secretGetter) Get(ctx context.Context, triggerNS string, sr *triggersv1beta1.SecretRef) ([]byte, error) {
 	return nil, nil
 }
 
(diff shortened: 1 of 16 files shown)


guillaumerose:
I just saw that the Lister was introduced to avoid throttling :/ 
What is the good solution? Happy to code it!

dibyom:
> What is the good solution? Happy to code it!

Yeah this one will require a bit of thinking - 
1. Could we use getters + a custom cache to cache the get requests vs a informer cache that caches all secrets?
2. Could we add a label to all secrets that the EL/Interceptors need? 
 
2. will simplify the implementation but is a breaking change for users so I'm leaning towards 1.

guillaumerose:
Yes I agree. I think the best is to have a single place where the code calls `Secrets(ns).Get(secretName)` (https://github.com/tektoncd/triggers/pull/1278), then we can work on a cache (maybe https://pkg.go.dev/k8s.io/apimachinery/pkg/util/cache#LRUExpireCache ?)

dibyom:
@guillaumerose Nice yeah! A LRU cache should be good enough for our use case

guillaumerose:
I added the LRU cache. It will store 1024 secrets for 5s maximum. That way, it will ensure if the listener receives a batch of notifications from the same repo, it will call once the apiserver.

I don't cache errors. If the secret is incorrect or absent, it will still issue a call to the apiserver.

WDYT?

dibyom:
Looks good. I think 5s might be reasonable to start with though we might this configurable later.

Did we do some testing with a bunch of concurrent requests? One question I have is how effective this will be for a bunch of requests (e.g. for the same event, multiple triggers) that arrive at the same time. Since each incoming request is processes in its own goroutine, I wonder if this will result in all (or majority) cache misses.

savitaashture:
> Looks good. I think 5s might be reasonable to start with though we might this configurable later.
> 
> Did we do some testing with a bunch of concurrent requests? One question I have is how effective this will be for a bunch of requests (e.g. for the same event, multiple triggers) that arrive at the same time. Since each incoming request is processes in its own goroutine, I wonder if this will result in all (or majority) cache misses.
> 
> See [#594 (comment)](https://github.com/tektoncd/triggers/pull/594#issuecomment-651819853)

@dibyom 
As discussed have added a print statement [here](https://github.com/tektoncd/triggers/pull/1274/files#diff-706a824f91e384361f83950baa4b8624bd38da1ec63805d8c15ec8037502a931R81) and sent multiple requests 
to EventListener which had multiple triggers(I use 3 triggers)

```
while true; do curl -H 'X-GitHub-Event: pull_request'    -H 'X-Hub-Signature: sha1=ba0cdc263b3492a74b601d240c27efe81c4720cb'    -H 'Content-Type: application/json'    -d '{"action": "opened", "pull_request":{"head":{"sha": "28911bbb5a3e2ea034daf1f6be0a822d50e31e73"}},"repository":{"clone_url": "https://github.com/tektoncd/triggers.git"}}'    http://localhost:8080; done;
```

I see the cache is working as its configured for 5s i see cache is true for 4-5 times and then its resetting

dibyom (approved):

savitaashture (commented):
/lgtm

:tada:

savitaashture (approved):
/lgtm

khrm (commented):
/lgtm

khrm (commented):
/lgtm

PR #1282, opened 2021-12-08
Fix eventlistener crash with multiple interceptors extensions

larhauga:
When multiple interceptors are present in an event listener and they produce an extension, a race condition can occur due to a global variable written to in a goroutine.

It is possible that this also affects which pipelines could be triggered if the extensions are used for trigger routing.

# Changes

A global shared variable was being written to when extensions were
present in the interceptor output. This resulted in a race condition
verified by `go test -race ./pkg/sink` with the new test.

# Submitter Checklist

review them:

# Release Notes
- Bug fixes

```release-note
Fix race condition in eventlistener when multiple interceptors return extentions
```

### DIFF
--- a/pkg/sink/sink_test.go
+++ b/pkg/sink/sink_test.go
@@ -269,14 +269,12 @@ func trResourceTemplate(t testing.TB) runtime.RawExtension {
 }
 
 func TestHandleEvent(t *testing.T) {
-	var (
-		eventListenerName = "my-el"
-		eventBody         = json.RawMessage(`{"head_commit": {"id": "testrevision"}, "repository": {"url": "testurl"}, "foo": "bar\t\r\nbaz昨"}`)
-		gitCloneTTSpec    = triggersv1beta1.TriggerTemplateSpec{
+	makeGitCloneTTSpec := func(name string) *triggersv1beta1.TriggerTemplateSpec {
+		return &triggersv1beta1.TriggerTemplateSpec{
 			Params: []triggersv1beta1.ParamSpec{
 				{Name: "url"},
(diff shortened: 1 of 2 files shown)


pkg/sink/sink_test.go
```diff
@@ -970,7 +1135,7 @@ func TestHandleEvent(t *testing.T) {
 
 			// Sort TaskRuns when comparing (we do not know what order they were created in)
 			compareTaskRuns := func(x, y pipelinev1.TaskRun) bool {
-				return x.Name < y.Name
+				return x.Name < y.Name && x.Labels["triggers.tekton.dev/trigger"] < y.Labels["triggers.tekton.dev/trigger"]
```
dibyom, line 1138: It seems like we are doing this because we are creating two TaskRuns with the same name. This is not an issue here since we are using FakeClients but for real use cases this will result in an error from the API server. 

We could probably turn `gitCloneTTSpec` into a simple function like `makeTTSpec(name, namespace string)` or something like that. If that is too much for this PR, would you mind opening an issue for this?

dibyom (approved):
Thank you @larhauga This was a great find. 

I have one minor comment on the test. If you are particularly motivated to fix that in this PR, that is fine -- just open an issue for it :) 

/approve

larhauga:
> Thank you @larhauga This was a great find.
> 
> I have one minor comment on the test. If you are particularly motivated to fix that in this PR, that is fine -- just open an issue for it :)

My pleasure! The tests are updated with the your suggestion of a function for the TT.

dibyom:
Thank you! Looks great
/lgtm

PR #1298, opened 2022-01-26
Generate API reference docs for Triggers

abayer:
# Changes

This will add `hack/update-reference-docs.sh` to `hack/update-codegen.sh`, resulting in `docs/triggers-api.md` being generated.

There probably needs to be a followup in the website repo to actually do something with the reference doc too. =)

/kind documentation

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```


hack/reference-docs-template/members.tpl
```diff
@@ -0,0 +1,48 @@
+{{ define "members" }}
```
dibyom, line 1: Once we have this in pipelines as well, it would be nice to have this in plumbing and vendor it from there instead of having to do this is each Tekton sub project.

dibyom (approved):

hack/reference-docs-template/members.tpl
abayer: Agreed!

abayer:
@dibyom I had to rebase and re-run `hack/update-codegen.sh`, but it should hopefully be good now. =)

PR #1333, opened 2022-03-24
Add changes to run clusterinterceptor as HTTPS

savitaashture:
# Changes

TEP: https://github.com/tektoncd/community/pull/662

# Submitter Checklist

review them:

# Release Notes

```release-note
Triggers now support end to end secure connection by running ClusterInterceptor server as HTTPS 
```


cmd/interceptors/main.go
```diff
@@ -72,8 +89,62 @@ func main() {
 	mux.Handle("/", service)
 	mux.HandleFunc("/ready", handler)
 
+	name := os.Getenv("SVC_NAME")
```
dibyom, line 92: IMO we should give this a more explicit name than SVC name e.g. INTERCEPTOR_TLS_SECRET or something like that

cmd/interceptors/main.go
```diff
@@ -19,25 +19,36 @@ package main
...
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/signals"
+	certresources "knative.dev/pkg/webhook/certificates/resources"
 )
 
 const (
-	// Port is the port that the port that interceptor service listens on
-	Port         = 8082
+	// HTTPSPort is the port where interceptor service listens on
+	HTTPSPort    = 8443
```
dibyom, line 44: question: is it necessary to change the port?

cmd/interceptors/main.go
```diff
@@ -72,8 +89,62 @@ func main() {
...
+			Namespace: namespace,
+		},
+		Data: map[string][]byte{
+			certresources.ServerKey:  serverKey,
+			certresources.ServerCert: serverCert,
+			certresources.CACert:     caCert,
+		},
+	}
+
+	secret.Data = newSecret.Data
```
dibyom, line 126: We could simplify this to 
```
secret.Data = map[string][]byte{
			certresources.ServerKey:  serverKey,
			certresources.ServerCert: serverCert,
			certresources.CACert:     caCert,
		},
```

config/200-clusterrole.yaml
```diff
@@ -65,7 +65,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["secrets"]
-    verbs: ["get", "list", "watch"]
+    verbs: ["get", "list", "watch", "update"]
```
dibyom, line 68: I don't think we should change the role to be able to update all secrets. Instead create another rule that can update the specific secret that we need

config/interceptors/interceptor-secrets.yaml
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: Secret
+metadata:
+  name: tekton-triggers-core-interceptors
```
dibyom, line 18: Should we add a certs suffix to make it clearer what this does?
```suggestion
  name: tekton-triggers-core-interceptors-certs
```

pkg/apis/triggers/v1alpha1/cluster_interceptor_types.go
```diff
@@ -112,12 +112,12 @@ func (it *ClusterInterceptor) ResolveAddress() (*apis.URL, error) {
...
 		return nil, ErrNilURL
 	}
-	port := defaultPort
+	port := defaultHTTPSPort
 	if svc.Port != nil {
 		port = *svc.Port
 	}
 	url := &apis.URL{
-		Scheme: "http", // TODO: Support HTTPs if caBundle is present
+		Scheme: "https",
```
dibyom, line 120: hmm..we'd have to be a bit careful here...if we move this to only support `https`, anyone who has written a cluster interceptor will not be able use Triggers until they upgrade their code. Should we support both for a release or two?

cmd/interceptors/main.go
dibyom, line 92: There is quite a bit of logic here...I'd move it to its own function for clarity.

dibyom (commented):
At a high level, the implementation looks good. I have some minor comments.

The biggest thing on my mind is that since we are making this change "by default", anyone who has written their own cluster interceptor will be broken until they update their code to HTTPS.

An alternative would be to update the CRD to specify the scheme (HTTP by default which we remove and change after a couple of releases). I see no issue with us switching to HTTPS for the core interceptors in this release itself.

config/200-clusterrole.yaml
savitaashture: `tekton-triggers-core-interceptors` cluster role is only used for `tekton-triggers-core-interceptors` deployment so i believe this is for specified one only

Let me know if this is your concern or something else

pkg/apis/triggers/v1alpha1/cluster_interceptor_types.go
savitaashture: Ah i see

Okay so do you mean we need to have support for both HTTP and HTTPS right/

If thats the case we need to take input from deplpyment yaml to know whether they want HTTP or HTTPS
Shall i change implementation that way ?

cmd/interceptors/main.go
savitaashture: In order to make it sync with `https` as `https` usually run on `8443` or `443` so just modified

cmd/interceptors/main.go
savitaashture: Modified 
PTAL

config/200-clusterrole.yaml
dibyom: Right the concern is that this allows the core interceptor server access to update any secret in the cluster which is a bit scary. What we should do is to make sure it can only update the TLS secret for the cluster interceptor.
See the example here where the webhook role can only update the webhook-certs secret using the `resourceName` field: https://github.com/tektoncd/pipeline/blob/main/config/200-role.yaml#L59-L65

pkg/sink/initialization.go
```diff
@@ -166,12 +168,36 @@ func ConfigureClients(ctx context.Context, clusterConfig *rest.Config) (Clients,
...
 		K8sClient:       kubeClient,
 		CEClient:        ceClient,
+		CertData:        certData,
 	}, nil
 }
+
+// getCertificateData is used to get certData which is used by eventlistener while connecting to clusterinterceptor
+func getCertificateData(ctx context.Context, kubeClient *kubeclientset.Clientset) ([]byte, error) {
+	var certData []byte
+	allSecrets, err := kubeClient.CoreV1().Secrets("").List(ctx, metav1.ListOptions{})
```
dibyom, line 188: why are we listing all secrets instead of just a get call given that we already know the name?

dibyom:
So, going through this again, it seems like the implementation is really designed to support HTTPS for the core interceptors that we already package in vs any other cluster interceptors that others might have written. Is the scope of this PR only to support HTTPS for the core cluster interceptors or for any cluster interceptor? 

If it is the former, we should make sure we also support also support HTTP for any non core cluster interceptors. 

Whether in this PR or in a future one, we should update the implementation to support HTTPS for all cluster interceptors.
To do so, we'd need to add a `caCert` field in the clusterinterceptor CRD, the EL should look for this field and use the value of the `caCert` field from the CRD to populate the certificatedata, and then use that in the HTTP client (vs querying secrets for this)

pkg/apis/triggers/v1alpha1/cluster_interceptor_types.go
dibyom: we discussed this in the WG -  we should take the input for whether it is HTTP/HTTPS from interceptor CRD itself.

pkg/sink/initialization.go
dibyom: also the fact that the EL has to know the exact secret name for the core interceptors seems not ideal. Ideally, the EL should be able to pick up where to pick up the certData from the CRD itself (e.g. the webhooks contain a caCert field with this info)

pkg/apis/triggers/v1alpha1/cluster_interceptor_defaults.go
```diff
@@ -18,17 +18,7 @@ package v1alpha1
 
 import (
 	"context"
-
-	"github.com/tektoncd/triggers/pkg/apis/triggers/contexts"
 )
 
 // SetDefaults sets the defaults on the object.
-func (it *ClusterInterceptor) SetDefaults(ctx context.Context) {
```
savitaashture, line 26: @dibyom Removed setting default port because earlier if no port specified the default port 80 use to set to svc.Port using setDefaults
But now default port will change based on caBundle and caBundle for clusterintercptor will be populated little late than webhook execution

So thats the reason removed defaulting from here and added as part of [cluster_interceptor_types.go](https://github.com/tektoncd/triggers/pull/1333/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-dd761b819fad42dcd5732dbd8fc12671dd1caa5ad8b0d8bd1870a92861e9a303R130-R139) 

Scenario :

1. if port specified as part of clusterintercepor spec then use that port
2. If no port specified and cabundle is specified then use `8443` port as default
3. If no port and cabundle is specified then use `80` port as default

pkg/interceptors/interceptors.go
```diff
@@ -159,7 +162,28 @@ func ResolveToURL(getter InterceptorGetter, name string) (*apis.URL, error) {
...
+	// So adding check to make sure its https before creating client for https.
+	if strings.Contains(url, "https") {
+		certPool := x509.NewCertPool()
+		if ok := certPool.AppendCertsFromPEM(certData); !ok {
+			return nil, fmt.Errorf("unable to parse cert from %s", certData)
+		}
+
+		client = &http.Client{
+			Transport: &http.Transport{
+				// disable "G402 (CWE-295): TLS MinVersion too low. (Confidence: HIGH, Severity: HIGH)"
```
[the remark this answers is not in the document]
savitaashture: @dibyom 
This is the build issue
```
pkg/interceptors/interceptors.go:182:23: G402: TLS MinVersion too low. (gosec)
				TLSClientConfig: &tls.Config{
					RootCAs: certPool,
				},
```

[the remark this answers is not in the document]
savitaashture: @dibyom added [MinVersion](https://github.com/tektoncd/triggers/pull/1333/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-706a824f91e384361f83950baa4b8624bd38da1ec63805d8c15ec8037502a931R184) to avoid build failure
Ref: https://golang.hotexamples.com/examples/crypto.tls/Config/MinVersion/golang-config-minversion-method-examples.html

PTAL
Thank you

cmd/interceptors/main.go
```diff
@@ -72,8 +91,30 @@ func main() {
...
+	keyFile, certFile, caCert, err := getCerts(ctx, secretLister, kubeClient)
+	if err != nil {
+		return
+	}
+
+	tc, err := triggersclientset.NewForConfig(cfg)
+	if err != nil {
+		return
+	}
+	clusterInterceptorList, err := tc.TriggersV1alpha1().ClusterInterceptors().List(ctx, metav1.ListOptions{})
```
dibyom, line 103: This should be in its own function e.g. updateCRDWithCaCert or something like that

cmd/interceptors/main.go
```diff
@@ -82,13 +123,64 @@ func main() {
...
+			// that's responsible for install/updates.  We simply populate the
+			// secret information.
+			log.Printf("secret %s is missing", interceptorSecretName)
+			return "", "", []byte{}, err
+		}
+		log.Printf("error accessing certificate secret %q: %v", interceptorSecretName, err)
+		return "", "", []byte{}, err
+	}
+
+	serverKey, serverCert, caCert, err := certresources.CreateCerts(ctx, interceptorSvcName, namespace, time.Now().Add(decade))
```
dibyom, line 156: Question: Wouldn't this create a new cert on each restart - Also if we have multiple replicas of the Core interceptor server running, wouldn't they overwrite each other? I think we might want to see if the certs already exist and are valid, if they do then we should just use them instead of creating a new one each time

pkg/interceptors/interceptors_test.go
```diff
@@ -515,33 +552,114 @@ func TestExecute_Error(t *testing.T) {
...
...iR2x1WlhNdWMzWmpNQ0FYRFRJeU1EUXhOVEUyCk1ERTFPRm9ZRHpJeE1qSXdNekl5TVRZd01UVTRXakJYTVJRd0VnWURWUVFLRXd0cmJtRjBhWFpsTG1SbGRqRS8KTUQwR0ExVUVBeE0yZEdWcmRHOXVMWFJ5YVdkblpYSnpMV052Y21VdGFXNTBaWEpqWlhCMGIzSnpMblJsYTNSdgpiaTF3YVhCbGJHbHVaWE11YzNaak1Ga3dFd1lIS29aSXpqMENBUVlJS29aSXpqMERBUWNEUWdBRUFHcHp1RjlQCjY5VnFhN0xIY0tmNGpWY2JqblJNWDAxYWRnakh0Zy9kZFdIaVBWdXVJZER1WnZzVTREaVp5Smh2WnpmaHQ0ZmsKT3FJc3dJeVlmbkpLRnFPQ0FVWXdnZ0ZDTUE0R0ExVWREd0VCL3dRRUF3SUNoREFkQmdOVkhTVUVGakFVQmdncgpCZ0VGQlFjREFRWUlLd1lCQlFVSEF3SXdEd1lEVlIwVEFRSC9CQVV3QXdFQi96QWRCZ05WSFE0RUZnUVVQRXFjCnEvRFJHd2FDUTdmOFc0dmlucGN5a09zd2dlQUdBMVVkRVFTQjJEQ0IxWUloZEdWcmRHOXVMWFJ5YVdkblpYSnoKTFdOdmNtVXRhVzUwWlhKalpYQjBiM0p6Z2pKMFpXdDBiMjR0ZEhKcFoyZGxjbk10WTI5eVpTMXBiblJsY21ObApjSFJ2Y25NdWRHVnJkRzl1TFhCcGNHVnNhVzVsYzRJMmRHVnJkRzl1TFhSeWFXZG5aWEp6TFdOdmNtVXRhVzUwClpYSmpaWEIwYjNKekxuUmxhM1J2Ymkxd2FYQmxiR2x1WlhNdWMzWmpna1IwWld0MGIyNHRkSEpwWjJkbGNuTXQKWTI5eVpTMXBiblJsY21ObGNIUnZjbk11ZEdWcmRHOXVMWEJwY0dWc2FXNWxjeTV6ZG1NdVkyeDFjM1JsY2k1cwpiMk5oYkRBS0JnZ3Foa2pPUFFRREFnTkpBREJHQWlFQTlhWFBtUFZzRVA3R0xTbzI0SnNmNnRGTmpyQWJRbEl0CjRCYXllcjBnaU5jQ0lRQ09XSm1NTXQxQkE1RXgwa0FYTWRtZjlFdXV4LzlyUUkzMm9VNjVSYm9mNEE9PQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg=="),
```
dibyom, line 570: minor: could we extract this to a const and use that instead of copying this large string everywhere?

cmd/interceptors/main.go
```diff
@@ -82,13 +111,108 @@ func main() {
...
+		log.Printf("secret %q is missing key %q", secret.Name, certresources.ServerCert)
+		createCerts = true
+	}
+	caCert, haskey = secret.Data[certresources.CACert]
+	if !haskey {
+		log.Printf("secret %q is missing key %q", secret.Name, certresources.CACert)
+		createCerts = true
+	}
+
+	// TODO: Certification validation and rotation is pending
```
savitaashture, line 185: @dibyom will create an issue to handle it in separate PR

cmd/interceptors/main.go
savitaashture: Yes you are right

As discussed in Wg call added condition check to verify existence of certs and if it is there no need to create

**Pending:** Certs validation and Rotation will create an issue to track this

Thank you :+1:

pkg/interceptors/interceptors.go
```diff
@@ -159,7 +162,31 @@ func ResolveToURL(getter InterceptorGetter, name string) (*apis.URL, error) {
...
+	}
+	// There is a chances that url may come from ClusterInterceptorSpec/ClientConfig of URL or Service
+	// So adding check to make sure its https before creating client for https.
+	if strings.Contains(url, "https") {
+		certPool := x509.NewCertPool()
+		if ok := certPool.AppendCertsFromPEM(ic.Spec.ClientConfig.CaBundle); !ok {
+			return nil, fmt.Errorf("unable to parse cert from %s", ic.Spec.ClientConfig.CaBundle)
+		}
+
+		client = &http.Client{
```
dibyom, line 180: I think we have to update this - creating a new client here means we will ignore all of the config options that we set in the already passed in HTTP client (such as the timeout values). Creating a new HTTP client per call means we can't reuse the underlying connections so we might be limiting the number of calls from the EL to the interceptor.

I think we'd have to add in the certs in the `sinker.Start` function - we could pass in a `clusterinterceptorlister` as an arg to the sinker struct - list all the clusterinterceptors, get the caCert and add it to the certPool. and then create a httpClient with all the certs. (This does mean that if we add a new interceptor in the middle, we'd have to restart each EL.


(Ideally in the future we'd have something that listens for changes to interceptors and updates the httpClient with newCerts on the fly)

cmd/interceptors/main.go
dibyom: could you open an issue to track?

cmd/interceptors/main.go
```diff
@@ -82,13 +111,108 @@ func main() {
...
+	secret, err := secretLister.Secrets(namespace).Get(interceptorSecretName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// The secret should be created explicitly by a higher-level system
+			// that's responsible for install/updates.  We simply populate the
+			// secret information.
+			log.Printf("secret %s is missing", interceptorSecretName)
+			return "", "", []byte{}, err
+		}
+		log.Printf("error accessing certificate secret %q: %v", interceptorSecretName, err)
```
dibyom, line 161: can we use logger as we do in the rest of the function vs log.Printf

dibyom (commented):
Almost there - I do think we need to modify how we are creating the http clients else we will break timeouts

cmd/interceptors/main.go
```diff
@@ -82,13 +111,108 @@ func main() {
...
+	}
+
+	// write serverKey to file so that it can be passed while running https server.
+	if err = ioutil.WriteFile(keyFile, serverKey, 0600); err != nil {
+		logger.Errorf("failed to write serverKey file %v", err)
+		return "", "", []byte{}, err
+	}
+	// write serverCert to file so that it can be passed while running https server.
+	// Expect WriteFile permissions to be 0600 or less (gosec)
+	//nolint:gosec
```
dibyom, line 212: Hmm why do we have 0600?

dibyom (approved):
Just one question, else LGTM!

cmd/interceptors/main.go
savitaashture: if given file permission is more than 600 the its giving security issue during build
```
cmd/interceptors/main.go:211:11: G306: Expect WriteFile permissions to be 0600 or less (gosec)
	if err = ioutil.WriteFile(certFile, serverCert, 0644); err != nil {
	         ^
```
For now instead of using 644 i m using 600 and removing `//nolint:gosec` 
Updated main.go 

Thanks for pointing it out :+1: I forgot to remove it earlier

savitaashture:
@dibyom PR is ready for final review

Thank You!!

dibyom (approved):

PR #1359, opened 2022-05-06
Avoid watching all secrets in the cluster

khrm:
# Changes
Replace the secret informer by a secret getter.
It implies that for each webhook an API call will be issued against the
k8s API to get the secret.
Previously, all secrets of the cluster were in the interceptor memory.

Before this change:
crictl stats reports 79.18MB with around 5k secrets of 2.5kB with kind.

After this change:
crictl stats reports 8.221MB

Proposition related to https://github.com/tektoncd/triggers/issues/1268

The other way to solve this would be to add a specific label on all secrets watched by the interceptor.
But it would introduce a breaking change between 2 releases.

Author - @guillaumerose

# Submitter Checklist

review them:

# Release Notes


khrm:
Closed this as I can directly push to  Guillaume Rose's branch.

PR #1376, opened 2022-06-15
WIP: Add Namespaced Scope Interceptor

### LINKED ISSUE #1364: Namespace scoped Interceptor
### Feature request
Today, interceptors can be configured directly in the EventListener definition, which allows Teams to have their own/custom interceptors configured and managed in their namespace without the need of having any specific permissions (only admin permissions in the namespace).

With [TEP-0026: Pluggable Interceptors](https://github.com/tektoncd/community/blob/main/teps/0026-interceptor-plugins.md) this will go away (as per Any current Interceptor configuration is still usable (until beta) in the new model without requiring any manual changes on part of the end-user. As part of the beta, we can deprecate the old syntax.) and therefore require elevated privileges to create/register their custom interceptors.

In OpenShift Container Platform 4 - Clusters hosting multiple tenants, this will cause massive disruptions as specific tenants will have difficulties registering/creating custom interceptors depending on the permissions given or else they will be shared with different tenants which are also not acceptable.

It's therefore requested to either re-consider the deprecation of the current functionality where the interceptors can be configured directly in EventListener or else provide an approach with the new implementation that allows to use it on tenant level, where the resources are completely isolated on namespace level.

The key goal should be that the namespace admin can create and manage its interceptors without requiring elevated permissions and that the resources are restricted and thus not visible to other tenants and therefore namespaces.

Request and Written by: Simon Reber (Red Hat)

khrm:
Added Namespaced Scope Interceptor. A namespaced admin can create and
manage its interceptor without requiring elevated permissions. Fixes #1364 

# Changes

# Submitter Checklist

review them:

# Release Notes


khrm:
Opening a new pr with different CRD name.

Issue #1387, opened 2022-07-06
Non HTTPs custom cluster interceptor broken in v0.20.1


afrittoli:
# Expected Behavior

After installing v0.20.1 I can continue to use my non-HTTPs custom cluster interceptor

# Actual Behavior

It looks like triggers reconciles a `caBundle` in the interceptor definition, which results in:

```
{"level":"error","ts":"2022-07-05T09:31:46.920Z","logger":"eventlistener","caller":"sink/sink.go:289","msg":"Post \"https://build-id-interceptor.tekton-ci.svc:80/buildid\": http: server gave HTTP response to HTTPS client","eventlistener":"tekton-ci","namespace":"tekton-ci","/triggers-eventid":"fdb22324-7af5-4af3-ab26-3955539b5baf","eventlistenerUID":"52ff3801-1e0b-4a68-b0af-3d9113cf345d","/triggers-eventid":"fdb22324-7af5-4af3-ab26-3955539b5baf","/triggergroup":"github-tektoncd-issue-comment-group","stacktrace":"github.com/tektoncd/triggers/pkg/sink.Sink.processTriggerGroups\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:289\ngithub.com/tektoncd/triggers/pkg/sink.Sink.HandleEvent.func2\n\tgithub.com/tektoncd/triggers/pkg/sink/sink.go:206"}
```

I manually delete the `caBundle` from my interceptor and things started working again for some time.
After a while things stopped working again, and I realised that the `caBundle` was back in.
I edited the interceptor again. Things started working again and then stopped working. This time no `caBundle` in the interceptor, but the event listener pod is in crash loop:

```
[...]

Issue #1374, opened 2022-06-07
Eventlistener using wrong URL for clusterinterceptor


quant-daddy:
# Expected Behavior
Event listener should use right address for interceptors. I have configured ClusterInterceptors to use `https://tekton-triggers-core-interceptors.tekton-pipelines.svc:8443/cel`
```yaml
apiVersion: triggers.tekton.dev/v1alpha1
kind: ClusterInterceptor
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"triggers.tekton.dev/v1alpha1","kind":"ClusterInterceptor","metadata":{"annotations":{},"name":"cel"},"spec":{"clientConfig":{"service":{"name":"tekton-triggers-core-interceptors","namespace":"tekton-pipelines","path":"cel"}}}}
  creationTimestamp: "2022-06-07T19:52:29Z"
  generation: 4
  name: cel
  resourceVersion: "498224521"
  uid: 37c35519-10c6-4784-9eda-7d09b40d890a
spec:
  clientConfig:
    caBundle: <redcated>
    url: https://tekton-triggers-core-interceptors.tekton-pipelines.svc:8443/cel
status:
  address:
    url: https://tekton-triggers-core-interceptors.tekton-pipelines.svc:8443/cel
```
# Actual Behavior
The listener may be the wrong URL or there is some issue with throttling?
```
el-listener-interceptor-79d9648974-hwshf event-listener I0607 22:39:07.736328       1 request.go:665] Waited for 1.194655471s due to client-side throttling, not priority and fairness, request: GET:https://10.100.0.1:443/apis/triggers.tekton.dev/v1alpha1/clusterinterceptors/cel
```
The service URL for `tekton-triggers-core-interceptors` service is below:
```
tekton-triggers-core-interceptors   ClusterIP   10.100.14.118   <none>        8443/TCP                             168m
```

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

 ```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:51:05Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"darwin/amd64"}
[...]

Issue #1284, opened 2021-12-09
request.go:668] Waited for *s due to client-side throttling, not priority and fairness, request: GET:https://172.20.0.1:443/apis/triggers.tekton.dev/v1alpha1/clusterinterceptors/github


qtheya:
# Expected Behavior
Event listener is working as expected
# Actual Behavior
Drasticall performance degradation
 ```1 request.go:668] Waited for 11.393899312s due to client-side throttling, not priority and fairness, request: GET:https://172.20.0.1:443/apis/triggers.tekton.dev/v1alpha1/clusterinterceptors/github```
# Steps to Reproduce the Problem

1. Install Operator > 0.51.1
2.
3.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

 ```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.4", GitCommit:"b695d79d4f967c403a96986f1750a35eb75e75f1", GitTreeState:"clean", BuildDate:"2021-11-17T15:48:33Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.2-eks-06eac09", GitCommit:"5f6d83fe4cb7febb5f4f4e39b3b2b64ebbbe3e97", GitTreeState:"clean", BuildDate:"2021-09-13T14:20:15Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}
```
- Tekton Pipeline version:

  **v0.29.0**

As i can see, after tekton-trigger has been moved to knative (v 0.17.x +) it got performance degradation. All webhooks from github are timed out
![image](https://user-images.githubusercontent.com/9653011/145374924-a2d4608f-9c51-46ac-8cbc-3c713269451f.png)
Pipelines run anyway but with start delay ~ 2 minutes. This is time when EventListener is parsing webhook and is trying to match CEL condition. In logs it looks very slow
```el-github-listener-interceptor-6b86d6f674-dp5hh event-listener {"level":"info","ts":"2021-12-09T09:57:19.693Z","logger":"eventlistener","caller":"sink/sink.go:301","msg":"... type push is not allowed","eventlistener":"github-listener-interceptor"...
[...]

PR #1389, opened 2022-07-07
[WIP] Fix http based custom interceptor connection issue

savitaashture:
# Changes
Fixes : https://github.com/tektoncd/triggers/issues/1387

# Submitter Checklist

review them:

# Release Notes


cmd/interceptors/main.go
```diff
@@ -131,10 +131,12 @@ func updateCRDWithCaCert(ctx context.Context, cfg *rest.Config, caCert []byte) e
 		return err
 	}
 	for i := range clusterInterceptorList.Items {
-		if bytes.Equal(clusterInterceptorList.Items[i].Spec.ClientConfig.CaBundle, []byte{}) {
-			clusterInterceptorList.Items[i].Spec.ClientConfig.CaBundle = caCert
-			if _, err := tc.TriggersV1alpha1().ClusterInterceptors().Update(ctx, &clusterInterceptorList.Items[i], metav1.UpdateOptions{}); err != nil {
-				return err
+		if v, ok := clusterInterceptorList.Items[i].GetLabels()["part-of"]; ok && v == "core-interceptors" {
```
dibyom, line 134: Using labels work but I don't think they are a good long term solution - anyone can add/modify labels + any other interceptor that wants to reuse this logic (e.g. Andrea's buildid one) will have to add this core-interceptor label

config/interceptors/core-interceptors.yaml
```diff
@@ -16,6 +16,9 @@ apiVersion: triggers.tekton.dev/v1alpha1
 kind: ClusterInterceptor
 metadata:
   name: cel
+  labels:
+    server/type: https
```
dibyom, line 20: If we need this, then I think we should make it part of the spec and not a label

dibyom (commented):
I think we should consider moving the getCerts and updateCRDwithCaCert functions to pkg/interceptors/server.  The main package contains a lot of logic at this point which is not ideal and I think moving it will make it a lot easier to test.

cmd/interceptors/main.go
savitaashture: I believe that for custom interceptors like  `buildid` (e.g. Andrea's buildid one)  if runs as `https` then its the user responsibility to create certs and give it in CI caBundle and in those cases core-interceptor label not needed
Let me know if this assumption is not correct 

also `core-interceptor` label added to differentiate from core interceptors

config/interceptors/core-interceptors.yaml
savitaashture: Sure will do that

savitaashture:
@dibyom PR is ready for the review

pkg/adapter/adapter.go
```diff
@@ -107,25 +108,27 @@ func (s *sinker) getHTTPClient() (*http.Client, error) {
 	var (
 		caCert    []byte
 		tlsConfig *tls.Config
+		count     int
 	)
 
 	certPool := x509.NewCertPool()
 	err := wait.PollImmediate(interval, timeout, func() (bool, error) {
-		clusterInterceptorList, err := clusterinterceptorsinformer.Get(s.injCtx).Lister().List(labels.NewSelector())
+		clusterInterceptorList, err := s.Clients.TriggersClient.TriggersV1alpha1().ClusterInterceptors().List(s.injCtx, metav1.ListOptions{
```
dibyom, line 116: Could we use a lister here instead of making a direct API call - this might fix the issue described in https://github.com/tektoncd/triggers/issues/1374#issuecomment-1174421352 and https://github.com/tektoncd/triggers/issues/1284#issuecomment-1173683986

dibyom (commented):
Thanks for making the fixes @savitaashture I think the only remaining thing would be to use a lister instead of making a direct API server call so that we do not keep getting throttled by the API server

khrm (approved):

cmd/interceptors/main.go
khrm: I think we can provide the same auto certificate creation feature that's available to the core interceptor depending upon some annotation/labels. It's a useful feature. This label/annotation can be added to the core-interceptor.

config/interceptors/core-interceptors.yaml
khrm: I think having `TLS` only isn't good. There will always be user who don't want added benefit of TLS for their microservice running in private servers. We should keep `HTTP` also. So having a field is good.

pkg/adapter/adapter.go
savitaashture: Latest code have lister only https://github.com/tektoncd/triggers/blob/main/pkg/adapter/adapter.go#L114 but still users getting issue related to timeout 
Do we need to check some other places :thinking:

savitaashture: Updated to use listers

dibyom:
> Error from server (InternalError): error when creating "https://github.com/knative/net-kourier/releases/download/v0.26.0/kourier.yaml": Internal error occurred: failed calling webhook "config.webhook.serving.knative.dev": failed to call webhook: Post "[https://webhook.knative-serving.svc:443/config-validation?timeout=10s](https://webhook.knative-serving.svc/config-validation?timeout=10s)": no endpoints available for service "webhook"

config/interceptors/core-interceptors.yaml
savitaashture: As discussed in Triggers WG call
As  part of this PR we are keeping label itself if require may be we can add as a field as part of another PR

dibyom:
Thanks @savitaashture Mind adding a release-note describing the new label/ and the difference in caCert behavior?

/lgtm

savitaashture:
> Thanks @savitaashture Mind adding a release-note describing the new label/ and the difference in caCert behavior?
> 
> /lgtm

Thank you

Updated Description

Issue #1396, opened 2022-07-19
Adding ClusterInterceptor requires restart of tekton-triggers-core-interceptors to get caBundle


Fabian-K:
# Expected Behavior
When adding a new ClusterInterceptor, the caBundle should automatically be added

# Actual Behavior
caBundle is only added when tekton-triggers-core-interceptors is restarted. I also *think* that in case the cert is deleted and a new one is created, existing ClusterInterceptors are not updated and keep the old caBundle.

# Steps to Reproduce the Problem

1. Deploy tekton triggers
2. wait for certificates to be generated (tekton-triggers-core-interceptors-certs)
3. Add a new ClusterInterceptor

# Additional Info

- Kubernetes version: `1.22.9`
- Tekton Triggers version: `0.20.1`

Fabian-K:
Hi @savitaashture 
I did not even think of custom interceptors but it perfectly makes sense that triggers do not maintain `caBundle` for them. 

The behavior I mentioned however also applies to the core interceptors. The example might be an edge case, however, if you (accidentally?) delete one of the core interceptors and then re-add them by applying the release, it will only get the caBundle when restarting `tekton-triggers-core-interceptors`

PR #1397, opened 2022-07-20
Add watching mechanism to watch on caBundle for core interceptor

savitaashture:
# Changes

Fixes : https://github.com/tektoncd/triggers/issues/1396

# Submitter Checklist

review them:

# Release Notes

```release-note
Added watching mechanism to watch on caBundle for core interceptor
```

### DIFF
--- a/cmd/interceptors/main.go
+++ b/cmd/interceptors/main.go
@@ -92,14 +92,24 @@ func main() {
 		return
 	}
 
-	clusterInterceptorList, err := clusterinterceptorsinformer.Get(ctx).Lister().List(labels.NewSelector())
-	if err != nil {
+	if err := listAndUpdateClusterInterceptorCRD(ctx, tc, service, caCert); err != nil {
 		return
 	}
-
-	if err := service.UpdateCRDWithCaCert(ctx, tc.TriggersV1alpha1(), clusterInterceptorList, caCert); err != nil {
-		return
-	}
+	ticker := time.NewTicker(time.Minute)
+	quit := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
+				if err := listAndUpdateClusterInterceptorCRD(ctx, tc, service, caCert); err != nil {
+					return
+				}
+			case <-quit:
+				ticker.Stop()
+				return
+			}
+		}
+	}()
 
 	srv := &http.Server{
 		Addr: fmt.Sprintf(":%d", HTTPSPort),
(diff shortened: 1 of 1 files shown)


cmd/interceptors/main.go
```diff
@@ -92,14 +92,20 @@ func main() {
...
-		return
-	}
-
-	if err := service.UpdateCRDWithCaCert(ctx, tc.TriggersV1alpha1(), clusterInterceptorList, caCert); err != nil {
-		return
-	}
+	go func() {
+		for {
+			select {
+			case <-time.NewTicker(time.Second).C:
```
dibyom, line 98: every second might be too aggressive....maybe every few seconds or so

savitaashture: If i keep every few seconds then during initial only it will take time 

So i just added for loop instead of select with everysecond

dibyom: A for loop without a delay would be even more aggressive (and kind of unnecessary) - I think the 1 second delay between refreshes was better.

Ideally, we'd separate out the initial cert update from the one that runs in a loop but we can add that in later

savitaashture: Updated 

PTAL 
Thank you

dibyom (approved):

khrm (commented):
/lgtm

cmd/interceptors/main.go
```diff
@@ -92,14 +92,24 @@ func main() {
...
-	clusterInterceptorList, err := clusterinterceptorsinformer.Get(ctx).Lister().List(labels.NewSelector())
-	if err != nil {
-		return
-	}
-
-	if err := service.UpdateCRDWithCaCert(ctx, tc.TriggersV1alpha1(), clusterInterceptorList, caCert); err != nil {
+	if err := listAndUpdateClusterInterceptorCRD(ctx, tc, service, caCert); err != nil {
 		return
 	}
+	ticker := time.NewTicker(time.Minute)
```
dibyom, line 98: time.Minute might be a bit too high for the initial cert fetch. maybe 10 * time.Second?

savitaashture: Initial cert fetch are done [here](https://github.com/tektoncd/triggers/pull/1397/files#diff-ffaa87bda08175ee16b659d2b32cc08cb3a71e1f24797ba38b5d32619f76b963R95) before starting ticker so kept 1 minute to sync/update if there are any core interceptor found without `caBundle` .

khrm (commented):
/lgtm

cmd/interceptors/main.go
dibyom: Ah I missed that. Thank you.

PR #1469, opened 2022-10-19
Respond appropriately to CloudEvent requests

bigkevmcd:
# Changes
This changes the behaviour of the triggers Sink to respond to CloudEvents with a CloudEvent response.

Previously the code always responded with a standard HTTP application/json response, which could cause issues for CloudEvents clients leading to errors and possibly retries.

The handler detects CloudEvent requests and responds with the same body in response.

# Submitter Checklist

review them:

# Release Notes

```release-note
CloudEvents sent to a Trigger will now receive a valid CloudEvent response.
```

### DIFF
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -27,6 +27,10 @@ import (
 	"os"
 	"sync"
 
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/binding"
+	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
+	"github.com/google/uuid"
 	"github.com/tektoncd/triggers/pkg/apis/triggers"
 	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1beta1"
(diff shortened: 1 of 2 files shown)


khrm (commented):
/lgtm

dibyom:
> One issue that I'm not sure about, is https://github.com/tektoncd/triggers/pull/1469/files#diff-040df62660d4581f9d1f6c27bebee01870d8b7b0cda5c21803369f5a55106e09R226
> Previously, it didn't return after sending the failed responses, which the other error cases did, there's no obvious reason why?

That looks like a bug but one that we'd probably never encounter since we were always returning valid json response back

dibyom:
Thank you @bigkevmcd - One thing I'd add is some documentation in to document this feature https://github.com/tektoncd/triggers/blob/main/docs/eventlisteners.md#understanding-eventlistener-response

bigkevmcd:
I've added an example response for CloudEvents, and added a Type and improved the `Source` field to include the namespace.

dibyom:
Hmm - the issue with the test seems to be that we are creating GKE clusters version 1.22 while trying to install the latest pipelines v0.41 which requires K8s 1.23

PR #1494, opened 2022-12-14
Add SecurityContext to Eventlistener containers

### LINKED ISSUE #1490: EventListeners do not have correct SecurityContext to run in tekton-pipelines namespace
@dibyom I apologize if this is the wrong place to provide feedback on this ticket, but I'm still getting this error on k8s v1.24, pipelines v0.42 and triggers v0.22 (I can confirm that `securityContext` fix is present in `tekton-triggers-controller` and `-webhook`):

```
Warning  FailedCreate  47s (x7 over 3m29s)  replicaset-controller  (combined from similar events): Error creating: pods "el-pulumi-el-8bc4bf7-nxznb" is forbidden: violates PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "event-listener" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "event-listener" must set securityContext.capabilities.drop=["ALL"]), seccompProfile (pod or container "event-listener" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

_Originally posted by @seunggs in https://github.com/tektoncd/triggers/issues/1477#issuecomment-1336448588_

dibyom:
# Changes

The security context is the same one that is applied to other Tekton workloads such as the Triggers and Pipeline controller pods. Eventlisteners already run as non-root, non-privileged containers. Adding this setting allows them to run in environments with pod security admission set to "restricted" (such as the tekton-pipelines namespace)

Fixes #1490

# Submitter Checklist

review them:

# Release Notes

```release-note
Eventlistener containers now contain the right security context to allow running with restricted pod security admission
```

/kind bug

### DIFF
--- a/pkg/reconciler/eventlistener/eventlistener_test.go
+++ b/pkg/reconciler/eventlistener/eventlistener_test.go
@@ -273,6 +273,19 @@ func makeDeployment(ops ...func(d *appsv1.Deployment)) *appsv1.Deployment {
 							Name:  "METRICS_PROMETHEUS_PORT",
 							Value: "9000",
 						}},
+						SecurityContext: &corev1.SecurityContext{
+							AllowPrivilegeEscalation: ptr.Bool(false),
+							Capabilities: &corev1.Capabilities{
+								Drop: []corev1.Capability{"ALL"},
+							},
+							// 65532 is the distroless nonroot user ID
+							RunAsUser:    ptr.Int64(65532),
+							RunAsGroup:   ptr.Int64(65532),
+							RunAsNonRoot: ptr.Bool(true),
+							SeccompProfile: &corev1.SeccompProfile{
+								Type: corev1.SeccompProfileTypeRuntimeDefault,
+							},
+						},
 					}},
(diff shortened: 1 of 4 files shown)


savitaashture:
shall we take input from EL spec
because with this change EL always runs as root and if someone wants to run EL as non-root then we don't have way specially cases like Openshift

khrm (commented):
Seems fine but need to check this once with OpenShift.

dibyom:
@savitaashture this changes ensures the container always runs as non-root. Do you have a use case for why someone might want it differently? If openshift or another platform does need it to be customized then yeah we could add it but I like the simplicity of always ensuring the EL container with the same security context

iancoffey:
Nice, LGTM

dibyom:
@savitaashture @khrm could you please take a look? 🙏

savitaashture (approved):
/lgtm

khrm (commented):
/kind bug

Issue #1487, opened 2022-11-30
Changed files in Github interceptor


kcloutie:
### Feature request

I would like the github interceptor to enrich the payload of an incoming request with the list of changed files related to a pull_request or push event similar to what is done in the [add-pr-body](https://github.com/tektoncd/plumbing/tree/main/tekton/ci/cluster-interceptors/add-pr-body) interceptor. In addition, it should also accept a regex expression that can be matched against the list of changed files to determine whether or not to halt processing.

Although the push event json payload contains the changed files, the file list is scattered. I would like the interceptor to normalize the list of changed files between the pull_request and push events so it is easier to extract and send the data down to the trigger template

### Use case

We have a pipeline that executes terraform. If no terraform code has been changed, we have no need for the pipeline to execute terraform. Today, the pipeline that executes terraform always runs, and one of the first steps is a custom task to determine if terraform files have changed.  If files have changed, then the pipeline continues; otherwise, it stops executing. Performing this action at the interceptor level fails fasts, simplifies the pipeline by removing a task and prevents unnecessary execution of the pipelineRun.

The same pipeline above will also need to know which directory (or directories) to execute terraform in. Today we need a task to get the changed files from git/github. We would like to remove that task and have the interceptor send that information down with the payload.

### Alternate Proposal (supporting the regex expression)

If the changed files are normalized and included in the payload, cel interceptor could be used to perform the regex match and halt processing as opposed to supporting that in the github interceptor

kcloutie:
I have created POC here https://github.com/kcloutie/cluster-interceptors

This cluster interceptor will add changed files to the github payload for a pull request and a push. The example uses the cel interceptor to verify the changed files match a regex pattern before continuing processing.

FYI - The `add-changed-files-enterprise` example will not be useable unless changes are made, it is used for internal testing at the moment.

khrm:
Implemented by #1507

PR #1507, opened 2023-01-10
add changed files to the github interceptor

kcloutie:
# Changes

Issue: #1487 

This change enriches the github payload of an incoming request with the list of changed files related to a pull_request or push event. The changed files list can be verified using a CEL interceptor to halt processing and/or can be passed down to the pipelineRun where the file list can be used by a task

# Submitter Checklist

review them:

# Release Notes

```release-note
The GitHub `Interceptor` has the ability to add a comma delimited list of all files that have changed (added, modified or deleted) for the `push` and `pull_request` events. The list of changed files are added to the `changed_files` property of the event payload in the top-level `extensions` field
```


khrm (commented):
/ok-to-test

examples/v1beta1/github-add-changed-files-push-cel/README.md
```diff
@@ -0,0 +1,53 @@
+## GitHub EventListener
+
+Creates an EventListener that listens for GitHub webhook events.
```
savitaashture, line 3: Could you change this line to give more info about add-changed file info

examples/v1beta1/github-add-changed-files-pr/README.md
```diff
@@ -0,0 +1,47 @@
+## GitHub EventListener
+
+Creates an EventListener that listens for GitHub webhook events.
```
savitaashture, line 3: Could you change this line to give more info about add-changed file info

pkg/interceptors/github/github.go
```diff
@@ -96,7 +118,209 @@ func (w *Interceptor) Process(ctx context.Context, r *triggersv1.InterceptorRequ
...
+	}
+
 	return &triggersv1.InterceptorResponse{
 		Continue: true,
 	}
 }
+
+func (w *Interceptor) getGithubTokenSecret(ctx context.Context, r *triggersv1.InterceptorRequest, p triggersv1.GitHubInterceptor) (string, error) {
+	if p.AddChangedFiles.PersonalAccessToken == nil {
+		return "", nil
```
savitaashture, line 175: If no PersonalAccessToken passed then we should return error :thinking: asking because if no PersonalAccessToken then further we cannot do anything right 
Or am i missing something here?

pkg/interceptors/github/github.go
```diff
@@ -96,7 +118,209 @@ func (w *Interceptor) Process(ctx context.Context, r *triggersv1.InterceptorRequ
...
+	if err != nil {
+		return "", err
+	}
+	return string(secretToken), nil
+}
+
+func parseBodyForChangedFiles(body string, eventType string) (payloadDetails, error) {
+	results := payloadDetails{}
+	if body == "" {
+		return results, fmt.Errorf("body was empty")
```
savitaashture, line 191:
```suggestion
		return results, fmt.Errorf("body is empty")
```

pkg/interceptors/github/github.go
kcloutie: If the repository does not require authentication, the PersonalAccessToken is not required to make the github API call. This can be seen in the `github-add-changed-files-push-cel` example which makes the github API call without the PAT.

examples/v1beta1/github-add-changed-files-push-cel/README.md
kcloutie: Updated

examples/v1beta1/github-add-changed-files-pr/README.md
kcloutie: Updated

pkg/interceptors/github/github.go
kcloutie: Updated

pkg/interceptors/github/github.go
```diff
@@ -96,7 +118,209 @@ func (w *Interceptor) Process(ctx context.Context, r *triggersv1.InterceptorRequ
...
+			if !ok {
+				return results, fmt.Errorf("payload body missing 'commits.*.modified' field")
+			}
+
+			removedFiles, ok := commit.(map[string]interface{})["removed"].([]interface{})
+			if !ok {
+				return results, fmt.Errorf("payload body missing 'commits.*.removed' field")
+			}
+			for _, fileName := range addedFiles {
+				changedFiles = append(changedFiles, fmt.Sprintf("%v", fileName))
```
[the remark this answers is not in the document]
savitaashture: I see ok

docs/interceptors.md
```diff
@@ -187,6 +187,43 @@ For reference, below is an example legacy GitHub `Interceptor` definition:
 
 For more information, see our [example](../examples/v1beta1/github) of using this `Interceptor`.
 
+#### Adding Changed Files
+
+The GitHub `Interceptor` also has the ability to add a comma delimited list of all files that have changed (added, modified or deleted) for the `push` and `pull_request` events. The list of changed files are added to the `changed_files` property of the event payload in the top-level `extensions` field
```
dibyom, line 192: Thanks for adding the docs...it seems like add changed files also supports private github repositories using a personal access token - could we document the fields that are needed to support that here as well?

pkg/apis/triggers/v1beta1/trigger_types.go
```diff
@@ -166,7 +166,13 @@ type BitbucketInterceptor struct {
 type GitHubInterceptor struct {
 	SecretRef *SecretRef `json:"secretRef,omitempty"`
 	// +listType=atomic
-	EventTypes []string `json:"eventTypes,omitempty"`
+	EventTypes      []string              `json:"eventTypes,omitempty"`
+	AddChangedFiles GithubAddChangedFiles `json:"addChangedFiles,omitempty"`
```
dibyom, line 170: Hmm...we used to have these structs in the trigger_types package since we used to support specifying `secretRef `and `eventTypes` as fields vs the param/value based approach that we do now. I wonder if it is possible to move this struct into the specific interceptor implementation itself. I think doing so would mean that newly added params to an interceptor don't show up in the `triggers-api.md` file

savitaashture: @dibyom do you mean moving `AddChangedFiles` to specific Interceptor i.e Github Interceptor  and keeping `secretRef` and `eventTypes` as it is in trigger_types.go in order to make sure if there are anything specific to Interceptor will be added to that respective Interceptor ?

docs/interceptors.md
kcloutie: @dibyom Added additional documentation on personal access token

pkg/apis/triggers/v1beta1/trigger_types.go
dibyom: yes, that is the intention. If that is too complicated to do in this PR, we could open a clean issue to track this

savitaashture: +1 for doing it as another PR 
WDYT @kcloutie

kcloutie: @dibyom / @savitaashture  I looked at the change and it doesn't appear to be very easy. Moving only `AddChangedFiles` struct will cause a circular reference and would require, at the very least, to move `GitHubInterceptor ` struct.. I vote for a new issue to do that change. Let me know if you agree

savitaashture: Yes create a new issue and that you can handle in new PR

savitaashture (approved):
:tada:

khrm (commented):
/kind feature

Issue #1463, opened 2022-10-13
tls: client offered only unsupported versions


theautoroboto:
# Expected Behavior
tekton-triggers-core-interceptors v0.21.0 pod to run as expected on OpenShift v4.10.30

The deployment process has been tested on v0.20.0, v0.20.1, and v0.20.2 without issue

# Actual Behavior
tekton-triggers-core-interceptors v0.21.0 pod fails with the error
```
http: TLS handshake error from 10.128.2.1:36306: tls: client offered only unsupported versions: [303]
```

# Steps to Reproduce the Problem

1. Deploy fresh or update to v0.21.0/interceptors.yaml
```
kubectl apply -f https://storage.googleapis.com/tekton-releases/triggers/previous/v0.21.0/interceptors.notags.yaml
or
kubectl apply -f https://storage.googleapis.com/tekton-releases/triggers/previous/v0.21.0/interceptors.yaml
```

# Additional Info

- Kubernetes version:
```
Client Version: 4.11.2
Kustomize Version: v4.5.4
Server Version: 4.10.30
Kubernetes Version: v1.23.5+012e945
```

- Tekton Pipeline version:
```
v0.40.2
```

- pods logs
```
{"level":"info","ts":1665673664.7329836,"logger":"fallback","caller":"injection/injection.go:61","msg":"Starting informers..."}
{"level":"info","ts":1665673668.7711773,"caller":"interceptors/main.go:178","msg":"Listen and serve on port 8443"}
2022/10/13 15:07:52 http: TLS handshake error from 10.128.2.1:54862: tls: client offered only unsupported versions: [303]
2022/10/13 15:08:02 http: TLS handshake error from 10.128.2.1:49866: tls: client offered only unsupported versions: [303]
2022/10/13 15:08:12 http: TLS handshake error from 10.128.2.1:33576: tls: client offered only unsupported versions: [303]
2022/10/13 15:08:22 http: TLS handshake error from 10.128.2.1:53678: tls: client offered only unsupported versions: [303]
```

savitaashture:
Hi @theautoroboto 

I tried installing Pipeline and Triggers on Openshift v4.10.30

Openshift Version
```
$ oc version
Client Version: 4.6.8
Server Version: 4.10.30
Kubernetes Version: v1.23.5+012e945
``` 

Tekton Component Version
```
$ tkn version
Client version: 0.23.1
Pipeline version: v0.40.2
Triggers version: v0.21.0
```

[...]

PR #1518, opened 2023-02-01
Fix tls: client offered only unsupported versions

savitaashture:
# Changes
* Fixes https://github.com/tektoncd/triggers/issues/1463 this issue occurs on Openshift cluster where **fips** are enabled
* Moved config/interceptors/interceptor-secrets.yaml to config/interceptors/00-interceptor-secrets.yaml so that secrets will be created before core-intercetor pod start this solves issue like `tekton-triggers-core-interceptors-certs` not found

# Submitter Checklist

review them:

# Release Notes


savitaashture:
@dibyom @khrm we may need v0.22.2 patch release for this fix as its blocking to use Triggers on a cluster where fips are enabled

khrm (commented):
/approve

khrm:
@savitaashture Removing min version in the tls, is flagged as a security issue by gosec. Does it really resolve FIPS?

dibyom:
> @savitaashture Removing min version in the tls, is flagged as a security issue by gosec. Does it really resolve FIPS?

+1. Don't know a lot about FIPS but from https://developers.redhat.com/articles/2022/05/31/your-go-application-fips-compliant it sounds like you have to build with a different set of crypto libraries for it to be FIPS compatible?

savitaashture:
> > @savitaashture Removing min version in the tls, is flagged as a security issue by gosec. Does it really resolve FIPS?
> 
> +1. Don't know a lot about FIPS but from https://developers.redhat.com/articles/2022/05/31/your-go-application-fips-compliant it sounds like you have to build with a different set of crypto libraries for it to be FIPS compatible?

Yes agree

Looked deeper and discussed with those who know about FIPS

Reason:

1. During peer-peer https communication, when TLS handshake is happening, at a stage where both server and client mutually agree for a common TLS version and cipher,, The Triggers Interceptor server is enabled or only allow TLS 1.3 and corresponding ciphers today, where the client(OCP) giving a TLS 1.2 tls version and cipher(Based on error `http: TLS handshake error from 10.128.2.1:36306: tls: client offered only unsupported versions: [303]` [Ref](https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L31),) 
which result in the TLS handshake not happening as there is no mutual TLS version to negotiate from both (server and client ) side

3. OCP uses MInTLS as 1.2 for all components


So now changed `MinVersion` to [tls.VersionTLS12](https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L31) and it works on both K8S and OCP

savitaashture:
> @dibyom @khrm we may need v0.22.2 patch release for this fix as its blocking to use Triggers on a cluster where fips are enabled

We may need to do v0.21.x because this issue exist in v0.21.x as well

dibyom:
Thanks @savitaashture - the explanation is very helpful. Can we update the PR description/commit message with this?
/lgtm
/hold 
(feel free to cancel the hold when you have updated!)

savitaashture:
> Thanks @savitaashture - the explanation is very helpful. Can we update the PR description/commit message with this? /lgtm /hold (feel free to cancel the hold when you have updated!)

Thank you @dibyom 

Updated commit message and Description

khrm (commented):
/kind bug

PR #1521, opened 2023-02-07
Change TLS MinVersion to tls.VersionTLS12 in order to make Triggers run on OCP where FIPS enabled

savitaashture:
# Changes
* **Fixes**: https://github.com/tektoncd/triggers/issues/1463 this issue occurs on Openshift cluster where **fips** are enabled

  **Issue**: `http: TLS handshake error from 10.128.2.1:36306: tls: client offered only unsupported versions: [303]`
  **RCA:**

  1. During peer-peer https communication, when TLS handshake is happening, at a stage where both server and client mutually agree for a common TLS version and cipher,, The Triggers Interceptor server is enabled or only allow TLS 1.3 and corresponding ciphers today, where the client(OCP) giving a TLS 1.2 tls version and cipher(Based on error http: TLS handshake error from 10.128.2.1:36306: tls: client offered only unsupported versions: [303] [Ref](https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L31),)
which result in the TLS handshake not happening as there is no mutual TLS version to negotiate from both (server and client ) side

   2. OCP uses MInTLS as 1.2 for all components

   **Fix:**
   1. Changed MinVersion to [tls.VersionTLS12](https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L31) and it works on both K8S and OCP

* Moved config/interceptors/interceptor-secrets.yaml to config/interceptors/00-interceptor-secrets.yaml so that secrets will be created before core-intercetor pod start this solves issue like `tekton-triggers-core-interceptors-certs` not found

# Submitter Checklist

review them:

# Release Notes

```release-note
Changed TLS MinVersion to `tls.VersionTLS12` in order to make Triggers run on OCP(Where FIPS enabled) as OCP uses MInTLS as 1.2 for all components
```


khrm (commented):
/lgtm

khrm (approved):
/lgtm

Issue #1512, opened 2023-01-18
affinity and podTopologySpread are both rejected by validation webook


treyhyde:
# Expected Behavior

As per https://pkg.go.dev/knative.dev/pkg/apis/duck/v1#PodSpecable, allowed to use keys from https://pkg.go.dev/k8s.io/api/core/v1#PodTemplateSpec in resources.kubernetesResources.spec.templet.spec

It's important to be able to distribute our EL service over multiple AZs and have some sort of availability guarantee, neither affinity or podTopologySpread which allow these guarantees are  permitted.

# Actual Behavior

  cannot patch "fulcrum" with kind EventListener: admission webhook "validation.webhook.triggers.tekton.dev" denied the request: validation failed: must not set the field(s): spec.resources.kubernetesResource.spec.template.spec.affinity, spec.resources.kubernetesResource.spec.template.spec.topologySpreadConstraints
  Error: plugin "secrets" exited with error

# Steps to Reproduce the Problem

add a EL with

```yaml
spec:
  resources:
    kubernetesResource:
      replicas: 3
      servicePort: 80
      spec:
        template:
          spec:
            topologySpreadConstraints:
            - maxSkew: 1
              topologyKey: topology.kubernetes.io/zone
              whenUnsatisfiable: DoNotSchedule
              labelSelector:
                matchLabels:
                  eventlistener: my-el-name
```

# Additional Info

- Kubernetes version:
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.4", GitCommit:"872a965c6c6526caa949f0c6ac028ef7aff3fb78", GitTreeState:"clean", BuildDate:"2022-11-09T13:36:36Z", GoVersion:"go1.19.3", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"24+", GitVersion:"v1.24.7-eks-fb459a0", GitCommit:"c240013134c03a740781ffa1436ba2688b50b494", GitTreeState:"clean", BuildDate:"2022-10-24T20:36:26Z", GoVersion:"go1.18.7", Compiler:"gc", Platform:"linux/amd64"}

- Tekton Pipeline version:
 tkn version
Client version: 0.29.0
Pipeline version: v0.40.2
Triggers version: v0.20.2
Dashboard version: v0.29.2

treyhyde:
[...]

PR #1540, opened 2023-02-27
Add support for Affinity and TopologySpreadConstraints

savitaashture:
# Changes
Addressed : https://github.com/tektoncd/triggers/issues/1512

Triggers now support Affinity and TopologySpreadConstraints for podSpec

# Submitter Checklist

review them:

# Release Notes

```release-note
Triggers now support Affinity and TopologySpreadConstraints as part of Kubernetes and Custom resource 
```

### DIFF
--- a/docs/eventlisteners.md
+++ b/docs/eventlisteners.md
@@ -270,14 +270,14 @@ Legal values for the `PodSpec` and `Containers` sub-fields for both `kubernetesR
 ServiceAccountName
 NodeSelector
 Tolerations
-Volumes
 Containers
+Affinity
+TopologySpreadConstraints
 ```
 
 Legal values for the `Containers` sub-field are:
 ```
 Resources
-VolumeMounts
 Env
 ```
 
--- a/pkg/apis/triggers/v1alpha1/event_listener_validation.go
+++ b/pkg/apis/triggers/v1alpha1/event_listener_validation.go
@@ -257,7 +257,6 @@ func podSpecMask(in *corev1.PodSpec) *corev1.PodSpec {
 	out.SecurityContext = nil
 	out.Hostname = ""
 	out.Subdomain = ""
-	out.Affinity = nil
 	out.SchedulerName = ""
 	out.HostAliases = nil
 	out.PriorityClassName = ""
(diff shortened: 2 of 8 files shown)


khrm (commented):
/kind feature

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
```diff
@@ -532,6 +532,45 @@ func Test_EventListenerValidate(t *testing.T) {
 				}},
 			},
 		},
+	}, {
+		name: "Valid EventListener with node affinity",
+		el: &v1alpha1.EventListener{
+			ObjectMeta: metav1.ObjectMeta{
```
dibyom, line 538: do we need to add this in both v1alpha1 and v1beta1?

dibyom:
@savitaashture let's add the new field to the docs as well!

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
savitaashture: Yes you are right we need to add for both because this is not something new for beta

savitaashture: Also we don't specify supported for alpha and beta here https://github.com/tektoncd/triggers/blob/main/docs/eventlisteners.md#specifying-resources that why added in both in order to not to confuse

savitaashture:
@dibyom @khrm PR is ready for review

pkg/apis/triggers/v1alpha1/event_listener_validation_test.go
dibyom: > Do you think new supported fields can be added only for beta not for alpha 🤔

Yeah, that's how pipelines did it. They essentially froze v1alpha1 and only added new features to v1beta1 for simplicity.

dibyom:
Thanks @savitaashture - question around whether we need to add this to v1alpha1 as well as v1beta1 - I think v1beta1 alone is enough unless there is some explicit need to add it to v1alpha1 as well. Otherwise, looks good.

/approve

savitaashture:
> Thanks @savitaashture - question around whether we need to add this to v1alpha1 as well as v1beta1 - I think v1beta1 alone is enough unless there is some explicit need to add it to v1alpha1 as well. Otherwise, looks good.

Agree 

Ya we can add to v1alpha1 based on need
Will remove that and update :+1:

savitaashture:
@khrm @dibyom addressed review comments and up for final review 
PTAL and merge if it looks good

khrm (commented):
/lgtm

PR #1551, opened 2023-03-06
Fix marshalJSON for lists of objects

### LINKED ISSUE #1538: The marshalJSON function doesn't marshal lists
# Expected Behavior
The expected behavior is that the pipeline starts and the correct JSON information is geven as parameter to the Pipeline.
Knative sends JSON data to the EventListener pod and uses the marshalJSON function to make a string of it for the next Task.

# Actual Behavior
The actual behavior is that the Pipeline doesn't even start at all. The EventListener pod gives the following error.
```
rpc error: code = InvalidArgument desc = error evaluating cel expression: expression \"body.spec.tags.marshalJSON()\" failed to evaluate: no such overload: marshalJSON(list)"
```
But the JSON code inside the `body.spec.tags.marshalJSON()` is valid JSON (checked separately).
This means there is a issue inside the marshalJSON function that can't handle a given list, while the documentation says it can handle a list.

1. Send JSON payload (from Knative in this case) to the EventListener pod
2. use 'marshalJSON()' inside TriggerTemplate to process this JSON payload
3. View logs inside the EventListener pod

# Additional Info

- Kubernetes version:

  ```
  OpenShift 4.12.0
  Kubernetes Version v1.25.4+77bec7a
  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

  ```
  Client version: 0.28.0
  Pipeline version: v0.41.0
  Triggers version: v0.22.0
  Operator version: v0.63.0
  ```

Right now I made a work-around that creates a list, even while there is already a list given, so the marshalJSON function can handle this.
The current cel expression in my EventListener yaml looks like this:
```
params:
- name: overlays
  value:
  - key: imagestream-tags
    expression: "{\"tags-list\": body.spec.tags}.marshalJSON()"
```

bigkevmcd:
# Changes
Fix marshalJSON for lists of objects

This fixes a bug in the CEL interceptor where the binding to marshalJSON was only for lists of strings.

The fix is to open up the binding to lists of any JSON objects.

Fixes #1538 

# Submitter Checklist

review them:

- [X] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#tests) (if functionality changed/added)
- [X] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commits)

# Release Notes
```release-note
Fix a bug in CEL interceptor's `marshalJSON` binding to allow marshaling of maps.
```

### DIFF
--- a/pkg/interceptors/cel/cel_test.go
+++ b/pkg/interceptors/cel/cel_test.go
@@ -433,6 +433,13 @@ func TestExpressionEvaluation(t *testing.T) {
 		"jsonArray": []string{
 			"one", "two",
 		},
+		"jsonObjects": []map[string]interface{}{
+			{
+				"testing1": map[string]interface{}{
+					"testing": []string{"test1", "test2"},
+				},
+			},
+		},
 	}
 
 	refParts := strings.Split(testRef, "/")
@@ -588,6 +595,11 @@ func TestExpressionEvaluation(t *testing.T) {
 			name: "marshal JSON array to string",
 			expr: "body.jsonArray.marshalJSON()",
 			want: types.String(`["one","two"]`),
+		},
+		{
+			name: "marshal JSON objects to string",
+			expr: "body.jsonObjects.marshalJSON()",
+			want: types.String(`[{"testing1":{"testing":["test1","test2"]}}]`),
 		},
 		{
 			name: "extension base64 decoding",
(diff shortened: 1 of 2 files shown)


khrm (commented):
/lgtm

savitaashture (approved):
Thank you @bigkevmcd

Issue #1513, opened 2023-01-24
Move interceptor structs into their specific interceptor implementations


kcloutie:
Move all the interceptor related structs into their specific implementation. Doing so would mean that newly added params to an interceptor will not show up in the `triggers-api.md` file  Please see [this conversation](https://github.com/tektoncd/triggers/pull/1507#discussion_r1082970488) for more details

khrm:
Fixed by #1588

PR #1553, opened 2023-03-06
Move interceptor types to implementation

kcloutie:
# Changes

Moved types used by interceptors from `trigger_types.go` to the interceptors specific implementation as requested in PR #1507 and issues #1513


khrm (commented):
/ok-to-test

khrm (commented):
/kind cleanup

pkg/interceptors/bitbucket/bitbucket.go
```diff
@@ -26,20 +26,27 @@ import (
 	"google.golang.org/grpc/codes"
 )
 
-var _ triggersv1.InterceptorInterface = (*Interceptor)(nil)
+var _ triggersv1.InterceptorInterface = (*BaseInterceptor)(nil)
 
-type Interceptor struct {
+type BaseInterceptor struct {
```
dibyom, line 31: So, the BaseInterceptor is the struct that implements the InterceptorInterface. Maybe we could just call it InterceptorImpl?

pkg/interceptors/bitbucket/bitbucket.go
```diff
@@ -26,20 +26,27 @@ import (
...
-	return &Interceptor{
+func NewInterceptor(sg interceptors.SecretGetter) *BaseInterceptor {
+	return &BaseInterceptor{
 		SecretGetter: sg,
 	}
 }
 
-func (w *Interceptor) Process(ctx context.Context, r *triggersv1.InterceptorRequest) *triggersv1.InterceptorResponse {
-	p := triggersv1.BitbucketInterceptor{}
+// Interceptor provides a webhook to intercept and pre-process events
```
dibyom, line 41: nit:  remove this comment since it doesn't describe the struct below

pkg/interceptors/bitbucket/bitbucket.go
```diff
@@ -26,20 +26,27 @@ import (
...
+type Interceptor struct {
```
dibyom, line 42: This Interceptor struct defines the structure of the `[InterceptorParams](https://github.com/tektoncd/triggers/blob/d54faeb18914b42f0f8a02ae69a3287879194e5c/pkg/apis/triggers/v1beta1/interceptor_types.go#L36)` for this particular InterceptorImplementation....does this struct have to be called Interceptor? Maybe `InterceptorParams` or just `BitBucketParams` since this is specific to this interceptor?

dibyom (commented):
Thank you for the refactoring @kcloutie just a couple of minor comments on making the struct names a bit more descriptive but otherwise looks great!

pkg/interceptors/bitbucket/bitbucket.go
dibyom: sure - this is subjective, I found it a bit confusing what the Interceptor struct meant here...really all it is is the set of parameters that this interceptor accepts. So, to me InterceptorParams or (event just `Params`) makes it a bit more descriptive

dibyom: I don't think feel too strongly about this, so feel free to keep the `Interceptor` name though in that case, please add a comment describing what the struct contents.

kcloutie: @dibyom made the required changes

dibyom (approved):
Thanks so much for doing the refactoring @kcloutie !

Issue #1599, opened 2023-05-30
CE response from sink fails for the Eventing MT ingress broker


khrm:
# Expected Behavior
CE response type should work with https://github.com/knative/eventing/blob/main/docs/broker/README.md

# Actual Behavior
Because type is empty, we get:

```
{"level":"warn","ts":"2023-05-29T21:39:29.919Z","logger":"mt_broker_ingress","caller":"ingress/ingress_handler.go:146","msg":"failed to validate extracted event","commit":"034bec9","error":"type: MUST be a non-empty string\n"}
```

khrm:
@afrittoli Created #1600 to fix this

PR #1600, opened 2023-05-30
Add SetType and SetSubject to CE sink response

khrm:
# Changes

# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -31,7 +31,6 @@ import (
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"github.com/cloudevents/sdk-go/v2/binding"
 	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
-	"github.com/google/uuid"
 	"github.com/tektoncd/triggers/pkg/apis/triggers"
 	triggersv1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1beta1"
 	triggersclientset "github.com/tektoncd/triggers/pkg/client/clientset/versioned"
@@ -235,8 +234,10 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 
 	} else {
 		responseEvent := cloudevents.NewEvent()
-		responseEvent.SetID(uuid.New().String())
-		responseEvent.SetSource(r.EventListenerName)
+		responseEvent.SetID(eventID)
+		responseEvent.SetType(events.EventAccepted)
+		responseEvent.SetSubject(r.EventListenerNamespace + "." + r.EventListenerName + " accepted " + eventID)
+		responseEvent.SetSource(r.EventListenerName) // We need to change this like in SendCloudEvents
 
 		_ = responseEvent.SetData(cloudevents.ApplicationJSON, body)
 
--- a/pkg/reconciler/events/event.go
+++ b/pkg/reconciler/events/event.go
@@ -31,6 +31,8 @@ const (
 	// TriggerProcessingDoneV1 is sent for Sink Triggers when we are done
 	// with eventlistener handler
 	TriggerProcessingDoneV1 = "dev.tekton.event.triggers.done.v1"
+	// EventAccepted is sent as response for CloudEvent compliant providers
+	EventAccepted = "dev.tekton.event.triggers.accepted.v1"
 )
 
 // Emit emits events for object


khrm:
Fixes #1599

pkg/sink/sink.go
```diff
@@ -235,10 +234,12 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 
 	} else {
 		responseEvent := cloudevents.NewEvent()
-		responseEvent.SetID(uuid.New().String())
-		responseEvent.SetSource(r.EventListenerName)
+		responseEvent.SetID(elUID)
```
matejvasek, line 237: I think this should be `eventID`.

pkg/sink/sink.go
```diff
@@ -235,10 +234,12 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
...
+		responseEvent.SetSubject(r.EventListenerNamespace + "." + r.EventListenerName + " processing " + elUID)
+		responseEvent.SetSource(r.EventListenerName + "." + r.EventListenerNamespace)
```
matejvasek, line 239: This looks all right. Alternatively it could be `elUID`.

pkg/sink/sink.go
```diff
@@ -235,10 +234,12 @@ func (r Sink) HandleEvent(response http.ResponseWriter, request *http.Request) {
 
 	} else {
 		responseEvent := cloudevents.NewEvent()
-		responseEvent.SetID(uuid.New().String())
-		responseEvent.SetSource(r.EventListenerName)
+		responseEvent.SetID(elUID)
+		responseEvent.SetSubject(r.EventListenerNamespace + "." + r.EventListenerName + " processing " + elUID)
```
matejvasek, line 238: IIRC subject is not mandatory.
Required are:
spec-version (set by "constructor" automatically)
ce-type
ce-source
ce-id

matejvasek: ce-subject and ce-time are optional, but it might be nice to have them

matejvasek:
It looks like the `events.TriggerProcessingDoneV1` event is also produced asynchronously if `CloudEventURI` is set.

See: https://github.com/tektoncd/triggers/blob/76f51dafe6453a37e6c6de35859692306d174666/pkg/sink/cloudevent/cloudevent.go#L49-L60

The event produced asynchronously and via HTTP response should be identical (at least id and source should be the same).

pkg/sink/sink.go
matejvasek: Actually the source is little bit more complicated:
See: https://github.com/tektoncd/triggers/pull/1600#issuecomment-1569007791

pkg/sink/sink.go
khrm: Done.

pkg/sink/sink.go
```diff
@@ -56,6 +55,11 @@ import (
 	v1 "knative.dev/pkg/apis/duck/v1"
 )
 
+const (
+	// EventProcessed is sent for CloudEvent compliant providers
+	EventProcessed = "dev.tekton.event.triggers.processed.v1"
```
savitaashture, line 60: Shall we document this

khrm: We didn't document CloudEvent Response till now. Probably, need to document not only this but CloudEvent Response behaviour.

savitaashture: Sure 
Can you create a issue so that it will be tracked

savitaashture: Sure 
Can you create a issue so that it will be tracked

khrm: Done here: https://github.com/tektoncd/triggers/issues/1601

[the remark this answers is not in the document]
dibyom: Also, EventAccepted might be a better name since when the response is sent, the event may not have been fully processed yet as the triggers may still be processing them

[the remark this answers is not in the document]
khrm: Done. I didn't include it pkg/reconciler earlier because those are k8s/cloudEvent used during processing. But I think it doesn't matter, it can cater to any events we are using in Triggers.

savitaashture (approved):
:tada:

savitaashture (commented):
/lgtm

PR #1650, opened 2023-09-24
Follow naming conventions for resourceTemplates

savitaashture:
As per the standards the JSON or YAML keys should follow capitalization but resourcetemplates keys doesn't
follow that.
Changing resourcetemplates to resourceTemplates

Signed-off-by: Savita Ashture <sashture@redhat.com>"

# Changes

# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/triggerrun/testdata/triggertemplate.yaml
+++ b/cmd/triggerrun/testdata/triggertemplate.yaml
@@ -13,7 +13,7 @@ spec:
   - name: message
     description: The message to print
     default: This is the default message
-  resourcetemplates:
+  resourceTemplates:
   - apiVersion: tekton.dev/v1beta1
     kind: PipelineRun
     metadata:
--- a/docs/getting-started/triggers.yaml
+++ b/docs/getting-started/triggers.yaml
@@ -12,7 +12,7 @@ spec:
       description: The git repository url
     - name: namespace
       description: The namespace to create the resources
-  resourcetemplates:
+  resourceTemplates:
     - apiVersion: tekton.dev/v1beta1
       kind: PipelineRun
       metadata:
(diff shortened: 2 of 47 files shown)


khrm (commented):
What would happen if some resource already exist with that version? Don't we need conversion function?

savitaashture:
> What would happen if some resource already exist with that version? Don't we need conversion function?

Its not needed because it works for both `resourceTemplates` and `resourcetemplates` i have verified

khrm:
> > What would happen if some resource already exists with that version? Don't we need a conversion function?
> 
> Its not needed because it works for both `resourceTemplates` and `resourcetemplates` i have verified

It's working because go default JSON decoder/encoder doesn't care about case or characters like `_`. So etcd would have two different field names after this pr - older with `resourcetemplates` and newly created with `resourceTemplates`.

savitaashture:
> > > What would happen if some resource already exists with that version? Don't we need a conversion function?
> > 
> > 
> > Its not needed because it works for both `resourceTemplates` and `resourcetemplates` i have verified
> 
> It's working because go default JSON decoder/encoder doesn't care about case or characters like `_`. So etcd would have two different field names after this pr - older with `resourcetemplates` and newly created with `resourceTemplates`.

Yes but I think that should be fine right

dibyom:
This seems like the right thing to do - as long as this is not a breaking change i.e. the old style works too

savitaashture:
Hi @dibyom @khrm if PR looks good can you guys review and merge it 
Thank you

khrm (commented):
/approve

dibyom (commented):
/lgtm

PR #1736, opened 2024-05-20
Update securityContext to include required configurations when the flag el-security-context is enabled

savitaashture:
As part of this PR, the securityContext is being updated to set RunAsUser and RunAsGroup when the el-security-context flag is enabled.
This change handles cases in environments where user ID 65532 is not allowed, such as OpenShift.

Signed-off-by: Savita Ashture <sashture@redhat.com>

# Changes

# Submitter Checklist

# Release Notes

```release-note
NONE
```


khrm (commented):
/kind feature

pkg/reconciler/eventlistener/resources/container.go
```diff
@@ -49,21 +49,19 @@ func MakeContainer(el *v1beta1.EventListener, configAcc reconcilersource.ConfigA
...
+		AllowPrivilegeEscalation: ptr.Bool(false),
+		Capabilities: &corev1.Capabilities{
+			Drop: []corev1.Capability{"ALL"},
+		},
+		RunAsNonRoot: ptr.Bool(true),
+		SeccompProfile: &corev1.SeccompProfile{
+			Type: corev1.SeccompProfileTypeRuntimeDefault,
+		},
+	}
 	if *c.SetSecurityContext {
```
khrm, line 62: We are already setting context now irrespective of flag. I believe we need another flag to enable/disable user in security context.

savitaashture: Ah i see what you mean 

then we may need to remove the existing flag `el-security-context`  and add new one `el-set-user-security-context` ??

khrm: No need to remove the older one. Just add a newer one for userid.

savitaashture: then what is the use of `el-security-context` ?

savitaashture: After thinking about upgrade and allel-security-context

pkg/reconciler/eventlistener/resources/container.go
```diff
@@ -56,16 +56,18 @@ func MakeContainer(el *v1beta1.EventListener, configAcc reconcilersource.ConfigA
...
-			RunAsUser:    ptr.Int64(65532),
-			RunAsGroup:   ptr.Int64(65532),
 			RunAsNonRoot: ptr.Bool(true),
 			SeccompProfile: &corev1.SeccompProfile{
 				Type: corev1.SeccompProfileTypeRuntimeDefault,
 			},
 		}
 	}
 
+	if *c.SetSecurityContextIDs {
```
savitaashture, line 66: @dibyom @khrm

I have few query
* should we actually maintain 2 flags `el-security-context` and  `el-security-context-run-as-ids`
* Is there any usecase where someone don't want to set securityContext for EventListener??
* If its okay to set securityContext by default on EL then planning to use just one flag `el-security-context-run-as-ids` to set runAsUser and runAsGroup as these are not needed for Openshift

Do let me know your thoughts

**Note**: if i change flag from `el-security-context` **--->**  `el-security-context-run-as-ids the upgrade is success

dibyom: For Openshift you  want to set the all fields of the default security context except for the RunAsUser and RunAsGroup? Maybe instead of adding another flag, we add read the default values of RunAsUser and RunAsGroup from [`config-defaults`](https://github.com/tektoncd/triggers/blob/main/config/config-defaults-triggers.yaml#L42) or the config-features config maps? We can set the default to the current value and for Openshift, you can override it to another value or set it to empty?

savitaashture: Yes make sense
So that way user can configure whatever value they want for runAsUser and runAsGroup :+1: 

Will change and update the PR 
Thank you

savitaashture: Updated PR
PTAL 
Thank you

dibyom:
One nit and looks like we need to fix the yaml files. Otherwise, LGTM

```
pkg/apis/config/testdata/config-defaults-triggers-empty-val.yaml
  24:1      error    too many blank lines (1 > 0)  (empty-lines)
pkg/apis/config/testdata/config-defaults-triggers.yaml
  24:1      error    too many blank lines (1 > 0)  (empty-lines)
```

/approve

savitaashture:
@khrm PTAL 
Thank you

pkg/reconciler/eventlistener/resources/container.go
khrm: This is a better approach. :)

khrm (commented):
/lgtm

PR #1747, opened 2024-07-05
Handle validation when value for runAsGroup and runAsUser is empty

savitaashture:
# Changes

As part of https://github.com/tektoncd/triggers/pull/1747/commits/48f257d9cc8539c9e05ad51536f403d43c43c96e and https://github.com/tektoncd/triggers/pull/1747/commits/add787c2d667ae8222d73d6a3257b7c703a49fed

* when `default-run-as-user` and `default-run-as-group` is **empty ("")** then don't set container SCC for runAsUser and runAsGroup 
because users on platform like OpenShift generally don't set these values and it will be set with default
So added check to handle the same

As part of https://github.com/tektoncd/triggers/pull/1747/commits/848ccb06b94069ef7eb181737ebc67568c1f8ceb

* Added new field **default-run-as-non-root** to `config-defaults-triggers` configmap so that RunAsNonRoot can be now configured through CM

* Reason for configuring runAsNonRoot through cm 
   *  when runAsUser and runAsGroup set to `0` which is valid value but because of `runAsNonRoot: true` 
   getting below error
   ```
   waiting:
        message: 'container''s runAsUser breaks non-root policy (pod: "el-github-listener-5878566fb4-hqt5r_default(086e7204-f323-499c-90a2-4b7e0e277e2d)",
          container: event-listener)'
        reason: CreateContainerConfigError
   ```
  * Allowing runAsNonRoot to be configured through CM will ease users to set runAsUser and runAsGroup to `0`
basically this can be the combination
   ```
     securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      runAsGroup: 0
      runAsNonRoot: false
      runAsUser: 0
      seccompProfile:
        type: RuntimeDefault
   ```

# Submitter Checklist

# Release Notes

```release-note
* Added new field default-run-as-non-root to config-defaults-triggers configmap so that RunAsNonRoot can be now configured through CM
```

### DIFF
--- a/config/config-defaults-triggers.yaml
+++ b/config/config-defaults-triggers.yaml
@@ -42,3 +42,4 @@ data:
     default-service-account: "default"
     default-run-as-user: "65532"
     default-run-as-group: "65532"
+    default-run-as-non-root: "true" # allowed values are true and false
--- a/pkg/apis/config/default.go
+++ b/pkg/apis/config/default.go
@@ -17,7 +17,9 @@ limitations under the License.
 package config
 
 import (
+	"fmt"
 	"os"
+	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 )
@@ -26,9 +28,11 @@ const (
 	defaultServiceAccountKey   = "default-service-account"
 	defaultRunAsUserKey        = "default-run-as-user"
 	defaultRunAsGroupKey       = "default-run-as-group"
+	defaultRunAsNonRootKey     = "default-run-as-non-root"
 	DefaultServiceAccountValue = "default"
 	defaultRunAsUserValue      = "65532"
 	defaultRunAsGroupValue     = "65532"
+	defaultRunAsNonRootValue   = true
 )
 
 // Defaults holds the default configurations
@@ -37,6 +41,7 @@ type Defaults struct {
 	DefaultServiceAccount string
 	DefaultRunAsUser      string
 	DefaultRunAsGroup     string
+	DefaultRunAsNonRoot   bool
 }
 
 // GetDefaultsConfigName returns the name of the configmap containing all
@@ -60,7 +65,8 @@ func (cfg *Defaults) Equals(other *Defaults) bool {
 
 	return other.DefaultServiceAccount == cfg.DefaultServiceAccount &&
 		other.DefaultRunAsUser == cfg.DefaultRunAsUser &&
-		other.DefaultRunAsGroup == cfg.DefaultRunAsGroup
+		other.DefaultRunAsGroup == cfg.DefaultRunAsGroup &&
+		other.DefaultRunAsNonRoot == cfg.DefaultRunAsNonRoot
 }
 
 // NewDefaultsFromMap returns a Config given a map corresponding to a ConfigMap
@@ -69,6 +75,7 @@ func NewDefaultsFromMap(cfgMap map[string]string) (*Defaults, error) {
 		DefaultServiceAccount: DefaultServiceAccountValue,
 		DefaultRunAsUser:      defaultRunAsUserValue,
 		DefaultRunAsGroup:     defaultRunAsGroupValue,
+		DefaultRunAsNonRoot:   defaultRunAsNonRootValue,
 	}
 
 	if defaultServiceAccount, ok := cfgMap[defaultServiceAccountKey]; ok {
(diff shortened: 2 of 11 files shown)


pkg/apis/config/default.go
```diff
@@ -78,27 +76,11 @@ func NewDefaultsFromMap(cfgMap map[string]string) (*Defaults, error) {
 	}
 
 	if defaultRunAsUser, ok := cfgMap[defaultRunAsUserKey]; ok {
```
khrm, line 78: Shouldn't this be 
```suggestion
	if defaultRunAsUser, ok := cfgMap[defaultRunAsUserKey]; !ok {
```
now? Otherwise, we are always setting the default value now.

khrm (commented):
I think "0" shouldn't be allowed in both the vanilla k8s and OpenShift.

pkg/apis/config/default.go
savitaashture: :thinking: 
```
	if defaultRunAsUser, ok := cfgMap[defaultRunAsUserKey]; ok {
		tc.DefaultRunAsUser = defaultRunAsUser
	}
```
overriding tc.DefaultRunAsUser to whatever coming from CM

khrm: Sorry. Read that line incorrectly.

savitaashture:
> I think "0" shouldn't be allowed in both the vanilla k8s and OpenShift.



> I think "0" shouldn't be allowed in both the vanilla k8s and OpenShift.

@khrm yes you are right its not allowed on k8s as well
Tried on kind
```
      waiting:
        message: 'container''s runAsUser breaks non-root policy (pod: "el-github-listener-5878566fb4-hqt5r_default(086e7204-f323-499c-90a2-4b7e0e277e2d)",
          container: event-listener)'
        reason: CreateContainerConfigError

```

khrm:
Do we allow setting runAsNonRoot as false?
If yes, then it might be a valid value.

savitaashture:
> Do we allow setting runAsNonRoot as false? If yes, then it might be a valid value.

Yes make sense
Added that https://github.com/tektoncd/triggers/pull/1747/commits/848ccb06b94069ef7eb181737ebc67568c1f8ceb

savitaashture:
@dibyom PR is ready PTAL

config/config-defaults-triggers.yaml
```diff
@@ -42,3 +42,4 @@ data:
     default-service-account: "default"
     default-run-as-user: "65532"
```
piyush-garg, line 43: i think we should default set them empty, and if user wants they can set as their desired value

pkg/apis/config/default.go
```diff
@@ -39,6 +41,11 @@ type Defaults struct {
 	DefaultServiceAccount string
 	DefaultRunAsUser      int64
```
piyush-garg, line 42: we dont need these default if we want to keep empty value in cm

config/config-defaults-triggers.yaml
savitaashture: Those values were set from starting and now if we remove we might break behavior
IMO its okay to set some default values when user won't set anything

We are setting default values here https://github.com/tektoncd/triggers/pull/1747/files#diff-572521168fb33345a93f07ab72e633d00cccb422144d8a05399a6fa98cb12025R33-R35

In the config-defaults-triggers.yaml is sample example to show how user can set as its under `_example: |`

pkg/apis/config/default.go
savitaashture: As mentioned here https://github.com/tektoncd/triggers/pull/1747#discussion_r1667315197 imo if user don't set empty then we can set them with default values 

It shouldn't be mandatory that user has to set values through CM its completely optional (which is the current behavior )

 Only thing is now we are providing way to configure different values by providing those keys in CM or remove default values for those keys if they don't want by setting `""`

khrm (commented):
Let's rebase these. @savitaashture We can merge the PR then.

savitaashture:
> Let's rebase these. @savitaashture We can merge the PR then.

@khrm Done
Its ready now

khrm (commented):
/approve

pkg/reconciler/eventlistener/resources/container.go
```diff
@@ -57,15 +57,19 @@ func MakeContainer(el *v1beta1.EventListener, configAcc reconcilersource.ConfigA
...
+			RunAsNonRoot: ptr.Bool(cfg.Defaults.DefaultRunAsNonRoot),
 			SeccompProfile: &corev1.SeccompProfile{
 				Type: corev1.SeccompProfileTypeRuntimeDefault,
 			},
 		}
 	}
 
-	containerSecurityContext.RunAsUser = ptr.Int64(cfg.Defaults.DefaultRunAsUser)
-	containerSecurityContext.RunAsGroup = ptr.Int64(cfg.Defaults.DefaultRunAsGroup)
+	if !cfg.Defaults.IsDefaultRunAsUserEmpty {
```
piyush-garg, line 67: can we have this also behind c.SetSecurityContext

piyush-garg: and other field like capability, priviledge escalation can be moved out of the flag

savitaashture: ya make sense

we can do that

PR #1748, opened 2024-07-07
fix(example): update default branch name in examples

l-qing:
# Changes

# Submitter Checklist

# Release Notes

```release-note
NONE
```

/kind documentation

### DIFF
--- a/examples/v1alpha1/cron/triggerbinding.yaml
+++ b/examples/v1alpha1/cron/triggerbinding.yaml
@@ -8,3 +8,5 @@ spec:
     value: main
   - name: gitrepositoryurl
     value: https://github.com/tektoncd/triggers
+  - name: contenttype
+    value: $(header.Content-Type)
--- a/examples/v1alpha1/cron/triggerbinding.yaml
+++ b/examples/v1alpha1/cron/triggerbinding.yaml
@@ -5,6 +5,6 @@ metadata:
 spec:
   params:
   - name: gitrevision
-    value: master
+    value: main
   - name: gitrepositoryurl
     value: https://github.com/tektoncd/triggers
--- a/examples/v1beta1/cron/triggerbinding.yaml
+++ b/examples/v1beta1/cron/triggerbinding.yaml
@@ -8,3 +8,5 @@ spec:
     value: main
   - name: gitrepositoryurl
     value: https://github.com/tektoncd/triggers
+  - name: contenttype
+    value: $(header.Content-Type)
--- a/examples/v1beta1/cron/triggerbinding.yaml
+++ b/examples/v1beta1/cron/triggerbinding.yaml
@@ -5,6 +5,6 @@ metadata:
 spec:
   params:
   - name: gitrevision
-    value: master
+    value: main
   - name: gitrepositoryurl
     value: https://github.com/tektoncd/triggers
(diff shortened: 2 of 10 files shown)


examples/v1alpha1/cron/triggerbinding.yaml
```diff
@@ -5,6 +5,6 @@ metadata:
 spec:
   params:
   - name: gitrevision
-    value: master
+    value: main
```
l-qing, line 8: In this demo, cloning the `master` branch will fail because it has been renamed to `main`.

examples/v1beta1/triggertemplates/triggertemplate.yaml
```diff
@@ -14,6 +14,7 @@ spec:
     default: This is the default message
   - name: contenttype
     description: The Content-Type of the event
+    default: text/plain
```
l-qing, line 17: In this demo, without a default value and without assigning value in `TriggerBinding`, the default value `$(tt.params.contenttype)` will be retained in `TaskRun`, which is quite strange.

khrm (commented):
/ok-to-test

examples/v1beta1/triggertemplates/triggertemplate.yaml
khrm: Do you mean to say we get `$(tt.params.contenttype)` when there's no value in TriggerBinding and no default value? It should during processing of events then,

l-qing: Yes, I understand that this value should be declared in the `TriggerBinding` as the source of the `contenttype` variable, which can be either `$(body.xxx)` or `$(extensions.xxx)`. However, I am not aware of any existing data source to obtain this parameter value, and it is not mentioned in the documentation.

So my current solution is to give it a default value first, which will not affect other `TriggerBindings` assigning it a suitable value in the future.

examples/v1beta1/cron/triggerbinding.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
 spec:
   params:
   - name: gitrevision
-    value: master
+    value: main
   - name: gitrepositoryurl
     value: https://github.com/tektoncd/triggers
+  - name: contenttype
+    value: $(header.Content-Type)
```
l-qing, line 12: Ref: https://github.com/tektoncd/triggers/blob/87e123e173028bd2d69547386b09ffda996a0394/examples/v1beta1/trigger-ref/triggerbinding.yaml#L5-L12

examples/v1beta1/triggertemplates/triggertemplate.yaml
l-qing: I referred to the code here and also added a configuration in `TriggerBinding` to assign a value to `contenttype`.

https://github.com/tektoncd/triggers/blob/87e123e173028bd2d69547386b09ffda996a0394/examples/v1beta1/trigger-ref/triggerbinding.yaml#L5-L12

dibyom (approved):

examples/v1beta1/triggertemplates/triggertemplate.yaml
l-qing: If there is no default value, it will not necessarily cause a problem.
It will only cause an issue if there is also no value assigned in the TriggerBinding.

savitaashture (commented):
/lgtm

PR #2019, opened 2026-05-11
Add TLS security profile support for core interceptors

jkhelil:
## Description

When deployed via the Tekton operator on OpenShift, the operator reads the cluster's `APIServer` TLS security profile (`config.openshift.io/v1`) and injects the following environment variables into the `tekton-triggers-core-interceptors` deployment:

- `TLS_MIN_VERSION` — e.g. `VersionTLS12`
- `TLS_CIPHER_SUITES` — e.g. `TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384`
- `TLS_CURVE_PREFERENCES` — e.g. `X25519,P-256,P-384`

This change makes the core interceptors HTTPS server read and apply those env vars to its `tls.Config`, honoring the cluster-wide TLS security profile.

Resolves: https://redhat.atlassian.net/browse/SRVKP-11953

## Changes

- **New package `pkg/interceptors/server/tlsconfig/`** — provides `Config` struct, `LoadFromEnv()`, and `ToTLSConfig()` to parse operator-injected TLS env vars into a `*tls.Config`. Mirrors the approach used in [tektoncd/results](https://github.com/tektoncd/results/blob/main/pkg/api/server/tlsconfig/tlsconfig.go).
  - Supports IANA cipher suite names and numeric IDs via Go's built-in `tls.CipherSuites()`
  - Supports multiple TLS version formats (`VersionTLS12`, `1.2`, `TLSv1.2`, numeric)
  - Supports curve names (`X25519`, `P-256`, `P-384`, `P-521`) including PQC hybrid (`X25519Kyber768Draft00`)
  - Includes human-readable formatting helpers for startup logging
- **Modified `cmd/interceptors/main.go`** — `startServer` now calls `tlsconfig.LoadFromEnv().ToTLSConfig()` to build the TLS configuration from env vars and logs the resolved TLS settings at startup. When no env vars are set, defaults to TLS 1.2 (preserving existing behavior).

## Release Note

```release-note
Core interceptors now honor the TLS security profile injected by the Tekton operator via TLS_MIN_VERSION, TLS_CIPHER_SUITES, and TLS_CURVE_PREFERENCES environment variables, allowing cluster-wide TLS policy enforcement on OpenShift.
```

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/interceptors/server/tlsconfig/tlsconfig.go
+++ b/pkg/interceptors/server/tlsconfig/tlsconfig.go
@@ -1,0 +1,303 @@
+/*
+Copyright 2026 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package tlsconfig provides TLS configuration parsing and management for the core interceptors server.
+// It supports loading configuration from environment variables injected by the Tekton operator,
+// and converts them to Go's tls.Config for use with the HTTPS server.
+package tlsconfig
+
+import (
+	"crypto/tls"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+// Config holds TLS configuration that can be loaded from environment variables
+type Config struct {
+	MinTLSVersion    string
+	CipherSuites     string
+	CurvePreferences string
+}
+
+// New creates a Config from explicit values
+func New(minVersion, cipherSuites, curvePreferences string) *Config {
+	return &Config{
+		MinTLSVersion:    minVersion,
+		CipherSuites:     cipherSuites,
+		CurvePreferences: curvePreferences,
+	}
+}
+
+// LoadFromEnv loads TLS configuration from environment variables.
+// This allows the configuration to be injected by the Tekton operator on OpenShift.
+func LoadFromEnv() *Config {
+	return New(
+		os.Getenv("TLS_MIN_VERSION"),
+		os.Getenv("TLS_CIPHER_SUITES"),
+		os.Getenv("TLS_CURVE_PREFERENCES"),
+	)
+}
+
+// ToTLSConfig converts the configuration to Go's tls.Config
+func (c *Config) ToTLSConfig() (*tls.Config, error) {
+	tlsConfig := &tls.Config{
+		MinVersion: tls.VersionTLS12,
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/interceptors/server/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,279 @@
...
+
+		if id, err := strconv.ParseUint(part, 10, 16); err == nil {
+			result = append(result, uint16(id))
+			continue
+		}
+
+		if id, ok := cipherMap[part]; ok {
+			result = append(result, id)
+			continue
+		}
```
khrm, line 169: Shouldn't there be some error when it doesn't match anything? Otherwise, a typo might be ignored.

khrm: Or at least a log.

pkg/interceptors/server/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,279 @@
...
+	parts := strings.Split(ciphers, ",")
+	result := make([]uint16, 0, len(parts))
+
+	for _, part := range parts {
+		part = strings.TrimSpace(part)
+		if part == "" {
+			continue
+		}
+
+		if id, err := strconv.ParseUint(part, 10, 16); err == nil {
```
khrm: Why aren't we supporting both Hex and Decimal?

pkg/interceptors/server/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,279 @@
...
+
+// parseTLSVersion converts a version string to tls.Version constant.
+// Supports: "1.0", "1.1", "1.2", "1.3" or numeric values like "771" (0x0303)
+func parseTLSVersion(version string) (uint16, error) {
+	if num, err := strconv.ParseUint(version, 10, 16); err == nil {
+		return uint16(num), nil
+	}
+
+	switch version {
+	case "1.0", "TLS1.0", "TLSv1.0", "VersionTLS10":
```
khrm: Isn't this a regression in terms of security? Earlier we were supporting a min. of TLS2 only. Now, we can use 1 and 1.1 also.

We shouldn't allow admin to set these.

pkg/interceptors/server/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,279 @@
...
+	"CurveP256": tls.CurveP256,
+	"CurveP384": tls.CurveP384,
+	"CurveP521": tls.CurveP521,
+	"X25519":    tls.X25519,
+	"P-256":     tls.CurveP256,
+	"P-384":     tls.CurveP384,
+	"P-521":     tls.CurveP521,
+
+	// Post-Quantum hybrid curve
+	"X25519Kyber768Draft00": tls.CurveID(25497),
```
khrm: Is this still valid in Go 1.25?  Check here: X25519MLKEM768     [CurveID](https://pkg.go.dev/crypto/tls#CurveID) = 4588

khrm (commented):
/approve

PR #2077, opened 2026-07-27
fix: don't pre-parse url-encoded body for webhook-first interceptor chains

### LINKED ISSUE #1886: Custom Slack Interceptor Unexpected Behavior After Triggers Upgrade
# Expected Behavior

Trigger hands off Slack urlencoded request to the custom slack interceptor, which then decodes the body into JSON and returns it to the Trigger for further processing

# Actual Behavior

After upgrade to Triggers v0.30.1, it appears that the Trigger is decoding the Slack request and passes JSON to the Slack interceptor instead of the original urlencoded body

# Additional Info

- Kubernetes version: v1.31.9

```
Client Version: v1.33.3
Kustomize Version: v5.6.0
Server Version: v1.31.9
```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

v0.68.1

# More Info

I'm in the process of upgrading my Tekton environment from and to the following versions, respectively:

Pipeline:  v0.44.4 ---> v0.68.1
Triggers:  v0.23.1 ---> v0.30.1

In preparation for the upgrade, I migrated the apiVersions of my manifests as follows:

Pipelines/Tasks/etc:  **tekton.dev/v1beta1** ---> **apiVersion: tekton.dev/v1**
Triggers/TriggerTemplates/EventListener:  **triggers.tekton.dev/v1alpha1** ---> **triggers.tekton.dev/v1beta1**

Before I upgraded any of the Tekton components, I confirmed everything was working as expected after the API migrations.  Everything was going fine...

However, I have a Slack integration with my Tekton setup.  Until now I've been using a [custom Slack interceptor](https://github.com/bitsofinfo/slack-payload-handler) to convert the urlencoded body of a Slack slash command to JSON.  The EL looks something like this:

```
---
apiVersion: triggers.tekton.dev/v1beta1
kind: EventListener
metadata:
  name: event-listener-v1
  annotations:
    tekton.dev/payload-validation: "false"
spec:
  triggers:
    - name: slackcmd-interceptor
      interceptors:
        - webhook:
            objectRef:
              kind: Service
              name: slack-payload-handler
              apiVersion: v1
              namespace: tekton-tasks

        - ref:
            name: cel
          params:
            - name: filter
              value: >-
                has(body.api_app_id) &&
                body.token.compareSecret('slack','el-secrets') &&
                has(body.command) &&
                has(body.text) &&
                body.command.indexOf('app-deploy') != -1

            - name: overlays
              value:
                - key: env
                  expression: "body.text.split(' ')[0]"

                - key: appName
                  expression: "body.text.split(' ')[1]"

                - key: appTag
                  expression: "body.text.split(' ')[2]"

                - key: slackUserId
                  expression: "body.user_id"

                - key: slackChannelId
                  expression: "body.channel_id"

                - key: b64body
                  expression: "base64.encode(bytes(body.marshalJSON()))"

      bindings:
        - ref: start-via-cmd-app-deploy-v1
        - name: b64body
          value: $(extensions.b64body)
      template:
        ref: start-v1
```

So the EL would pass Slack's urlencoded request to the slack-payload-handler, which would return JSON similar to this:

```
{
  "api_app_id": "ABCDE",
  "channel_id": "FGHIJ",
  "channel_name": "my-channel",
  "command": "/app-deploy",
  "response_url": "https://hooks.slack.com/commands/T...",
  "team_domain": "my-company",
  "team_id": "KLMNO",
  "text": "env appName appTag",
  "token": "XXXXX",
  "trigger_id": "12345...",
  "user_id": "YYYYY",
  "user_name": "my.user"
}
```

This stopped working after I upgraded Triggers to v0.30.1.  Did some troubleshooting and found that the Slack request that the EL hands off to the slack-payload-handler is no longer in Slack's urlencoded format, but seems to be already decoded into JSON with the following format

```
{
  "api_app_id": ["ABCDE"],
  "channel_id": ["FGHIJ"],
  "channel_name": ["my-channel"],
  "command": ["/app-deploy"],
  "response_url": ["https://hooks.slack.com/commands/T..."],
  "team_domain": ["my-company"],
  "team_id": ["KLMNO"],
  "text": ["env appName appTag"],
  "token": ["XXXXX"],
  "trigger_id": ["12345..."],
  "user_id": ["YYYYY"],
  "user_name": ["my.user"]
}
```

So I guess my first question is - why is the Event Listener decoding the incoming Slack request before it even gets sent to the slack-payload-handler webhook?  I know that Tekton has since added its own Slack interceptor, but I'm not using it in my spec.

Second question - why is it converting the string value for each key in the JSON object to a list.  This messes up tasks in my pipeline that are unable to parse it properly.

Any help would be greatly appreciated.  Again I know that Tekton now has an official Slack interceptor (which I did actually try to use and got the same unwanted result), but I'd like to still use the custom one.

Thank you!

pujitha24:
/kind bug

Motivation:

EventListener.ExecuteInterceptors() converts a url-encoded request body
(Content-Type: application/x-www-form-urlencoded) into a JSON string
(map[string][]string) before running any interceptor in the trigger's
chain. This conversion was added to let ref-style interceptors (CEL, the
built-in Slack ClusterInterceptor) address form fields as `body.<field>`.

However the conversion runs unconditionally, so it also applies when the
first interceptor in the chain is a deprecated `webhook` interceptor
(Interceptor.Webhook, an arbitrary external URL called directly by the
sink). That interceptor type is documented to forward the request body
to an external service, and users writing custom webhook interceptors to
parse form-encoded payloads themselves (e.g. Slack slash commands)
reasonably expect to receive the original raw body.

As reported in #1886, after upgrading Triggers a custom webhook
interceptor started receiving a pre-parsed JSON body with every field
value wrapped in a single-element array (e.g. `"token":["abc"]`) instead
of the original url-encoded string, breaking the interceptor's own
parsing logic. User-visible behavior for the built-in Slack/CEL
interceptor path is unchanged; this only affects the deprecated
`webhook` interceptor when it is the first interceptor in the chain and
the request Content-Type is application/x-www-form-urlencoded.

Approach:

Skip the url-encoded-to-JSON pre-conversion when the first interceptor
in the trigger's interceptor chain is a `webhook`-type interceptor, so
it receives the original, unmodified body. Ref-style interceptors
(ClusterInterceptor/Interceptor, including CEL and the built-in Slack
interceptor) are unaffected when they are first in the chain.

Validation:

  go test ./pkg/sink/...

...passed, including a new regression test
(TestExecuteInterceptor_WebhookFirstGetsRawURLEncodedBody) verifying a
webhook interceptor placed first in the chain receives the raw
url-encoded body. Confirmed the new test fails without this change
(receives the pre-parsed JSON body instead) and passes with it.

```release-note
Fix a bug where a `webhook`-type interceptor placed first in a trigger's
interceptor chain would receive a pre-parsed JSON body instead of the
original request body for `application/x-www-form-urlencoded` requests
(e.g. Slack slash commands), breaking custom interceptors that parse the
raw form-encoded payload themselves.
```

Fixes #1886

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -483,7 +483,11 @@ func (r Sink) ExecuteInterceptors(trInt []*triggersv1.TriggerInterceptor, in *ht
 	parsedQuery, _ := net.ParseQuery(request.Body)
 
 	// parse form-data payload
-	if v := in.Header.Get("Content-Type"); v == "application/x-www-form-urlencoded" && len(parsedQuery) > 1 {
+	// Skip this when the chain starts with a deprecated webhook interceptor: that
+	// interceptor forwards the request body verbatim to an external URL, so it
+	// must receive the original body rather than a pre-parsed JSON representation.
+	if v := in.Header.Get("Content-Type"); v == "application/x-www-form-urlencoded" && len(parsedQuery) > 1 &&
+		(len(trInt) == 0 || trInt[0].Webhook == nil) {
 		// Convert the map into a JSON string
 		jsonString, err := json.Marshal(parsedQuery)
 		if err != nil {
--- a/pkg/sink/sink.go
+++ b/pkg/sink/sink.go
@@ -483,15 +483,18 @@ func (r Sink) ExecuteInterceptors(trInt []*triggersv1.TriggerInterceptor, in *ht
 	parsedQuery, _ := net.ParseQuery(request.Body)
 
 	// parse form-data payload
-	// Skip this when the chain starts with a deprecated webhook interceptor: that
-	// interceptor forwards the request body verbatim to an external URL, so it
-	// must receive the original body rather than a pre-parsed JSON representation.
+	// Skip this when the chain starts with a deprecated webhook interceptor and there
+	// are no extensions to merge into the body: that interceptor forwards the request
+	// body verbatim to an external URL, so it must receive the original body rather
+	// than a pre-parsed JSON representation. If extensions are present, they still
+	// need to be merged in below via extendBodyWithExtensions, which requires a JSON
+	// body, so the conversion must happen.
 	if v := in.Header.Get("Content-Type"); v == "application/x-www-form-urlencoded" && len(parsedQuery) > 1 &&
-		(len(trInt) == 0 || trInt[0].Webhook == nil) {
+		(trInt[0].Webhook == nil || len(request.Extensions) > 0) {
 		// Convert the map into a JSON string
 		jsonString, err := json.Marshal(parsedQuery)
 		if err != nil {
-			log.Errorf("Error converting map to JSON:", err)
+			log.Errorf("Error converting map to JSON: %v", err)
 			return nil, nil, nil, err
 		}
 		request.Body = string(jsonString)
--- a/pkg/sink/sink_test.go
+++ b/pkg/sink/sink_test.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"log"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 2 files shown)


pujitha24:
Thanks for the review! Both points are fixed in the latest commit:

1. The urlencoded→JSON conversion now also runs when `request.Extensions` is non-empty, since `extendBodyWithExtensions` needs a JSON body to merge extensions into (previously it would silently drop the original form fields for a webhook-first chain that also has extensions, e.g. from a preceding TriggerGroup interceptor). I also removed the now-unreachable `len(trInt) == 0` check, since the function already returns early for an empty interceptor list.
2. Fixed the `log.Errorf` call to include the `%v` verb so the underlying error is actually logged.

Added a regression test (`TestExecuteInterceptor_WebhookFirstWithExtensionsGetsJSONBody`) that verifies the webhook interceptor receives a JSON body with both the original form fields and the merged extensions when extensions are present. I confirmed it fails against the old condition and passes with the fix.

khrm (commented):
/lgtm

khrm (commented):
/approve

PR #2084, opened 2026-08-05
Don't require a ClusterInterceptor to exist before EventListeners can start

### LINKED ISSUE #2034: EventListener crashes on startup when any ClusterInterceptor exists without `server/type: https` label + caBundle
# Expected Behavior

An EventListener with **no interceptors** in its triggers should start successfully, without requiring any ClusterInterceptor resources to exist.

# Actual Behavior

The EventListener pod enters `CrashLoopBackOff` with:

```
Timed out waiting on CaBundle to available for clusterInterceptor: ...
Timed out waiting on CaBundle to available for Interceptor: ...
empty caBundle in clusterInterceptor spec
```

This happens even though the EventListener has zero interceptors configured. The binary unconditionally scans **all** ClusterInterceptors in the cluster and demands at least one with both:

- `metadata.labels["server/type"] = "https"`
- `spec.clientConfig.caBundle` non-empty

Root cause is three interacting problems in the codebase:

1. **`pkg/adapter/adapter.go:161-174`** — `getCertFromInterceptor()` polls all ClusterInterceptors. If `httpsCILen == 0` (no https-labeled ClusterInterceptor found), it returns `errors.New("empty caBundle in clusterInterceptor spec")` — rejecting startup.

2. **`pkg/apis/triggers/v1alpha1/cluster_interceptor_defaults.go:27-28`** — `SetDefaults()` returns immediately when `IsUpgradeViaDefaulting(ctx)` is false (always the case for normal webhook requests), so the `server/type` label is never defaulted. Users must set it manually.

3. **`cmd/webhook/main.go:79,104`** — Both mutating and validating webhooks have `disallowUnknownFields: true`. This causes `caBundle` to be rejected as `json: unknown field "caBundle"`, even though `ClientConfig` at `pkg/apis/triggers/v1alpha1/cluster_interceptor_types.go:68` defines `CaBundle []byte` with `json:"caBundle,omitempty"`.

Combined: it is impossible to deploy a working EventListener when any ClusterInterceptor exists, unless the separate `interceptors` server image injects caBundle programmatically via `pkg/interceptors/server/server.go:236`.

1. Install Tekton Triggers v0.34.0 or v0.35.0.
2. Create a `ClusterInterceptor` (any kind, e.g. `github`) **without** `caBundle` and **without** the `server/type` label.
3. Create an `EventListener` with **no interceptors** in its triggers (e.g. just bindings and a template).
4. Observe the EventListener pod crash with the error above.

# Additional Info

- Kubernetes version:

  ```
  Client Version: v1.32.0
  Kustomize Version: v5.5.0
  Server Version: v1.32.0
  ```

- Tekton Pipeline version:

  **Output of `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

  ```
  v1.6.0
  ```

- Tekton Triggers version: v0.34.0 (also confirmed on v0.35.0 — adapter.go is identical)
- Workaround: temporarily delete both triggers webhooks, create the ClusterInterceptor with `labels: {server/type: https}` and a valid base64-encoded PEM CA cert in `spec.clientConfig.caBundle`, then restore the webhooks.

pujitha24:
# Changes

`getCertFromInterceptor()` in `pkg/adapter/adapter.go` builds the TLS cert pool
the EventListener sink uses to call interceptors. Before accepting the pool as
ready, it required BOTH "at least one ClusterInterceptor labeled
`server/type: https` exists" AND "every such ClusterInterceptor has a non-empty
caBundle." The first requirement is unnecessary: if zero https-labeled
ClusterInterceptors exist, there's nothing to validate, so startup should not
be blocked. In practice this meant every EventListener pod in the cluster
would crash-loop with `empty caBundle in clusterInterceptor spec` whenever no
https-labeled ClusterInterceptor existed cluster-wide (e.g. a minimal install
without the core-interceptors manifest, or a custom ClusterInterceptor created
without the `server/type` label) -- even for EventListeners whose triggers use
no interceptors at all.

This drops the `httpsCILen == 0` requirement, leaving only the genuine
mismatch check: some https-labeled ClusterInterceptor exists but lacks a
caBundle. The equivalent loop over `Interceptor` resources a few lines below
never had this extra requirement, so this brings the `ClusterInterceptor` loop
in line with it.

Validated with `go build ./pkg/...` and `go test ./pkg/adapter/... -v
-count=10` (all pass, no flakiness). `TestGetHTTPClientNoClusterInterceptors`
reproduces the reported crash path with a fake clientset/informer holding zero
ClusterInterceptors and confirms `getHTTPClient()` now succeeds instead of
erroring. `TestGetHTTPClientPartialCaBundleMismatch` confirms a genuine
caBundle mismatch is still correctly rejected.

This does not address the issue's other two named root causes (defaulting
only applying the `server/type` label during upgrade-via-defaulting, and a
claim about `disallowUnknownFields` rejecting `caBundle`) -- those are
independent of this code path.

# Submitter Checklist

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) if any changes are user facing, including updates to minimum requirements e.g. Kubernetes version bumps
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind bug`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep

# Release Notes

```release-note
Fix EventListener pods crash-looping with "empty caBundle in clusterInterceptor spec" when no ClusterInterceptor labeled `server/type: https` exists in the cluster, even when the EventListener's triggers use no interceptors.
```

Fixes #2034

### DIFF
--- a/pkg/adapter/adapter.go
+++ b/pkg/adapter/adapter.go
@@ -170,7 +170,7 @@ func (s *sinker) getCertFromInterceptor(certPool *x509.CertPool) error {
 			}
 		}
 
-		if httpsCILen == 0 || httpsCILen != count {
+		if httpsCILen != count {
 			return false, errors.New("empty caBundle in clusterInterceptor spec")
 		}
 
--- a/pkg/adapter/adapter_test.go
+++ b/pkg/adapter/adapter_test.go
@@ -32,7 +32,7 @@ import (
 	"knative.dev/pkg/logging"
 )
 
-func TestGetHTTPClientEmptyCaBundle(t *testing.T) {
+func TestGetHTTPClientNoClusterInterceptors(t *testing.T) {
 	recorder, err := sink.NewRecorder()
 	if err != nil {
 		log.Fatal(err.Error())
@@ -50,14 +50,11 @@ func TestGetHTTPClientEmptyCaBundle(t *testing.T) {
 	}
 
 	c, err := s.getHTTPClient()
-	if err != nil && !strings.Contains(err.Error(), "empty caBundle in clusterInterceptor spec") {
-		t.Fatal(err)
+	if err != nil {
+		t.Fatalf("getHTTPClient() should not fail when no ClusterInterceptors exist, got error: %v", err)
 	}
-	if err == nil {
-		t.Fatalf("test should fail as clusterinterceptor spec cabundle is empty")
-	}
-	if diff := cmp.Diff(c, &http.Client{}); diff != "" {
-		t.Errorf("Diff: -want +got: %s", cmp.Diff(c, &http.Client{}))
+	if c == nil {
+		t.Fatal("expected a non-nil http.Client")
 	}
 }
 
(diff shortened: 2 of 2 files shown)


savitaashture (approved):

khrm (commented):
/lgtm

PR #2143, opened 2026-09-20
Add a request body size limit to the EventListener sink

### LINKED ISSUE #2021: EventListener: Add request body size limit to prevent OOM from large payloads
## Summary

The EventListener reads the entire request body into memory with no size limit via `io.ReadAll(request.Body)` at two call sites:

- `pkg/sink/sink.go:153`
- `pkg/sink/validate_payload.go:29`

An attacker (or misbehaving webhook sender) streaming large payloads can exhaust the pod's memory, causing OOM kills. With memory limits set (standard for production), sustained large requests can keep the pod in CrashLoopBackOff.

The existing `ReadTimeout` (default 5s) does not mitigate this — in-cluster, data arrives fast enough to OOM the pod well within the timeout window.

## Proposed Fix

Replace bare `io.ReadAll` with `http.MaxBytesReader` at both call sites, with a default cap of **3 MiB**. This:

- Rejects oversized requests with `413 Request Entity Too Large`
- Closes the connection immediately rather than draining the stream

3 MiB provides ample headroom — typical webhook payloads from GitHub, GitLab, and Bitbucket are 5–100 KB; the largest observed are ~1 MB.

The limit should be configurable via an EventListener flag (e.g. `--el-max-body-size`).

## Example

```go
// Before
event, err := io.ReadAll(request.Body)

// After
request.Body = http.MaxBytesReader(response, request.Body, maxBodySize)
event, err := io.ReadAll(request.Body)
```

KR-Ravindra:
# Changes

Fixes https://github.com/tektoncd/triggers/issues/2021

## Problem

The EventListener sink reads the whole request body into memory with no upper bound. `IsValidPayload` (which wraps
every sink request in `pkg/adapter/adapter.go`) calls `io.ReadAll(request.Body)` and then buffers the result again so
`HandleEvent` can re-read it, so a single large POST is held twice in the sink's heap. The server's `ReadTimeout`
bounds how long a request may take, not how many bytes it may send, and a client that streams quickly stays well
inside it. One request is enough to OOM the sink pod.

## Fix

Wrap the body in `http.MaxBytesReader` before reading it, and answer an over-sized request with `413 Request Entity
Too Large` in the sink's existing JSON error shape (`EventListener`, `Namespace`, `ErrorMessage`) rather than the
current `500`. The failure is counted through the same `recordCountMetrics(failTag)` path as every other refused event.

The limit is plumbed exactly like the existing payload-validation setting: a new controller flag `-el-max-body-size`
(default 3 MiB) becomes the sink's `--max-body-size`, so it can be tuned per install and set to `0` to disable the
limit entirely and keep the old behaviour. 3 MiB was chosen because it comfortably exceeds the largest payload the
supported interceptors receive (a GitHub `push` webhook is well under 1 MiB) while keeping the worst case bounded.

## How tested

New case `event body over the size limit` in `pkg/sink/validate_payload_test.go`, added before the change. Against the
unmodified sink it fails:

```
--- FAIL: TestSink_IsValidPayload/event_body_over_the_size_limit
    validate_payload_test.go:138: Status code mismatch: got 202, want 413
```

and passes with the fix. `go test ./cmd/... ./pkg/...` is green (eleven golden container-argument lists in the
EventListener reconciler tests were updated for the new flag), and `go build ./...`, `go vet ./pkg/sink/...
./pkg/adapter/...` and `make fmt` are clean. `golangci-lint` is not installed here, so repository lint was not run
locally.

# Submitter Checklist

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) if any changes are user facing, including updates to minimum requirements e.g. Kubernetes version bumps
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
The EventListener sink now rejects event requests whose body exceeds a configurable limit with `413 Request Entity Too Large`. The limit is set by the new controller flag `-el-max-body-size` (bytes, default 3145728 = 3 MiB); set it to `0` to keep the previous unbounded behaviour.
```

---
AI usage disclosure: this change was prepared by an AI agent I operate; I reviewed the diff and ran the tests myself
before opening it. I point the agent only at problems I have actually hit in my own work — here, an EventListener
sink being OOM-killed — not at producing contributions.


KR-Ravindra:
Round 1 self-review.

1. Duplicate check: #2021 has no other PR linked to it and no open PR in this repo touches `pkg/sink/validate_payload.go`; the design follows the one proposed in the issue (`http.MaxBytesReader`, 413, configurable via a controller flag, 3 MiB default).
2. Verified against `main`: `IsValidPayload` wraps the `/` handler in `pkg/adapter/adapter.go`, so the limit applies to every event request and `HandleEvent` re-reads the buffered copy rather than the raw body. `http.MaxBytesError` is available on the module's Go version. The golden container-argument updates plus the passing `Check generated code` job show nothing else needs regenerating.
3. Brought the body in line with the PR template (Changes, Submitter Checklist, Release Notes block) so the release-note label blocker can clear.
4. Lint, unit tests, build and coverage are green; I will mark this ready once the e2e job finishes.

/kind feature

This comment was drafted with AI assistance and checked against the code; I point the tool only at problems I have hit in my own work, not at producing contributions.

KR-Ravindra:
@khrm @savitaashture gentle ping when you have a moment: this has been ready since 09-21 with CI green on 5cb3ee0 (issue #2021). Happy to change the 3 MiB default or the flag name if you would prefer something else.

This comment was drafted with AI assistance and checked against the code.

pkg/sink/validate_payload_test.go
```diff
@@ -71,13 +74,50 @@ func TestSink_IsValidPayload(t *testing.T) {
...
+						TriggerRef: "test",
+					}},
+				},
+			}},
+		},
+		eventBody:      json.RawMessage(`{"head_commit": {"id": "testrevision"}}`),
+		maxBodySize:    1024,
+		wantStatusCode: http.StatusAccepted,
+	}, {
+		name: "event body over the size limit",
```
khrm: Can we have a disable test also? When we set 0?

khrm (approved):

pkg/sink/validate_payload_test.go
KR-Ravindra: @khrm done in a1e5743: added the "size limit disabled with zero" case, which sends a 4 MiB body (over the 3 MiB default) with the limit set to 0 and expects 202. Verified locally that the same case returns 413 when the default limit is applied. Could you take another look when you have a moment?

This comment was drafted with AI assistance and checked against the code.

KR-Ravindra:
@khrm the zero-limit test made the verbose CI run log the whole 4 MiB payload as one line and the `test` job stalled for over an hour on a1e5743, so in 484d43c the case now sends the same 512-byte body the 64-byte case rejects, with the limit set to 0, and expects 202. Test-only change; `go test -race ./pkg/sink` passes. Could you re-lgtm when CI is green?

Drafted with AI assistance and checked against the code.

pkg/sink/validate_payload_test.go
khrm: LGTM.

khrm (commented):
/lgtm

